PR #11, opened 2020-09-17
Fix tests build

brusdev:


RoddieKieley:
@brusdev This does leave an outstanding issue related to the qpid-dispatch-test test, but will raise that for review independently.

PR #15, opened 2020-09-23
ENTMQBR-2885: Provide Init Containers For Container Image

### LINKED ISSUE #6: Add tool to convert an ActiveMQ Artemis CR instance to a jinja template for YamlConfiger
With the move to generating broker configuration via the [YamlConfiger](https://github.com/rh-messaging-qe/YamlConfiger) tool a way to get the users desired configuration from the custom resource and into an appropriate template is required. 

An independently buildable cr2jinja tool can be added to the operator's tools directory like the previously added csv-gen tool. This tool can then be built independently yet depend on other aspects of the operator codebase such as the CRD's and associated go side structs.

Originally the thought here was that this would be internal to the operator itself and the functionality would not be exposed externally. However it is also desired to make the tool generally useful in concert with the YamlConfiger tool for configuring ActiveMQ Artemis brokers in general, whether running within kubernetes or not. Being able to be built and executed independently will allow for usage independent of deployment on kubernetes.

gaohoward:
Added address-settings support
  version bumped to v3alpha2


RoddieKieley:
@gaohoward Manually committed this via [426c6cef611313b528511e717aa3443822e94473](https://github.com/artemiscloud/activemq-artemis-operator/commit/426c6cef611313b528511e717aa3443822e94473) with an amended comment to use v2alpha3 instead of v3alpha2

PR #22, opened 2020-10-20
Allow user to update with new address settings

### LINKED ISSUE #20: Allow user to update with new address settings
Currently user must include their address-settings in the cr on first deployment.
if address settings need to change it has to redeploy with the new cr.

The user should be able to modify the address-settings in the original cr and use
"oc apply" to update it.

**Additional context**

gaohoward:


pkg/controller/broker/v2alpha3/activemqartemis/activemqartemis_reconciler.go
```diff
@@ -225,6 +233,28 @@ func (reconciler *ActiveMQArtemisReconciler) ProcessDeploymentPlan(customResourc
...
+	}
+
+	//we need to compare old with new and update if they are different.
+	return compareAddressSettings(&prevCustomResource.Spec.AddressSettings, &customResource.Spec.AddressSettings)
+}
+
+//returns true if currentAddressSettings need update
+func compareAddressSettings(currentAddressSettings *brokerv2alpha3.AddressSettingsType, newAddressSettings *brokerv2alpha3.AddressSettingsType) bool {
+
+	if len((*currentAddressSettings).AddressSetting) != len((*newAddressSettings).AddressSetting) || *(*currentAddressSettings).ApplyRule != *(*newAddressSettings).ApplyRule || !config.IsEqual((*currentAddressSettings).AddressSetting, (*newAddressSettings).AddressSetting) {
```
RoddieKieley, line 252: Just because you can doesn't mean you should. We should strive to write in a way that is immediately obvious without having to think about it.

PR #33, opened 2020-11-05
Add Jolokia JVM Agent

brusdev:


tools/csv-gen/csv-gen.go
```diff
@@ -77,7 +77,7 @@ func main() {
...
...          \"sslEnabled\": false,\n                   \"enabledCipherSuites\": \"SSL_RSA_WITH_RC4_128_SHA,SSL_DH_anon_WITH_3DES_EDE_CBC_SHA\",\n                   \"enabledProtocols\": \"TLSv1,TLSv1.1,TLSv1.2\",\n                   \"needClientAuth\": true,\n                   \"wantClientAuth\": true,\n                   \"verifyHost\": true,\n                   \"sslProvider\": \"JDK\",\n                   \"sniHost\": \"localhost\",\n                   \"expose\": true\n                }\n             ]\n          }\n       },\n      {\n          \"apiVersion\": \"broker.amq.io/v2alpha1\",\n          \"kind\": \"ActiveMQArtemisAddress\",\n          \"metadata\": {\n              \"name\": \"ex-aaoaddress\"\n          },\n          \"spec\": {\n              \"addressName\": \"myAddress0\",\n              \"queueName\": \"myQueue0\",\n              \"routingType\": \"anycast\"\n          }\n      },\n      {\n          \"apiVersion\": \"broker.amq.io/v2alpha1\",\n          \"kind\": \"ActiveMQArtemisScaledown\",\n          \"metadata\": {\n              \"name\": \"ex-aaoscaledown\"\n          },\n          \"spec\": {\n              \"localOnly\": \"true\"\n          }\n      }"}
```
RoddieKieley, line 80: @brusdev v2alpha2?

pkg/resources/environments/environment.go
```diff
@@ -96,7 +96,7 @@ func DetectOpenshift() (bool, error) {
 	return err == nil, nil
 }
 
-func AddEnvVarForBasic(requireLogin string, journalType string) []corev1.EnvVar {
+func AddEnvVarForBasic(requireLogin string, journalType string, jolokiaAgentEnabled string, managementRBACEnabled string) []corev1.EnvVar {
```
RoddieKieley, line 99: @brusdev a part of the idea here with keeping the original versions was to keep the original code intact and untouched leaving the previous implementation as is in case needed. That would also however mean new func's would need to be added which built on the previous one instead of being changed directly.

RoddieKieley (changes requested):

tools/csv-gen/csv-gen.go
brusdev: I have just reverted this change.

pkg/resources/environments/environment.go
brusdev: I have just reverted this change and created new environment functions.

Issue #37, opened 2020-11-13
ConfigDeleteAddresses parameter is interpreted wrong by the init container


gaohoward:
**Describe the bug**
If address-setting property configDeleteAddresses is set to "OFF" in AddressSettings CR
This generates attached broker.xml which is not valid, it will be translated wrongly as "False"

The root cause is that yacfg's yaml loader always translated certain string values as boolean
see https://yaml.org/type/bool.html
So values like 'ON', 'OFF' etc will get changed into boolean True and False.
Also the empty string will be translated into None if the item is a string in yaml

**To Reproduce**
1.Create a CR with the above address setting value.
2. Deploy the cr

**Platform (please complete the following information):**
 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**
Optional, use if not obvious. A clear and concise description of what you expected to happen.

**Screenshots**
Optional, add screenshots to help explain your problem.

**Additional context**
Optional. Add any other context about the problem here.

PR #38, opened 2020-11-13
ENTMQBR-4177/issue #37 ConfigDeleteAddresses parameter is interpreted wrong

gaohoward:
To workaround this problem we add a map of special values and pass it
to yacfg so it can overrides those wrong conversions.

Depends on https://github.com/rh-messaging-qe/yacfg/issues/24


gaohoward:
Note: This PR depends on an update to yacfg/init-image: https://github.com/rh-messaging-qe/yacfg/pull/25

gaohoward:
We also need a new init-image tag to include the new yacfg update.

PR #49, opened 2021-01-26
ENTMQBR-4448 '*' char is not properly handled in address-settings config

gaohoward:
Also general improvement on special char handling.


gaohoward:
depends on this pr https://github.com/rh-messaging-qe/yacfg/pull/36

PR #53, opened 2021-02-11
Correct logging that contained non printable chars

remixtj:
While debugging we find out in logs that instead of the number of the
pods running there was a strange char shown. We found that chars printed
where ^C and ^B, which respectively corresponds to the char 3 and 2 of
ascii table. With this fix we use correct fmt function to print out the
correct value.


gaohoward:
@remixtj Hi can you explain why the char 3 and 2 get printed as ^C and ^B?

remixtj:
@gaohoward As you can see here: http://www.physics.udel.edu/~watson/scen103/ascii.html ^B and ^C are respectively the char 2 and 3 of ascii table. So as far as i can see, string(len(status.Ready))) uses the numeric value as ascii char, and not formatting it as string.

gaohoward:
@remixtj I see. Thanks for the explanation.

PR #135, opened 2022-01-27
activemqartemis instances showing "The field podStatus is invalid." when installed from operatorhub #134

### LINKED ISSUE #134: activemqartemis instances showing "The field podStatus is invalid." when installed from operatorhub
**Describe the bug**
Installing activemqartemis broker to Openshift via operator will get this error message on the operator console (operatorhub)

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:
ENTMQBR-5601

manually tested.


brusdev:
@howardgao looking openshift/console[1][2], this issue could be due to the missing statusDescriptor
https://github.com/RHsyseng/operator-utils#pods-deployment-status
https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#statusdescriptors-1
`statusDescriptors:
        - description: The current pods
          displayName: Pods Status
          path: podStatus
          x-descriptors:
            - "urn:alm:descriptor:com.tectonic.ui:podStatuses"`

[1]https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/common.tsx#L15
[2]https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/status/index.tsx#L26

brusdev:
@gaohoward looking operator-utils doc[1], it seems we are using the wrong client to update the status resource, indeed int the operator-sdk doc there is:
https://github.com/operator-framework/operator-sdk/blob/v0.10.0/doc/user/client.md#update
```
Update does *not* update the resource's status subresource
```
https://github.com/operator-framework/operator-sdk/blob/v0.10.0/doc/user/client.md#updating-status-subresource
```
When updating the [status subresource](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource) from the client, the StatusWriter must be used which can be gotten with Status()
```

gtully:
It may be relevant, in my attempts to put together a round trip test with the sdk scaffolding, where i deploy a dud CR and verify the status, the comments above from @brusdev  provided the missing link. In addition to using the wrong stateful set name, it looks like the Status() client is also needed. That sorted my test case. see: https://github.com/artemiscloud/activemq-artemis-operator/pull/138

gaohoward:
@gtully @brusdev I think Gary's commit fixes the concern Dom pointed out. Should we go ahead to merge it or further changes needed?

brusdev:
@howardgao I'm not sure we need to rename `PodStatus`, I think the fixes included by @gtully should also fix the error "The field podStatus is invalid."

gaohoward:
@brusdev @gtully verified that the error message is gone with Gary's fix. Will drop the PR. Thanks guys!

PR #138, opened 2022-02-07
add cr/status check round trip test test using the operator sdk scaff…

gtully:
…olding. Use share state for openshift check and fix status update logic, name and use of Status() from the sdk client


gtully:
the intent here is to have a simple round trip test to from the basis for TTD in an easy to debug environment. It seems to hang together nicely so should be a good start. Feedback welcome!

andytaylor (approved):
looks gtm, altho Howard may have mor eimput

PR #141, opened 2022-02-08
Added KUBE_TOOL var in Makefile so that user can use oc to build stuff under openshift env by

gaohoward:
Added KUBE_TOOL var in Makefile so that user can use oc to build stuff under openshift env by

make KUBE_TOOL=oc ....


brusdev (approved):
LGTM +1, maybe I would replace `KUBE_TOOL` with `KUBE_CLI`

gaohoward:
thanks!

PR #142, opened 2022-02-08
fix use of detect openshift before auto_detect start

gtully:


gaohoward:
It works fine on openshift/k8s. will merge away.

PR #148, opened 2022-02-14
poc with properties via config map

gtully:


gtully:
some output of what this does:

[gtully@gtully activemq-artemis-operator]$ kubectl exec ex-aao-ss-0 -n activemq-artemis-operator -- ls -lr /amq/extra/configmaps/broker-properties/broker.properties
Defaulted container "ex-aao-container" out of: ex-aao-container, ex-aao-container-init (init)
lrwxrwxrwx. 1 root root 24 Feb 14 15:37 /amq/extra/configmaps/broker-properties/broker.properties -> ..data/broker.properties
[gtully@gtully activemq-artemis-operator]$ kubectl exec ex-aao-ss-0 -n activemq-artemis-operator -- cat /amq/extra/configmaps/broker-properties/broker.properties
Defaulted container "ex-aao-container" out of: ex-aao-container, ex-aao-container-init (init)
# generated by crd
#
globalMaxSize=512MB

gtully:
added a system property via JAVA_OPTS to the init container to reference the mounted broker properties file to tie this together.

examples/artemis-basic-deployment.yaml
```diff
@@ -2,7 +2,11 @@ apiVersion: broker.amq.io/v1beta1
 kind: ActiveMQArtemis
 metadata:
   name: ex-aao
+  namespace: activemq-artemis-operator
 spec:
   deploymentPlan:
     size: 1
     image: placeholder
+  brokerProperties: {
+    globalMaxSize: 512MB,
```
gtully, line 11: we would need to do some extra work to support some sort of postfix or type modifier  - MB,GB etc.
probably best if the broker extracts those. I think that will be preferable to having just a raw number there.
the operator could also support these but I think it is broker config, rather than kube resource references etc.

gtully:
currently this configures a broker.properties independent of what is in the CR. it will be empty now by default. Any values form the CR brokerProperties map will override.
I am thinking that the operator may want to have some config in the broker.properties along with CR supplied values. 
I guess the operator can track those in an internal map, would it make sense to expose those too in the CR?

gaohoward (changes requested):

controllers/activemqartemis_reconciler.go
```diff
@@ -1955,6 +1958,15 @@ func NewPodTemplateSpecForCR(fsm *ActiveMQArtemisFSM) corev1.PodTemplateSpec {
...
 
+	// this depends on init container passing --java-opts to artemis create via launch.sh *and* it
+	// not getting munged on the way.
+	// REVISIT: should an existing value be respected here?
+	javaOpts := corev1.EnvVar{
+		Name:  "JAVA_OPTS",
+		Value: "-Dbroker.properties=/amq/extra/configmaps/broker-properties/broker.properties",
+	}
+	environments.Create(Spec.InitContainers, &javaOpts)
+
```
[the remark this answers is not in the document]
gtully: not sure, it seems all scripts respect JAVA_OPTS but none set them. So who or what else would set an initial value for JAVA_OPTS. it seems to me that we can be in control of the initial value for that variable.
Or am I missing some other operator that owns it?

examples/artemis-basic-deployment.yaml
gtully: support in https://github.com/apache/activemq-artemis/pull/3959

controllers/activemqartemis_reconciler.go
```diff
@@ -2018,6 +2030,58 @@ func NewPodTemplateSpecForCR(fsm *ActiveMQArtemisFSM) corev1.PodTemplateSpec {
...
+			APIVersion: "k8s.io.api.core.v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:                       configMapName.Name,
+			GenerateName:               "",
+			Namespace:                  configMapName.Namespace,
+			SelfLink:                   "",
+			UID:                        "",
+			ResourceVersion:            "",
+			Generation:                 0,
```
gtully, line 2052: what about theses defaults? Do we need to be wiser here?

controllers/activemqartemis_reconciler.go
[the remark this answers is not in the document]
gaohoward: my bad. the JAVA_OPTS is not changed in the script or elsewhere I know of. So it is fine.

gaohoward (approved):

gtully:
yes, the intent is to have it readonly, such that there is no runtime overhead checking for updates. not all of the broker can support reload and those bits that can require a reload from xml to trigger the required work.

these properties are applied after xml loading or xml reloading!

gtully:
working on the update scenario - have an issue with leaking config maps that I need to tidy.

andytaylor (approved):
lgtm

PR #154, opened 2022-02-23
#153 Broker created by OpenShift operator does not support "oc port-forward"

### LINKED ISSUE #153: Allow acceptors to bind to all interfaces
**Describe the bug**
Currently the operator configures broker acceptors to bind to pod's hostname. This makes 'port-forward'
not work. The port-forward is useful during dev/test (not in production). Allow acceptor to bind to all
interfaces (0.0.0.0) can solve the issue.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:
Issue #153
Added a bindToAllInterfaces option for acceptors. The acceptor will
bind to 0.0.0.0 if this option is set to true.


gtully:
I wonder if a boolean option is the best, would a comma list be better? I note that kubectl port-forward [1] has --address  as a comma list, where 0.0.0.0 is a valid option, as would localhost,w.x.y.z or localhost,CLUSTER_IP
All interfaces would work as an option, but it may be two open/exposed
[1] https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#port-forward/

gaohoward:
@gtully The comma sep list supplied in --address option is the listening host's address, we don't need to specify that in CR. The broker acceptors can either bind to the pod's hostname or 0.0.0.0 (only for port-forwad, which is most for testing, not productization env). I think that'll be sufficient right?

gtully:
my point is that is is not a binary decision, either or. The --addresses is an example where the list of bind addresses works for any case. It can contain a single entry for the all interfaces case or individual ips.

gaohoward:
> my point is that is is not a binary decision, either or. The --addresses is an example where the list of bind addresses works for any case. It can contain a single entry for the all interfaces case or individual ips.

what's the use case where use need the acceptor to bind multiple addresses in a pod?

gaohoward:
@gtully Gary I'll mrge this. Pls raise an issue if you think there is more to improve.

PR #165, opened 2022-03-04
ensure broker properties data is consistent across generations, unord…

gtully:
…ered map is sorted in the same way as for the hash such that update comparison works as expected


gtully:
# generated by crd\n#\ngen=1\nglobalMaxSize=64g\n,
needs to match:
# generated by crd\n#\nglobalMaxSize=64g\ngen=1\n,

the map needs to be sorted like for the hash!

gtully:
this is at least part of the reason the CI is failing atm.

PR #167, opened 2022-03-07
ENTMQBR-4468 Encrypting amq broker pod credentials from env variables

### LINKED ISSUE #166: Remove credential from broker container's env vars
**Describe the bug**
The default credentials are exposed in broker container as environment variables:
AMQ_USER/AMQ_PASSWORD/AMQ_CLUSTER_USER/AMQ_CLUSTER_PASSWORD
which are not ideal for security concerns. They can be removed as they are actually not used
in main container since init container was used.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:
Instead of encrypt the credentials we can simply remove them from the
main container (they are used only in init container)
Issue #166

### DIFF
--- a/controllers/activemqartemis_controller_test.go
+++ b/controllers/activemqartemis_controller_test.go
@@ -773,6 +773,49 @@ var _ = Describe("artemis controller", func() {
 			Eventually(checkCrdDeleted(crd.Name, namespace, &crd), timeout, interval).Should(BeTrue())
 		})
 	})
+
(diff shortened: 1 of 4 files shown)


gaohoward:
close this as it will break things that rely on those env vars

PR #171, opened 2022-03-11
Statefulset updates via processresources

gtully:


gtully:
next but it to figure out if TRIGGERED_ROLL_COUNT is necessary, it does lots of secret updates outside of processresources, which I want to avoid, it seems unnecessary now.
there may also be  an ordering issue with the Env vars. I won't get to that for a while, hence the share.

There is lots of tidying up to be completed also.

gtully:
fixed the issue with env var ordering and made the Probes optional via a pointer in the crd, to avoid overriding default values of an empty probe

gtully:
there is a problem with the secrets getting regenerated.... investigating

gtully:
this looking good now. the triggered_roll_count is a checksum and the existing secrets are respected.

RoddieKieley (approved):
I grabbed the latest and did some testing with various CR instances. The reconcile in general is much improved, although as discussed not all cases are covered quite yet. In particular I had problems with acceptors while the connector configuration operated as expected.

That being said this improvement has been outstanding for some time and after reviewing the built and executed code while also running test deployments of CR's I think we need to get it committed and move ahead with it.

PR #178, opened 2022-03-24
Added ENABLE_WEBHOOKS option

RoddieKieley:
Initial take at utilizing ENABLE_WEBHOOKS env var both at runtime and at build time. Not a full usage throughout however as some sections in the config/default/kustomization.yaml are commented out without an obvious way to use the kustomize tool to manipulate them even though the bases could be modified. Note that if kustomize edit add/remove base is used it removes the following commented entries which includes the [WEBHOOK] and [CERT-MANAGER] required sections.

Behaviour by default is to transparently apply ENABLE_WEBHOOKS=false at both build and runtime, i.e. you should be able to build, deploy and run without webhooks WITHOUT specifying ENABLE_WEBHOOKS manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,8 +83,16 @@ help: ## Display this help.
 
 ##@ Development
 
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+manifests: controller-gen kustomize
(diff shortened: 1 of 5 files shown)


gaohoward:
looking good

gaohoward:
@RoddieKieley can you merge this?

PR #184, opened 2022-03-30
remove state from broker/artemis controller

gtully:


gtully:
fyi https://github.com/artemiscloud/activemq-artemis-operator/issues/185 prevents the removing of state from the address controller.

controllers/activemqartemis_controller.go
```diff
@@ -51,22 +49,8 @@ func GetBrokerConfigHandler(brokerNamespacedName types.NamespacedName) (handler
 }
 
 func UpdatePodForSecurity(securityHandlerNamespacedName types.NamespacedName, handler common.ActiveMQArtemisConfigHandler) error {
-	success := true
-	for nsn, fsm := range namespacedNameToFSM {
-		if handler.IsApplicableFor(nsn) {
-			fsm.SetPodInvalid(true)
-			glog.Info("Need update fsm for security", "fsm", nsn)
-			if err, _ := fsm.Update(); err != nil {
```
gtully, line 59: This may need more work, and tests. The Update is a reconcile, but it may be what is necessary is a pod restart.

gtully:
going to sort that security update. but please review independent of that!

controllers/activemqartemis_controller.go
gtully: sorted this with events chan source to force reconcile of the matching cr's

controllers/scaling_state.go
```diff
@@ -1,126 +0,0 @@
...
-	// Log where we are and what we're doing
-	reqLogger := ctrl.Log.WithValues("ActiveMQArtemis Name", ss.parentFSM.customResource.Name)
-	reqLogger.Info("Updating ScalingState")
-
-	var err error = nil
-	var nextStateID int = ScalingID
-
-	currentStatefulSet := &appsv1.StatefulSet{}
-	ssNamespacedName := types.NamespacedName{Name: ss.parentFSM.GetStatefulSetName(), Namespace: ss.parentFSM.customResource.Namespace}
-	err = ss.parentFSM.r.Client.Get(context.TODO(), ssNamespacedName, currentStatefulSet)
```
gtully, line 79: this logic needs some review, if it is really necessary or if we can depend fully on external state (From the kube api server source of truth).  I  need to write some tests around scaling to see how it hangs together and see if I can demonstrate a breakage.
Some further research is necessary.... will revert to draft pending some more verification

RoddieKieley: I've been reviewing this PR further both to see what it would take to get it into main atm, as well as with a specific focus on the scaling / message migration aspect. An initial idea had been to take a "pause" during scaling so that we would be cautious while message migration was going on as well as while brokers were going up and down due to rolling updates. 

During this "pause" the operator would be in the "scaling" state where the most obvious impact was that any incoming updates were requeue'd on the kubernetes side for later reception and action. As well during this time if scaledown was occurring the scaledown controller would be taking note of the orphaned PVC and taking action via launching a drainer pod to migrate the messages to another broker pod.

While it is certainly still true that we want to be careful around message migration, and that the operator needs to understand what the current status of the broker cluster is, the implementation details of the state machine are possibly heavier than are required at the moment. We don't right now have a need to explicitly expose the transition into and out of the "state" as we actually do in the code.

I think a good balance might be to use a simplified reconciliation where at the beginning we note any discrepancy between the ready and set replicas and size in the custom resource instance and if scaling then requeue to kubernetes and finish. This would give a cautious behaviour, as now, for when brokers are starting and stopping but still allow for the simplification of the reconciliation.

gtully: I think the key question is to understand why we need to re queue events, why not just reconcile as normal. Both scaling and rolling updates are managed by kube. If we request stale down and scale up... it is the scaledown controller that has to deal with those requests and ensure that in the end we scale up!, probably after we scale down.

One concern I have is that the drain controller seems to poll for state, being triggered by the broker controller, rather than being event driven from kube independently. But I need to understand that better. In any event, if the drain controller gets the PV with rw-once, we should be safe. I guess the drain controller needs to be wise enough to know it should back off if it cannot get the PV.

my plan is to write an envTest with minikube that can exercise scale down and verify that it sill works with this PR, Ie: when the broker operator is stateless.

gaohoward (approved):

gtully:
I have re based, test good locally, however I saw this CI fail without good reason, so there is some timing related issue in play. It would be good if somone could try and reproduce a local failure. The ci just does make test.

gaohoward:
will try that on my local

gaohoward:
but didn't see the failure? all passed @gtully

gtully:
@gaohoward thanks. I think i have located the problem on the latest failed ci test - and have just sorted it, the eventually needed to include the get.

controllers/scaling_state.go
gtully: the new scaledown tests verify the functionality currently on main and work as expected with this branch.

PR #186, opened 2022-03-30
fix off by one is ss status - or pod name ordinal, closes #177

### LINKED ISSUE #177: CR status for stateful set ready is off by one w.r.t to the pod ordinal
**Describe the bug**

Off by one...

from the controller log - status update:
.... {"ActiveMQArtemis Name": "ex-aao", "info:": {"ready":["ex-aao-ss-1"]}}
Ready Count........................        {"ActiveMQArtemis Name": "ex-aao", "info:": 1}

then get pods:
ex-aao-ss-0                                            1/1       Running   0          45m

I would expect the name to match.

I note that the address controller observer is aware of the off by one:
https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/address_observer.go#L57

**To Reproduce**
deploy any trivial example with deployment size=1

gtully:


gtully:
this will need a rebase after https://github.com/artemiscloud/activemq-artemis-operator/pull/184 which resolves the flaky tests

PR #191, opened 2022-04-06
use a list in place of map for broker properties to respect order

gtully:


gtully:
the flakiness in the tests is resolved via: https://github.com/artemiscloud/activemq-artemis-operator/pull/184

PR #193, opened 2022-04-06
Added acceptor and connector options for store type and provider.

RoddieKieley:


api/v1beta1/activemqartemis_types.go
```diff
@@ -286,6 +286,14 @@ type AcceptorType struct {
 	SuppressInternalManagementObjects *bool `json:"suppressInternalManagementObjects,omitempty"`
 	// Whether to let the acceptor to bind to all interfaces (0.0.0.0)
 	BindToAllInterfaces *bool `json:"bindToAllInterfaces,omitempty"`
+	// Type of keystore being used; "JKS", "JCEKS", "PKCS12", etc. Default is "JKS"
+	KeyStoreType string `json:"keyStoreType,omitempty"`
```
gtully, line 290: I think we can live wothout the type...
the jdk can now peek/introspect in a file to figure out the type it needs to read.
does it provide some other value?

controllers/activemqartemis_controller_test.go
```diff
@@ -1645,7 +1644,291 @@ var _ = Describe("artemis controller", func() {
...
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			key := types.NamespacedName{Name: cr.Name, Namespace: namespace}
+			Eventually(func() bool {
+				_, ok := namespacedNameToFSM[key]
+				return ok
+			}, timeout, interval).Should(BeTrue())
+
+			fsm := namespacedNameToFSM[key]
```
gtully, line 1680: the tests need to restrict them selves to the kube api, they can be remote from the controller. They can use the namers but their own instance, they should not depend on the FSM. There are other examples to peek at in this file.

api/v1beta1/activemqartemis_types.go
gtully: peeking now to see if i can find a reference to this claim of introspection

controllers/activemqartemis_controller_test.go
gtully: eg: https://github.com/artemiscloud/activemq-artemis-operator/blob/bcad426a7dfba540a7fbfadf2d1e85ce13743656/controllers/activemqartemis_controller_test.go#L1994

gtully: some of the mods to the related tests are complete in: https://github.com/artemiscloud/activemq-artemis-operator/pull/184/files#diff-d78c1465cfcf5a1ea5e8e62663d05d50bc3d12175de63981c05441f8b31636dfR1703

api/v1beta1/activemqartemis_types.go
gtully: I did a quick sanity check on artemis main and it seems jks is the default type if nothing is configured. The JDK doing the right thing even if the type is configured wrong is https://bugs.openjdk.java.net/browse/JDK-8062552
In short, I think we can safely omit the keyStoreType

RoddieKieley: See below for commit addition with removal of keyStoreType.

controllers/activemqartemis_controller_test.go
RoddieKieley: I disagree that we should be keeping ourselves to the kube API in tests. It is also important to test the other structural code within the operator even if it is indirect supporting code for the main interaction with kubernetes and the brokers. What was your thinking behind a restriction for test code to only cover the kube API usage?

gaohoward: we'll try improve that in the future.

gtully: for the kube api, that is only for the envTests, those start a kube api server to behave like a sub set of real kube. They sould work whether kube is local or remote, hense just use the kube api.
Other tests (more unit tets) can for sure use internal apis... but the point of the envTests is to have a bit of kube present.

gtully: for sure it may make sense to have more test modules and split out envTests.

gtully: here is a good point of reference for envTests: https://github.com/kubernetes-sigs/controller-runtime/blob/master/FAQ.md#q-how-should-i-write-tests--any-suggestions-for-getting-started

gtully (approved):

PR #194, opened 2022-04-07
Added PodSecurityContext passthrough and test.

RoddieKieley:


gtully (commented):
looks good, it is a pointer, so can be absent and if present it totally controls the context

controllers/activemqartemis_reconciler.go
```diff
@@ -2052,6 +2054,18 @@ func configureAffinity(podSpec *corev1.PodSpec, affinity *corev1.Affinity) {
...
 
+func configurePodSecurityContext(podSpec *corev1.PodSpec, podSecurityContext *corev1.PodSecurityContext) {
+	clog.V(1).Info("Configuring PodSecurityContext")
+
+	if nil != podSecurityContext {
+		clog.V(5).Info("Incoming podSecurityContext is NOT nil, assigning")
+		podSpec.SecurityContext = podSecurityContext
+	} else {
+		clog.V(5).Info("Incoming podSecurityContext is nil, creating with default values")
+		podSpec.SecurityContext = &corev1.PodSecurityContext{}
```
gtully, line 2065: I wonder about this, we whack it.... but there is configPodSecurity where we supply any RunAsUser... I think that will get whacked by this.

gtully: If we don't reset on nil we won't whack anything.

gtully: https://github.com/artemiscloud/activemq-artemis-operator/blob/b8004a3a9afdf5c4624d442790b8972123b479a3/controllers/activemqartemis_reconciler.go#L2088 we may clash with this chap!

RoddieKieley: As below this is indeed the case. I moved the call and made a note in the source that the new fully featured PodSecurityContext would be set after the PodSecurity and override the runAsUser if provided in PodSecurity and then again in PodSecurityContext.

This however is not ideal, and ideally we would restructure the CRD to expand to using the actual PodSecurityContext which would also mean that the service account name would need to migrate. A conversion webhook could then be used to migrate one version to another.

gtully (approved):

PR #196, opened 2022-04-08
tidy up tests - repeated failure due to expectation of single secret …

gtully:
…which does not hold on shared test env with timing, kube gc is eventually responsilbe. Remove fmt output to keep console clean

### DIFF
--- a/controllers/activemqartemis_controller_test.go
+++ b/controllers/activemqartemis_controller_test.go
@@ -34,10 +34,11 @@ import (
 	"strconv"
 	"strings"
 
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


gtully:
something strange with vscode auto format, it disagrees with the last commit!
https://github.com/artemiscloud/activemq-artemis-operator/pull/196/files#diff-d78c1465cfcf5a1ea5e8e62663d05d50bc3d12175de63981c05441f8b31636dfR37

controllers/activemqartemis_controller_test.go
```diff
@@ -34,10 +34,11 @@ import (
 	"strconv"
 	"strings"
 
+	"time"
```
gtully, line 37: I wonder if there is something i need to update or some other ide needs to update to get this in sync.

gtully: that be the work of vscode :-)

PR #206, opened 2022-04-12
add test-mk target for manual verification against minikube, a deploy…

gtully:
… without webhooks is ncessary. Fix up flaky assertion that failed in mk


gtully:
minikube start
make deploy 
make test-mk

gaohoward (approved):

PR #220, opened 2022-04-21
add scaledown mk test

gtully:


gtully:
it is a little slow, it may be the default probes. takes just under 3mins to complete.
the lack of status updates makes it difficult to assert. scaledown needs clustering and clustering could also move the messages, but I cannot simply assert the status of the running set of the broker cr being 1 at the moment b/c updates are pause.
In any event this is a good start point.

gtully:
the service account does not seem to be necessary for mk, it may be that the controller client has admin privileges that filter through.

controllers/activemqartemisscaledown_controller.go
```diff
@@ -100,12 +101,7 @@ func (r *ActiveMQArtemisScaledownReconciler) Reconcile(ctx context.Context, requ
 	reqLogger.Info("====", "namespace:", namespace)
 	reqLogger.Info("====", "localOnly:", localOnly)
 
-	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
-	if err != nil {
-		reqLogger.Error(err, "Error building kubeconfig: %s", err.Error())
-	}
-
-	kubeClient, err = kubernetes.NewForConfig(cfg)
+	kubeClient, err = kubernetes.NewForConfig(r.Config)
```
gtully, line 104: it is not clear if any content from the first CR is relevant, in line with the rest of the controllers, the configuration form the manager should be sufficient but in a similar manner to the lack of service account and the lifecycle being dictated by the scaledown cr's, it is not 100% clear if there is any potential problem here.
For envTest, it is important that the config is passed in.

go.sum
```diff
@@ -455,6 +455,7 @@ github.com/mitchellh/iochan v1.0.0/go.mod h1:JwYml1nuB7xOzsp52dPpHFffvOCDupsG0Qu
 github.com/mitchellh/mapstructure v0.0.0-20160808181253-ca63d7c062ee/go.mod h1:FVVH3fgwuzCH5S8UJGiWEs2h04kUh9fWfEaFds41c1Y=
 github.com/mitchellh/mapstructure v1.1.2/go.mod h1:FVVH3fgwuzCH5S8UJGiWEs2h04kUh9fWfEaFds41c1Y=
 github.com/mitchellh/mapstructure v1.3.2/go.mod h1:bFUtVrKA4DC2yAKiSyO/QUcy7e+RRV2QTWOzhPopBRo=
+github.com/moby/spdystream v0.2.0 h1:cjW1zVyyoiM0T7b6UoySUFqzXMoqRckQtXwGPiBhOM8=
```
gtully, line 458: is this expected? and OK?

gtully: looks like duplication!

gtully: it is ok and expected. I pushed a further update with the result of go mod tidy.
there is a checksum for the any go.mod (the module dependencies) and the contents of the archive

gaohoward (approved):

PR #225, opened 2022-04-25
speed up scaledown test - exposed a missing exec on the broker proces…

gtully:
…s, causing it to loose sigterm from #195 and delay for the hard coded terminationGracePeriodSeconds of 60s


gtully:
There may still be a good case for making terminationGracePeriodSeconds configurable via the cr, but that needs further investigation to justify and test.
For sure, missing the exec for launch.sh is a bug, changing the default pid and loosing the signal handling.
Thanks @howardgao for the insights.

gaohoward (approved):

PR #233, opened 2022-05-13
menu builder for website

andytaylor:


andytaylor:
plus a couple of doc tweaks

PR #235, opened 2022-05-13
add roundtrip test with go amqp client and auto sharding via broker p…

gtully:
…roperties


gtully:
question, is there any simple way to limit the references to go amqp to testing? the mvn dependency scope equivalent? it may need the introduction of a testing module

gtully:
note, I pulled the setenv for the default images to the suite_test, as those are necessary for default or placeholder deployment in an external cluster. Had not noticed those before and one test was setting and leaking them.

gtully:
leaving at draft till we research multi module and a separate test module

gaohoward (approved):

PR #236, opened 2022-05-18
provide hard coded defaults for latest image names to avoid the need …

gtully:
…to env vars in tests - the env vars can happily override - resolve space in existing hard coded value in test such that test-mk can work!


gtully:
this fixes:
"[quay.io/artemiscloud/activemq-artemis-broker-init](http://quay.io/artemiscloud/activemq-artemis-broker-init): 1.0.3" invalid image name.

Issue #239, opened 2022-05-24
Addresses are not created in all the deployed broker pods if the address crs are deployed at the same time with security


gaohoward:
**Describe the bug**
When address CRs are deployed together with a security CR, not all addresses are created
on all the pods deployed

**To Reproduce**
1. deploy the operator
2. deploy broker CR with 5 pods
3. deploy 5 address Crs (5 queues) and 1 security CR
4. check that on some of the pods not all addresses are created.

**Platform (please complete the following information):**
 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**
Optional, use if not obvious. A clear and concise description of what you expected to happen.

**Screenshots**
Optional, add screenshots to help explain your problem.

**Additional context**
Optional. Add any other context about the problem here.

PR #240, opened 2022-05-24
#239 Addresses are not created in all the deployed

gaohoward:
broker pods if the address crs are deployed at the
same time with security

Added a test option "DEPLOY_OPERATOR" to test-mk
to allow tests run against operator deployed
in existing cluster


gtully:
do all of the existing tests work in this mode?
if not, it may be time to introduce a test/some_subset_if_tests directory where we do the full deploy scenario only on the controller tests that need it

gaohoward:
> do all of the existing tests work in this mode? if not, it may be time to introduce a test/some_subset_if_tests directory where we do the full deploy scenario only on the controller tests that need it

I'm not sure. The concern I can think of is that the tests will be take longer time to reach a expected state. So currently many tests may get timeout error. Will do some investigation

gtully:
yep. the timeout and interval need to be tweeked based on USE_EXTERNAL_CLUSTER. I guess we can do that after.

gtully (approved):

PR #241, opened 2022-05-31
set owner reference on drain controller pod, fix contention between s…

gtully:
…ecurity and artemis controller threads, use ginkgo v2 such that a label can be used for test-mk-do target, where controller is deployed to existing cluster.


controllers/activemqartemisaddress_controller_test.go
```diff
@@ -202,8 +195,126 @@ var _ = Describe("Address controller", func() {
 			}
 		})
 	})
+
+	Context("Address delete and scale down", func() {
```
gtully, line 199: @howardgao I started trying to reproduce the issue with deletion when there is a consumer from the the cluster connection, but this test with a cluster in play works fine on main. It did surface a panic when there was no queue and just an address in the CR, but worked otherwise.

gaohoward (approved):

gtully:
pushed update to avoid the dud storage class when USE_EXTERNAL_CLUSTER is in play. test-mk target now good.

PR #248, opened 2022-06-08
[247] Add JolokiaAgentOptions

brusdev:


gtully (approved):

gaohoward:
@brusdev Hi Dom can you rebase the PR?

gaohoward:
@brusdev @gtully we may need bump the CRD version. thoughts?

brusdev:
@gaohoward I have create a draft branch to bump ActiveMQArtemis CRD, see [bump_activemqartemis_crd_v1beta2](https://github.com/brusdev/activemq-artemis-operator/tree/bump_activemqartemis_crd_v1beta2). However I would avoid to bump ActiveMQArtemis CRD according with the Kubernetes [Changing the API](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) doc, see the [Adding a field](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-a-field) paragraph:
```
It is generally allowed to add new fields without changing the API version
```

gtully:
I think that makes sense. only bump when necessary. So long as new fields have good/sensible defaults in play that don't effect older deployments there is no real value in the version change.

PR #249, opened 2022-06-09
test to validate populateValidatedUser via crd properties

gtully:


controllers/activemqartemis_controller_test.go
```diff
@@ -2567,6 +2567,64 @@ var _ = Describe("artemis controller", func() {
...
+			}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
+			By("Consuming 1")
+
+			command = []string{"amq-broker/bin/artemis", "browser", "--user", "Jay", "--password", "activemq", "--url", "tcp://" + podWithOrdinal + ":61616", "--destination", "queue://DLQ", "--verbose"}
+
+			Eventually(func(g Gomega) {
+				stdOutContent := execOnPod(podWithOrdinal, crd.Name, defaultNamespace, command, g)
+
+				// ...ActiveMQMessage[ID:3a356c2a-e7e1-11ec-ae1c-5ec4168c91b2]:PERSISTENT/ClientMessageImpl[messageID=19, durable=true, address=DLQ,userID=3a356c2a-e7e1-11ec-ae1c-5ec4168c91b2,properties=TypedProperties[__AMQ_CID=3a2f9fc7-e7e1-11ec-ae1c-5ec4168c91b2,_AMQ_ROUTING_TYPE=1,_AMQ_VALIDATED_USER=73ykuMrb,count=0,ThreadSent=Producer ActiveMQQueue[DLQ], thread=0]]
```
gtully, line 2617: question?
what is odd is that there is some random string in the user field rather than the user name Jay that the producer uses. What can be the reason for that. 
Note: _AMQ_VALIDATED_USER=73ykuMrb
ideally I would assert the presence of _AMQ_VALIDATED_USER=Jay

gtully: with the config as it is, there is a guest login module configured with an auto generated user name for the guest name, that is what is being propagated

gtully:
added a second variant of the test with a valid user configured.

gaohoward (approved):

PR #252, opened 2022-06-24
Bump ActiveMQArtemis CRD v1beta2

brusdev:


gtully:
I think we should only update the version when we need to, ie: when we break existing CRs in the wild. For additions to the CR that have sensible defaults, there is no need to update the version. This follows what the kube apis do, and the more we reference kube types in our crd, the more normal it will be.

PR #261, opened 2022-07-01
fix intermittent failure due to missing reconcile event triggered fro…

gtully:
…m security contoller


gtully:
A flaky test pointed to the problem with a missing reconcile event:

[FAIL]artemis controller [It] deploy security cr while broker is not yet ready
/home/runner/work/activemq-artemis-operator/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:3411
[FAIL]security controller [It] reconcile after Broker CR deployed, verify force reconcile
/home/runner/work/activemq-artemis-operator/activemq-artemis-operator/controllers/activemqartemissecurity_controller_test.go:965

PR #282, opened 2022-08-09
[#281] If liveness probe is specified in the CR it is not applied to …

gaohoward:
…broker pods


gaohoward:
I'm going to merge it as it's a simple fix.

PR #287, opened 2022-08-24
[#286]  Operator need to check and cleanup PVC's owner references

### LINKED ISSUE #286: Operator need to check and cleanup PVC's owner references of any existing broker deployment
**Describe the bug**
Before fix for issue #278 the operator will add owner reference to PVCs. If users upgrades from such an installation,
their existing PVCs need to be checked and their owner references (if any) should be cleaned up to avoid potential data loss.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:
of any existing broker deployment

### DIFF
--- a/controllers/activemqartemis_controller_test.go
+++ b/controllers/activemqartemis_controller_test.go
@@ -30,6 +30,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/artemiscloud/activemq-artemis-operator/api/v1beta1"
 	brokerv2alpha4 "github.com/artemiscloud/activemq-artemis-operator/api/v2alpha4"
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/resources/environments"
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/resources/secrets"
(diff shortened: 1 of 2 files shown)


controllers/activemqartemis_reconciler.go
```diff
@@ -1229,6 +1232,33 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) deleteRequestedResource(customR
...
+	for i = 0; i < instance.Spec.DeploymentPlan.Size; i++ {
+		ordinalString := strconv.Itoa(int(i))
+		pvcKey := types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name + "-" + namer.CrToSS(instance.Name) + "-" + ordinalString}
+		pvc := &corev1.PersistentVolumeClaim{}
+		err := client.Get(context.TODO(), pvcKey, pvc)
+
+		if err == nil {
+			owners := pvc.OwnerReferences
+			if len(owners) > 0 {
+				log.Info("removing owner ref from pvc to avoid data loss")
```
RoddieKieley, line 1248: @gaohoward It's not for sure that this issue would cause data loss. It is only in the case that the PV has not been configured to 'Retain'. If it were configured to Retain then the PV would enter a state 'Released' when deleted where it was unavailable to be re-used, but the data would be sitting there awaiting manual human intervention.

So I think this statement is just a bit strong in the sense that we are removing the owner ref to avoid **potential** data loss, not guaranteed data loss. The difference is subtle but important I think.

controllers/activemqartemis_controller_test.go
```diff
@@ -682,6 +683,69 @@ var _ = Describe("artemis controller", func() {
 		})
 	})
 
+	Context("PVC upgrade owner reference test", Label("pvc-owner-reference-upgrade"), func() {
+		It("faking a broker deployment with owned pvc", func() {
+			crd := generateArtemisSpec(defaultNamespace)
+			crd.Spec.DeploymentPlan.Size = 1
+			crd.Spec.DeploymentPlan.PersistenceEnabled = true
+
+			if os.Getenv("USE_EXISTING_CLUSTER") == "true" {
```
RoddieKieley, line 692: @gaohoward I can't recall, does this log when this condition is not met and the bulk of the test is actually not run? Does it show as a skip, or a pass? If the latter making a note via a logged message in the else case might be useful although obviously not necessary.

RoddieKieley (changes requested):

controllers/activemqartemis_controller_test.go
gaohoward: +1, the USE_EXISTING_CLUSTER means it only run if an existing cluster (like minikube). Otherwise it just sliently skip the test. (see test-mk target). I'll add a log message.

controllers/activemqartemis_reconciler.go
gaohoward: good point. +1

controllers/activemqartemis_reconciler.go
```diff
@@ -1229,6 +1232,33 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) deleteRequestedResource(customR
...
+	for i = 0; i < instance.Spec.DeploymentPlan.Size; i++ {
+		ordinalString := strconv.Itoa(int(i))
+		pvcKey := types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name + "-" + namer.CrToSS(instance.Name) + "-" + ordinalString}
+		pvc := &corev1.PersistentVolumeClaim{}
+		err := client.Get(context.TODO(), pvcKey, pvc)
+
+		if err == nil {
+			owners := pvc.OwnerReferences
+			if len(owners) > 0 {
+				log.Info("removing owner ref from pvc to avoid potential data loss")
```
gtully, line 1248: we need to verify that the owner ref points to the CR, otherwise we may whack an owner reference set by the beta feature at: https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/

gtully: checking equality on the GUID will do it: not from here how the owner ref was set in the ss template:
https://github.com/artemiscloud/activemq-artemis-operator/pull/279/files#diff-32e4c985bb650a7028cf649306e140d1cf176b91d4f36dbfbce53083d0f63bd6L2047

gaohoward: ok, good point.

controllers/activemqartemis_reconciler.go
```diff
@@ -1229,6 +1232,44 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) deleteRequestedResource(customR
...
+		pvcKey := types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name + "-" + namer.CrToSS(instance.Name) + "-" + ordinalString}
+		pvc := &corev1.PersistentVolumeClaim{}
+		err := client.Get(context.TODO(), pvcKey, pvc)
+
+		if err == nil {
+			totalLen := len(pvc.OwnerReferences)
+			if totalLen > 0 {
+				found := -1
+				newOwnerReferences := make([]metav1.OwnerReference, 0)
+				for i, oref := range pvc.OwnerReferences {
```
gtully, line 1250: This will work.

small things, but I would not reuse i, just use a bool for found, and I don't see totalLen reused, so just use len() in the if statement. Just to make it easier to understand.

gaohoward: +1

gtully (approved):

PR #290, opened 2022-08-25
fix examples dir location reference

hodrigohamalho:


brusdev:
@hodrigohamalho thanks for your contribution

Issue #295, opened 2022-08-31
Allow setting []corev1.EnvVar for our statefull set from the CR


gtully:
From TZ to JDK_JAVA_OPTIONS there are numerous reasons why a user may wish to augment the environment that the broker has, but currently there is no easy way to do that short of using an init container to manipulate the launch scripts.

Ideally a user should be able to specify, for example:
> crd.Spec.Env = []corev1.EnvVar{			
>                      {Name: "TZ", Value: "en_IE"},
>			{Name: "JDK_JAVA_OPTIONS", Value: "-XshowSettings:system"},
>}

PR #296, opened 2022-08-31
[#295] support envVar[] in the CR

gtully:


gaohoward (approved):

gaohoward:
@gtully @brusdev @RoddieKieley We should probably bump the crd version first? i.e. merge this pr befor that:
https://github.com/artemiscloud/activemq-artemis-operator/pull/252

gtully:
For additions, we can stick with the CRD version. We really only have to bump when we are breaking existing CRs.

gaohoward:
@gtully @brusdev @RoddieKieley ok. I'd then suggest we cut a new tag before merging any additions.

controllers/activemqartemis_reconciler.go
```diff
@@ -1633,13 +1631,12 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
...
 	// this depends on init container passing --java-opts to artemis create via launch.sh *and* it
-	// not getting munged on the way.
-	// REVISIT: should an existing value be respected here, only when we expose JAVA_OPTS in the CR?
+	// not getting munged on the way. We CreateOrAppend to any value from spec.Env
 	javaOpts := corev1.EnvVar{
 		Name:  "JAVA_OPTS",
 		Value: "-Dbroker.properties=/amq/extra/configmaps/" + brokerPropertiesConfigMapName + "/broker.properties",
 	}
-	environments.Create(podSpec.InitContainers, &javaOpts)
+	environments.CreateOrAppend(podSpec.InitContainers, &javaOpts)
```
brusdev, line 1639: as final user maybe I would expect see my env vars with the same value defined in the CR
as operator maybe I should avoid to change my env vars
a solution could be that the operator ignore internal vars as JAVA_OPTS and include a prefix/suffix env var in the internal env vars. i.e. JAVA_OPTS="$JAVA_OPTS_PRE ... $JAVA_OPTS_SUF"
so users are not able to set JAVA_OPTS but they are able to set $JAVA_OPTS_PRE or $JAVA_OPTS_SUF

gtully: I think JAVA_OPTS is some what of a special case b/c it is interpreted by the init container, artemis create command and is used for brokerProperties. I think we can in the future configure the brokerProperties url in some better way and then not have to treat is specially w.r.t append. 
But for sure I don't want to encourage the use of env JAVA_OPTS. it is particular to the artemis launch scripts... way too low level. Ading PRE POST special cases would make it worse imho.
To configure the JVM, use JDK_JAVA_OPTIONS

brusdev (approved):

gtully:
going to add some doc, and noticed the name annotation is wrong!

PR #297, opened 2022-09-06
test for spec.version update

gtully:


gtully:
@gaohoward how best to manage the version, to keep this test fresh? I need some ideas!
currently the tests use Latest, as the default image url, but can we be sure that latest-1 will exist?

gaohoward:
> @gaohoward how best to manage the version, to keep this test fresh? I need some ideas! currently the tests use Latest, as the default image url, but can we be sure that latest-1 will exist?

I don't have better idea. The latest-1 should exist, if not the test should fail, which is not a bad thing.

PR #302, opened 2022-09-14
Version refactoring

brusdev:


gtully (approved):
that does seem to clean up a lot, and move to images that more closely aline with broker versions

PR #303, opened 2022-09-15
[#32] allow user to config logging for broker

### LINKED ISSUE #32: allow user to config logging for broker
Currently users can not configure logging properties and that may not suit specific needs.
Such as when a broker continues to log large amount of warning messages over a period of
time and causes PV to fill up.

Users can change the broker logging configuration by providing their own in a configmap or secret. The name of the configmap
or secret must have the suffix **-logging-config**. There must be one entry in the configmap or secret. The key of the entry
must be **logging.properties** and the value must of the full content of the logging configuration. (The broker is using slf4j with
log4j2 binding so the content should be log4j2's configuration in Java's properties file format).

Then you need to give the name of the configmap or secret in the broker custom resource. For example

`for configmap`
```yaml
kind: ActiveMQArtemis
metadata:
  name: broker
spec:
spec:
  deploymentPlan:
    size: 1
    image: placeholder
    extraMounts:
      configMaps:
      - "my-logging-config"
```
`for secret`
```yaml
kind: ActiveMQArtemis
metadata:
  name: broker
spec:
spec:
  deploymentPlan:
    size: 1
    image: placeholder
    extraMounts:
      secrets:
      - "my-logging-config"
```

**Additional context**

gaohoward:
To use custom logging configuration for broker user has to make a configmap with the following rules:
1. The name of the configmap takes the form [cr-name]-logging-config
2. The configmap has one entry, the key name must be "logger.properties", the value is the whole content of the logging configuration (normally lines of key=value)
The operator doesn't care about the content of the configmap. It only finds the configmap and mounted to a location
/amq/extra/configmaps/<configmap name>/logger.properties
The operator then also export an env var LOGGER_PROPERTIES that points to the above mounted location.

When the broker image is launched it checks the env var LOGGER_PROPERTIES and creates a sym link to ${instanceDir}/etc/logging.properties. if the var is defined. Otherwise it takes the default logging.properties and changes all logging to console output (unless persistence is enabled)

If the configmap is created after the broker has been deployed, all the pods will be restarted to update the logging configuration.

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -210,7 +211,8 @@ func validate(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Cli
 
 func validateExtraMounts(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, scheme *runtime.Scheme) (*metav1.Condition, error) {
 	for _, cm := range customResource.Spec.DeploymentPlan.ExtraMounts.ConfigMaps {
-		found, err := validateExtraMount(cm, customResource.Namespace, &corev1.ConfigMap{}, client, scheme)
+		configMap := corev1.ConfigMap{}
+		found, err := validateExtraMount(cm, customResource.Namespace, &configMap, client, scheme)
 		if err != nil {
 			return nil, err
 		}
@@ -222,9 +224,21 @@ func validateExtraMounts(customResource *brokerv1beta1.ActiveMQArtemis, client r
 				Message: fmt.Sprintf("Missing required configMap %v", cm),
 			}, nil
 		}
+		//validate logging
+		if strings.HasSuffix(cm, loggingConfigSuffix) {
+			if _, ok := configMap.Data["logging.properties"]; !ok {
+				return &metav1.Condition{
+					Type:    common.ValidConditionType,
+					Status:  metav1.ConditionFalse,
+					Reason:  common.ValidConditionFailedReason,
+					Message: fmt.Sprintf("Logging configmap %v must have key logging.properties", cm),
+				}, nil
+			}
+		}
 	}
 	for _, s := range customResource.Spec.DeploymentPlan.ExtraMounts.Secrets {
-		found, err := validateExtraMount(s, customResource.Namespace, &corev1.Secret{}, client, scheme)
+		secret := corev1.Secret{}
+		found, err := validateExtraMount(s, customResource.Namespace, &secret, client, scheme)
 		if err != nil {
 			return nil, err
 		}
@@ -235,6 +249,17 @@ func validateExtraMounts(customResource *brokerv1beta1.ActiveMQArtemis, client r
 				Reason:  common.ValidConditionMissingResourcesReason,
 				Message: fmt.Sprintf("Missing required secret %v", s),
 			}, nil
+		}
+		//validate logging
+		if strings.HasSuffix(s, loggingConfigSuffix) {
+			if _, ok := secret.Data["logging.properties"]; !ok {
+				return &metav1.Condition{
+					Type:    common.ValidConditionType,
+					Status:  metav1.ConditionFalse,
+					Reason:  common.ValidConditionFailedReason,
+					Message: fmt.Sprintf("Logging secret %v must have key logging.properties", s),
+				}, nil
+			}
 		}
 	}
 	return nil, nil
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -67,6 +67,7 @@ const (
 	defaultLivenessProbeInitialDelay = 5
 	TCPLivenessPort                  = 8161
 	jaasConfigSuffix                 = "-jaas-config"
+	loggingConfigSuffix              = "-logging-config"
 )
 
 var defaultMessageMigration bool = true
@@ -206,7 +207,7 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessStatefulSet(customResour
 	}
 
 	log.Info("Reconciling desired statefulset", "name", ssNamespacedName, "current", currentStatefulSet)
-	currentStatefulSet, err = reconciler.NewStatefulSetForCR(customResource, namer, currentStatefulSet)
+	currentStatefulSet, err = reconciler.NewStatefulSetForCR(customResource, namer, currentStatefulSet, client)
 	if err != nil {
 		reqLogger.Error(err, "Error creating new stafulset")
 		return nil, err
@@ -1556,7 +1557,7 @@ func MakeContainerPorts(cr *brokerv1beta1.ActiveMQArtemis) []corev1.ContainerPor
 	return containerPorts
 }
 
-func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customResource *brokerv1beta1.ActiveMQArtemis, namer Namers, current *corev1.PodTemplateSpec) (*corev1.PodTemplateSpec, error) {
+func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customResource *brokerv1beta1.ActiveMQArtemis, namer Namers, current *corev1.PodTemplateSpec, client rtclient.Client) (*corev1.PodTemplateSpec, error) {
 
 	reqLogger := ctrl.Log.WithName(customResource.Name)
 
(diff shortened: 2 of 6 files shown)


gaohoward:
ref: https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/pull/45

gtully:
this is great. It has got me thinking...

I am wondering if the operator should _always_ use the config map. Use one that exists or create one with the contents from the image and the broker always runs with a logging config system property pointing to the mounted config map projected volume.

In that way, editing the logging configuration can be done with config map editing. If the logger supports reload - then that too can happen via the volume projection.
The operator can own the config map or not, but it any event it won't modify the contents in any way, once it is created.

One problem would be the deletion of custom edits. To avoid that, if the config map does not exist before the artemis cr, then the owner reference would need to be dropped on edit. I wonder if that is possible? AFAIK it is not normal on an update.

If the custom changes are put in a config map in advance, then the operator won't own it and all will be good.

gaohoward:
> 

I can see the benefit of always using a configmap, but if we let operator to fill in the contents the operator need a way to get the contents from the image without knowing the logging config details (just treat the content like an opaque data), don't know how to do it. and yes that will bring in the ownership problem.  

I think it simpler that the operator doesn't own the configmap at all. It totally depends on user to manage it. (separate of concerns)

gtully:
I think I agree. If the operator creates a config map, it will want to own it (being problematic for user updates) and we probably want to allow the initial contents to be configured! In that case, we will want the CR to be the source of truth. But I don't want to see the entire content of a logging.properties in the artemis CR!

Even if we magic it from the image, it won't be much better b/c we will want to reconcile against that.

It seems we need to first fix our images, such that the default logging configuration is a good fit. Does **_not_** use persistence and does just log to stdout.

For customisation, I wonder how far we need to go? 

We have extra mounts, we can just have that reference a user created config map and use [CR.env](https://github.com/artemiscloud/activemq-artemis-operator/issues/295) = { JDK_JAVA_OPTIONS=-Dlogging.properties=/var/extramounts/configmaps/user_map/my.custom_logging.properties } and leave it free form?

if the file name needs to be different or the system property needs to be updated, that will be in the CR.

That will override anything that is set from the artemis.profile or any other script as the JDK reads and evaluates JDK_JAVA_OPTIONS

The same approach would work if the config needed to be in a secret, just the volume mount point would different in the JDK_JAVA_OPTIONS.
The operator is largely out of the way.

gaohoward:
Yes I think it's a good idea using CR.env.

gaohoward:
Maybe use JDK_JAVA_OPTIONS or DEBUG_ARGS directly in operator directly is not very flexible for users who also want to define more other options using this env var.

gaohoward:
My current thought is that, if a user wants custom logging, there are 2 options:
Option 1 - create a configmap (or secret) named *<cr-name>-logging-config* with custom logging config in it. Then the operator can automatically picks that up and setup mounts and append the -Dlogging.configuration=<url> to DEBUG_ARGS env var. (that would be logger specific option, we may put it to image size to do it)
Option 2 - use can use a different configmap other than default by add an env var to the CR.Env,
Env['LOGGING_CONFIG_NAME']: my-logging-configmap-name (or secret name)
Then the operator detects this and retrieves/mounts this configmap (instead of the default one in option 1), and the rest is same as option 1.

gaohoward:
hmm, **option 2** seems not right, the cr.Env is supposed to use inside the broker's container, not operaotr's

gaohoward:
So instead of using DEBUG_ARGS directly, the operator exposes a DEBUG_ARGS_LOGGING that carries the value of custom logging java option (-Dlogging.configuration=file:/...) and on the container image side append to the DEBUG_ARGS as in launch.sh

`  if [[ "${DEBUG_ARGS_LOGGING}" == *"-Dlogging.configuration"* ]]; then
    DEBUG_ARGS="$DEBUG_ARGS ${DEBUG_ARGS_LOGGING}"
    echo "There is a custom logger configuration defined in DEBUG_ARGS_LOGGING: ${DEBUG_ARGS}"
    export DEBUG_ARGS
  else
    echo "Using default logging configuration"
    defaultLoggingConfigFile=${instanceDir}/etc/logging.properties
    sed -i "s/logger.handlers=FILE,CONSOLE/logger.handlers=CONSOLE/g" $defaultLoggingConfigFile
    sed -i "s/handlers=AUDIT_FILE/handlers=CONSOLE/g" $defaultLoggingConfigFile
  fi`

gtully:
DEBUG_ARGS is just a bad name, what we need is JDK_JAVA_OPTIONS_APPEND b/c JDK_JAVA_OPTIONS is prepend! I wonder if such an option exists in the jdk image. It seems a very reasonable ask!
DEBUG_ARGS is in the right place, but badly named. 
I don't think there is a problem is a user wants to set two system properties or n system properties to override. They want to take control of the logging configuration, it is best to make that explicit - and not have some conditional logic is a launch.sh that a user needs to be aware of.
The real problem is the name of debug args!

gaohoward:
yeah, don't know why we have this called DEBUG_ARGS. :)

gaohoward:
@gtully Updated to be aligned with new JAVA_ARGS_APPEND options.

gaohoward:
@gtully Updated to use JAVA_ARGS_APPEND directly

gtully (commented):
One thought - peeking at the jaas-config support. https://github.com/artemiscloud/activemq-artemis-operator/blob/be4eba9d141fc15f5fce1362dd1d88010d504c5a/controllers/activemqartemis_reconciler.go#L1885

that avoids the need for a config map watch b/c the extra config is explicitly indicated through extra mounts. I think that is nice..
also - it can be a secret or cm, which is good if the logging config needs to be hidden for some reason.

I love the tests btw!

gtully:
I think from a consistency point of view, and avoiding the config map watch, it may be better to respect a "-logging-config" suffix via extra mounts. I wonder if there is some use case that wont fit into that pattern?

gtully:
final thought, can we validate the content of the supplied config map - ie: validate and reject if there is no logging.properties in there?

gaohoward:
Good point, will add it. Thx

gaohoward:
@gtully updated PR.

controllers/activemqartemis_controller.go
```diff
@@ -223,6 +227,19 @@ func validateExtraMounts(customResource *brokerv1beta1.ActiveMQArtemis, client r
 			}, nil
 		}
 	}
+	loggingConfigMapName := customResource.Name + "-logging-config"
```
gtully, line 230: If we use "-logging-config" as a special postfix, we can just look for entries matching that in the extraMounts and use that as the trigger to populate the env var. We don't need the crd.name prefix any more, just something ending with "-logging-config".
In this way, the reference is explicit in the CR and we don't need to watch config maps any more. It will be simpler.

controllers/activemqartemis_controller.go
```diff
@@ -223,6 +227,19 @@ func validateExtraMounts(customResource *brokerv1beta1.ActiveMQArtemis, client r
...
+	loggingConfigMap := corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: loggingConfigMapName, Namespace: customResource.Namespace}, &loggingConfigMap)
+	if err == nil {
+		if _, ok := loggingConfigMap.Data["logging.properties"]; !ok {
+			return &metav1.Condition{
+				Type:    common.ValidConditionType,
+				Status:  metav1.ConditionFalse,
+				Reason:  common.ValidConditionFailedReason,
+				Message: fmt.Sprintf("Logging configmap %v must have key logging.properties", loggingConfigMapName),
```
gtully, line 239: validation is great. nice!

controllers/activemqartemis_reconciler.go
```diff
@@ -1630,6 +1634,23 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
 	configMapsToCreate := []string{brokerPropertiesConfigMapName}
 	configMapsToCreate = append(configMapsToCreate, customResource.Spec.DeploymentPlan.ExtraMounts.ConfigMaps...)
 
+	loggingConfigMapName := customResource.Name + "-logging-config"
+	customLogging := false
+
+	_, err := configmaps.RetrieveConfigMap(namespacedName.Namespace, loggingConfigMapName, client)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
gtully, line 1642: requiring the entry"xx-logging-config" in extra mounts will give us this validation via extra mounts validation.

controllers/activemqartemis_controller.go
```diff
@@ -310,7 +327,33 @@ func (r *ActiveMQArtemisReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&brokerv1beta1.ActiveMQArtemis{}).
 		Owns(&appsv1.StatefulSet{}).
-		Owns(&corev1.Pod{})
+		Owns(&corev1.Pod{}).
+		Watches(
```
gtully, line 331: I think we can avoid the watch by requiring an entry "xx-logging-config" in extra mounts. This makes it an explicit attribute of the CR.

gtully (changes requested):

controllers/activemqartemis_controller.go
gaohoward: make sense. It'll be just like how xx-jaas-config works.

controllers/activemqartemis_controller.go
gaohoward: yeah agreed

gaohoward:
@gtully also added secret support to be consistent with jaas-config.

pkg/resources/configmaps/configmap.go
```diff
@@ -0,0 +1,46 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func RetrieveConfigMap(namespace string, configMapName string, client client.Client) (*corev1.ConfigMap, error) {
+	if client == nil {
+		// this could happen in some tests
+		return nil, errors.New("caller doesn't provide a client")
```
gtully, line 16: I wonder if this can still happen now that the presence is explicit in extraMounts, it seems odd that it could

gtully (approved):

gtully:
we may want to update the linked issue - so that it can describe the solution - and I guess we need some doc on this behaviour

gaohoward:
yeah I will add the doc.

pkg/resources/configmaps/configmap.go
gaohoward: I'll check it out.

gaohoward: yep no longer needed.

PR #304, opened 2022-09-16
tests for extra mounts with config map, update and logging configurat…

gtully:
…ion override


gtully:
https://github.com/apache/activemq-artemis/pull/4223 will help a little with this use case

PR #307, opened 2022-09-21
[#306] Print commit hash and build timestamp at startup

brusdev:


gtully (approved):
this is a great addition :-)

Issue #315, opened 2022-09-29
Cr labels are applied to StateFull set in error


gtully:
We apply custom labels from the CR to the Statefull set in error, and not just to the Pod Template in the spec.
By default there are two auto labels that match the CR name, those shoud be applied across the board, however only the CR lables need to be applied to the Pod template.

To Reproduce
create a CR with crd.Spec.DeploymentPlan.Labels["bla"] = "bla"
and note the SS metadata has these values. Then update and note that the SS needs to be deleted to rollout changes b/c the lables are part of the ss.Spec.Selector that does not support update. 

Expected behavior
The cr custom labels should not be in the statefull set other than the spec PodTemplate. They are duplicated in error

follows on from: https://github.com/artemiscloud/activemq-artemis-operator/issues/312

PR #316, opened 2022-09-29
[#315] only set CR labels on PodTemplate, not on Stateful set

gtully:


brusdev:
@gtully I pushed another commit to skip CR labels not matching statefulset selector labels, WDYT?

brusdev (approved):

Issue #318, opened 2022-10-04
Addresses / Queues are not created when the broker scales up


brusdev:
- Install operator for all namespaces
- Create project amq-test-1
- Apply the supplied YAML with the broker/addresses definition on the namespace amq-test-1 and wait some time
- Scale down the broker on namespace amq-test-1 from 3 to 2 and wait some time
- Scale up the broker on namespace amq-test-1 from 2 to 3 and wait some times.
- Check for queues on the 3 broker pods on namespace amq-test-1. All queues available on all pods
- Create project amq-test-2
- Apply the supplied YAML with the broker/addresses definition on the namespace amq-test-2 and wait some time
- Scale down the broker on namespace amq-test-2 from 3 to 2 and wait some time
- Scale up the broker on namespace amq-test-2 from 2 to 3 and wait some times.
- Check for queues on the 3 broker pods on namespace amq-test-2. Queues available only on pod 0 and 1

PR #319, opened 2022-10-04
[#318] Fix address creation for new pods

brusdev:


gtully (approved):
the proxy looks like a nice addition to avoid the need to deploy the operator to kube. Will be easier to debug. Nice work :-)

Issue #323, opened 2022-10-07
Make brokerProperties mutable


gtully:
Currently  updates to the CR brokerProperties require a new config map and a broker restart via a ss update.

The broker supports configuration reload, and recently has added the location of additional properties to the watch list. This will allow an config map projection update to be applied to a running broker. 

The broker configuration decides if updates will be applied via: configurationFileRefreshPeriod which must be enabled. A simple way to do this is via brokerProperties - configurationFileRefreshPeriod=5000
 
**Describe the solution you'd like**
Changes to the CR brokerProperties attribute should be applied to mutable config map

Any existing deployed immutable config map will be respected and left unchanged until the SS is recreated.

PR #324, opened 2022-10-07
[#323] make brokerProperties mutable - and lots of test suite tidy

gtully:


controllers/activemqartemis_reconciler.go
```diff
@@ -1268,6 +1273,34 @@ func checkExistingPersistentVolumes(instance *brokerv1beta1.ActiveMQArtemis, cli
 	}
 }
 
+var orderedTypes *([]reflect.Type)
```
gaohoward, line 1276: I think the pointer is not necessary. we can just use the slice type directly.

gaohoward (approved):

PR #331, opened 2022-10-10
add conditions to API

### LINKED ISSUE #330: Add conditions to Broker status
ActiveMQArtemis resources are lacking a set of conditions that properly reflect the latest observations and current state.

To add the an array of `metav1.Condition` including a `Ready` condition that is `true` when all the other conditions are satisfied. For the sake of simplicity it should be good if we took `True` as the desired state of any condition.

**Additional context**
Given the upcoming work around the Broker synchronisation with ConfigMaps the availability of conditions will help propagating the Broker status to the custom resource

ruromero:
Fix https://github.com/artemiscloud/activemq-artemis-operator/issues/330

Signed-off-by: ruromero <rromerom@redhat.com>


controllers/activemqartemis_controller.go
```diff
@@ -143,6 +146,8 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
 
 		UpdatePodStatus(customResource, r.Client, request.NamespacedName)
 
+		err = UpdateCRStatus(customResource, r.Client, request.NamespacedName)
```
gtully, line 149: I think the retry on error logic is broken because it used a 10h default retry period (in error). That may need to be fixed to make any retry on status error visible in a sensible time frame.

controllers/activemqartemis_controller.go
```diff
@@ -232,3 +237,30 @@ func (r *ActiveMQArtemisReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	err := client.Get(context.TODO(), namespacedName, current)
+	if err != nil {
+		clog.Error(err, "unable to retrieve current resource", "ActiveMQArtemis", namespacedName)
+		return err
+	}
+
+	if !reflect.DeepEqual(current.Status.PodStatus, cr.Status.PodStatus) {
+		return resources.UpdateStatus(client, cr)
+	}
+	if len(current.Status.Conditions) != len(cr.Status.Conditions) {
```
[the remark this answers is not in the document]
gtully: ah sorry, my bad, I now see the deepEqual is on the nested PodStatus and not the Status!

controllers/activemqartemis_controller.go
gtully: 30s is better. thanks!

gtully:
there was a test assertion for the existing default sync period. sorted in main.

for future reference, i missed to suggest, put the issue id in the commit message [#002], makes it easier to track issues with commits.

PR #339, opened 2022-10-14
[#338] Ingress doesn't work for console exposure with ssl enabled

### LINKED ISSUE #338: Ingress doesn't work for console exposure with ssl enabled
**Describe the bug**
When console.expose is true and sslEnabled is true for console,
the created ingress is broken.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:
Added test for both ingress and route


controllers/activemqartemis_controller_test.go
```diff
@@ -228,6 +229,148 @@ var _ = Describe("artemis controller", func() {
...
+					brokerKs, ferr := os.ReadFile("../test/resources/broker.ks")
+					Expect(ferr).To(BeNil())
+					clientTs, ferr := os.ReadFile("../test/resources/client.ts")
+					Expect(ferr).To(BeNil())
+
+					certData["broker.ks"] = brokerKs
+					certData["client.ts"] = clientTs
+					stringData["keyStorePassword"] = "password"
+					stringData["trustStorePassword"] = "password"
+
```
gtully, line 270: By("deploying well known secret name that the operator will look for")

is it correct that that secret is only referenced by the operator?

gtully: I don't see it in the CR that is.

gtully (commented):
does it make sense to have a variant of the test where we supply a different secret name via the CR?  I wonder what is the use case for that? Is it really necessary.

If it is, then we should test that scenario also, or possibly deprecate.

controllers/activemqartemis_reconciler.go
```diff
@@ -1369,12 +1369,18 @@ func MakeVolumes(customResource *brokerv1beta1.ActiveMQArtemis, namer Namers) []
 	}
 
 	if customResource.Spec.Console.SSLEnabled {
-		secretName := namer.SecretsConsoleNameBuilder.Name()
-		if "" != customResource.Spec.Console.SSLSecret {
-			secretName = customResource.Spec.Console.SSLSecret
+		isOpenshift, _ := environments.DetectOpenshift()
+		if customResource.Spec.Console.Expose && !isOpenshift {
+			//if console is exposed on k8s, the ssl is enabled at ingress
+			clog.V(1).Info("Not to make volumes for ssl console exposure on k8s")
```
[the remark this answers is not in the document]
gaohoward: +1

controllers/activemqartemis_controller_test.go
gaohoward: The operator uses the secret to mount the certs into pod for broker to set up the secured console

gaohoward: The console has a option for specifying the secret name:
https://github.com/artemiscloud/activemq-artemis-operator/blob/main/api/v1beta1/activemqartemis_types.go#L347
If not set operator looks for the default name which is <cr-name>-console-secret

gaohoward:
> does it make sense to have a variant of the test where we supply a different secret name via the CR? I wonder what is the use case for that? Is it really necessary.
> 
> If it is, then we should test that scenario also, or possibly deprecate.

Yes it is valid case where user uses a secret whose name is not the default. Will add one more test for that.

gtully (approved):

Issue #340, opened 2022-10-17
Error setting the application label


brusdev:
Setting the application label causes the following error:
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: ex-aao
spec:
  deploymentPlan:
    size: 1
    image: placeholder
    labels:
      application: test
```
```
ERROR    package k8s_actions    Failed to create new *v1.StatefulSet    {"ActiveMQArtemis Name": "ex-aao-ss", "Namespace": "test-namespace", "error": "StatefulSet.apps \"ex-aao-ss\" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{\"ActiveMQArtemis\":\"ex-aao\", \"application\":\"test\"}: `selector` does not match template `labels`"}
```

brusdev:
Resolved in 1.0.6

PR #341, opened 2022-10-17
[#340] Add labels validation

brusdev:


api/v1beta1/activemqartemis_types.go
```diff
@@ -224,7 +224,7 @@ type DeploymentPlanType struct {
 	// Whether or not to install the artemis metrics plugin
 	EnableMetricsPlugin *bool               `json:"enableMetricsPlugin,omitempty"`
 	Tolerations         []corev1.Toleration `json:"tolerations,omitempty"`
-	//custom labels provided in the cr
+	// Assign labels to a broker pod, the keys ActiveMQArtemis and application are not allowed
```
gtully, line 227: if it is ok to use ` or ' in a comment, then 
 keys `ActiveMQArtemis` and `application` are not allowed

would make it more clear. But only a small point!

brusdev: quote added

gtully (approved):

PR #346, opened 2022-10-17
Track Broker Properties status

### LINKED ISSUE #345: Check properties update status
The broker.properties sha is exposed in Jolokia. The fact that the sha in the properties match the generated sha from the defined properties means that the broker is in sync.

A condition should be used to reflect these states. The operator creates a configMap containing the properties that the broker will read, that means the broker has to be synced. The broker will periodically query the Jolokia endpoint until the broker status finally shows the matching sha from the status. After that, the condition will be updated.

Depends on https://github.com/artemiscloud/activemq-artemis-management/issues/4

ruromero:
Fix #345


api/v1beta1/activemqartemis_types.go
```diff
@@ -412,4 +412,11 @@ const (
 	DeployedConditionNotReadyReason  = "PodsNotReady"
 	DeployedConditionZeroSizeReason  = "ZeroSizeDeployment"
 	DeployedConditionZeroSizeMessage = "Pods not scheduled. Deployment size is 0"
+
+	BrokerPropsConditionType                             = "BrokerProperties"
```
gtully, line 416: naming is hard!

I think maybe the condition is BrokerPropertiesInSync or BrokerPropertiesApplied, actually think Applied will help in the future when the status contains errors on failure to apply.

Applied = True -> in sync and no errors

gtully: Applied = Unknown for older broker that does not expose Status or when no jolokia agent yet ready

gtully: I think any Unknown should filter up to the  (total) Ready condition being unknown

ruromero: Changes applied, I am having some issues with the IT, don't merge yet

gtully (commented):
that error re queue struct is a nice enhancement

gtully (approved):

Issue #348, opened 2022-10-19
ActiveMQArtemis CRDs v1.0.0 are not backward compatible with v0.20.1


brusdev:
The `redeliveryDelayMultiplier` and `RedeliveryCollisionAvoidanceFactor` fields are *float32 in v2alpha5/v0.20.1 and they are *string in v2alpha5/v1.0.0.

gtully:
in the absence of a webhook conversion, supporting both types is not an option. further more, the inability to parse the v2alpha5 causes all operator watches to fail, leading to a head of the queue blocking problem, and the operator cannot get notification of further cr changes. Given that there is an alternative way to configure address settings in v1beta1 via broker properties and given that we don't yet have webhook conversion, and even that would have trouble with the type variance, the best option is to ignore these fields going forward.

gtully:
with this change, it won't be possible to create new cr's with redeliveryDelayMultiplier or redeliveryCollisionAvoidanceFactor AddressSettingsType fields in yaml:
```
$ kubectl apply  -f ./examples/artemis-basic-address-settings-deployment.yaml 
error: error validating "./examples/artemis-basic-address-settings-deployment.yaml": error validating data: ValidationError(ActiveMQArtemis.spec.addressSettings.addressSetting[1]): unknown field "redeliveryDelayMultiplier" in io.amq.broker.v1beta1.ActiveMQArtemis.spec.addressSettings.addressSetting; if you choose to ignore these errors, turn validation off with --validate=false
```

If necessary, they will have to be in 
```
Spec.BrokerProperties:
 - "addressesSettings.#.redeliveryDelayMultiplier=2.1"
 - "addressesSettings.#.redeliveryCollisionAvoidanceFactor=1.2"
```

PR #349, opened 2022-10-19
[#348] Revert redelivery factor fields to *float32

brusdev:


gaohoward (approved):

gaohoward:
the checkFloat32AsString should be changed to checkFloat32

gtully:
I don't see any down side to reverting these changes to the previous cr versions we retain, and I think it is a necessary part of being able to write a test to verify the need for conversion of some sort.

gtully:
I will try and make a test on this branch

gtully:
There may be some value in using json.Number, not sure yet what that entails. will investigate, would work for a string value that is a float format as well as a plain string.
If we can get into this branch:
https://github.com/kubernetes-sigs/json/blob/main/internal/golang/encoding/json/decode.go#L1088

brusdev:
@gtully your suggestion to use json.Number to fix the conversion issue worked

PR #355, opened 2022-10-24
Implement conversion webhook

ruromero:
Implement conversion webhook between ActiveMQArtemises v2alpha5 and v1beta1


gtully:
This needs a re base from main, the conversion got a little easier after dropping the two problematic float variables. It will allow us to cleanly move to a v1 without interruption in the future so it is a great feature. nice!

gtully:
do we need to create a self signed cert as part of the build? Having something hard coded that can expire will be problematic in the future I guess.

gaohoward:
@ruromero  We may need add conversion for security and address crds, may be do that separately.
To avoid static caBundle values, we may try use cert manager to auto inject the certs.
Also did you test/figure out how it works within OLM framework, we may need some doc here.

gaohoward:
another problem with "static certs" is that it tied to a specific namespace. That's because the cert's CN name must match the webhook service name (so that the service's hostname <service-name>.<namespace>.svc can match the CN so it can be trusted).

gaohoward:
on other thing, If I set ENABLE_WEBHOOKS to *false* and I can't deploy a v1beta1 CR.

ruromero:
@gaohoward I did not test with OLM only on OpenShift with basic deployment, as you noticed, static certs won't be trusted if it is deployed in a different namespace. I just used the existing cert secret generated with the operator-sdk. 

> on other thing, If I set ENABLE_WEBHOOKS to _false_ and I can't deploy a v1beta1 CR.

That is correct, you have to deploy it with the envVar set to true, I can make it the default as this PR requires webhooks to be deployed.

ruromero:
I have updated the kustomize patch during the `make generate-deploy` target patch to set the ENABLE_WEBHOOKS to `true`

gaohoward:
@ruromero Hi, I got some error using the cert manager option. Below is what I did
`$ export USE_CERTMANAGER=true
$ make deploy
$ create -f v2alpha5.yaml 
Error from server (InternalError): error when creating "v2alpha5.yaml": Internal error occurred: conversion webhook for broker.amq.io/v2alpha5, Kind=ActiveMQArtemis returned invalid object: must have the same name: ex-aao != 

$ cat v2alpha5.yaml
apiVersion: broker.amq.io/v2alpha5
kind: ActiveMQArtemis
metadata:
  name: ex-aao
spec:
  deploymentPlan:
    size: 1
    image: placeholder
`

ruromero:
Make sure you're creating the resource in the same namespace where the operator is deployed. I will try again with a fresh install just in case

Makefile
```diff
@@ -8,9 +8,10 @@ VERSION ?= 1.0.0
 KUBE_CLI=kubectl
 OPERATOR_VERSION := 1.0.6
 OPERATOR_ACCOUNT_NAME := activemq-artemis-operator
-OPERATOR_CLUSTER_ROLE_NAME := operator-role
+OPERATOR_CLUSTER_ROLE_NAME := operator
```
ruromero, line 11: @gtully why was this working before?

ruromero:
@gaohoward @gtully I have tested the deployment on minikube and OCP with cert-manager and also on OCP with OLM (see the new target for deploying the catalog source) but another pair of eyes would help in case I missed something.

Thanks

gaohoward:
as discussed with @ruromero I'll be closing this PR and send a new one for review.

brusdev:
This project is now part of the ArkMQ initiative. This repository has been archived as all activities are now happening in the [corresponding ArkMQ repository](https://github.com/arkmq-org/activemq-artemis-operator). See [here](https://artemiscloud.io/community/transition/) for the ArkMQ transition details.

PR #359, opened 2022-10-27
JaaS authentication configmap

### LINKED ISSUE #345: Check properties update status

ruromero:
Fix #345 

If an extraMount is present with the suffix `-jaas-config` it will be used instead of the generated jaas configuration files.
The secret or configMap must contain the login.config file and may or may not contain the additional user/roles properties files. They might be part of a different mount.

This allows automatic reloading of the authentication configuration and users/roles without a restart

```
spec:
  deploymentPlan:
    extraMounts:
      configMaps:
        - ex-aao-jaas-config 
  env:
    - name: JDK_JAVA_OPTIONS
      value: -XshowSettings:system
    - name: AMQ_USER
      value: ruben
    - name: AMQ_PASSWORD
      value: ruben01!
```

In this case I added the env vars for the user/pwd the controller will use by default to connect to the broker.

The configMap can look like this:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: ex-aao-jaas-config
data:
  login.config: |-
    activemq {
       org.apache.activemq.artemis.spi.core.security.jaas.PropertiesLoginModule required
           debug=false
           reload=true
           org.apache.activemq.jaas.properties.user="other-users.properties"
           org.apache.activemq.jaas.properties.role="other-roles.properties";
    };
  other-roles.properties: admin=ruben
  other-users.properties: ruben=ruben01!
```

### DIFF
--- a/api/v1beta1/activemqartemis_types.go
+++ b/api/v1beta1/activemqartemis_types.go
@@ -422,8 +422,6 @@ const (
 	DeployedConditionZeroSizeReason  = "ZeroSizeDeployment"
 	DeployedConditionZeroSizeMessage = "Pods not scheduled. Deployment size is 0"
 
-	ValidConditionType                      = "Valid"
-	ValidConditionSuccessReason             = "ValidationSucceeded"
 	ValidConditionFailedReservedLabelReason = "ReservedLabelReference"
 
 	ConfigAppliedConditionType                            = "BrokerPropertiesApplied"
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -18,11 +18,14 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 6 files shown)


borcsokj:
I've created a PR to handle LDAP login modules and it can be replaced by this feature. Can you add support for environment variable (ie. HAWTIO_REALM=my-console-realm) in https://github.com/artemiscloud/activemq-artemis-broker-init-image repo setting command line argument in artemis.profile so custom/separate JAAS configuration could be used for web console?

*activemq-artemis-broker-init-image* / `script/cfg/broker-etc/artemis.profile`
```
-    JAVA_ARGS="-XX:+PrintClassHistogram -XX:+UseG1GC -XX:+UseStringDeduplication  -Dhawtio.disableProxy=true -Dhawtio.realm=activemq -Dhawtio.offline=true -Dhawtio.rolePrincipalClasses=org.apache.activemq.artemis.spi.core.security.jaas.RolePrincipal -Djolokia.policyLocation=${ARTEMIS_INSTANCE_ETC_URI}jolokia-access.xml -Djava.net.preferIPv4Stack=true -Dbroker.properties=/amq/extra/configmaps/ex-aao-props-00000001/broker.properties"
+    JAVA_ARGS="-XX:+PrintClassHistogram -XX:+UseG1GC -XX:+UseStringDeduplication  -Dhawtio.disableProxy=true -Dhawtio.realm=${HAWTIO_REALM:-activemq} -Dhawtio.offline=true -Dhawtio.rolePrincipalClasses=org.apache.activemq.artemis.spi.core.security.jaas.RolePrincipal -Djolokia.policyLocation=${ARTEMIS_INSTANCE_ETC_URI}jolokia-access.xml -Djava.net.preferIPv4Stack=true -Dbroker.properties=/amq/extra/configmaps/ex-aao-props-00000001/broker.properties"
```

*activemq-artemis-broker-init-image* / `script/cfg/broker-etc/artemis.profile`
```
-               profile_content = re.sub(r'-Dhawtio.realm=activemq ', '', profile_content)
+               profile_content = re.sub(r'-Dhawtio.realm=${HAWTIO_REALM:-activemq} ', '', profile_content)
```

gtully:
this is great, having a special naming convention for a mount point that we know to be jaas config makes great sense. To avoid the need to specify the env var and keep it very simple to configure. Users can still do different if they please.

gtully:
@borcsokj exactly. Any jaas config is now possible.

You can use DEBUG_ARGS in he short term, it is added at the end of the command line.... but I think we need JDK_JAVA_OPTIONS_APPEND (from the jdk ideally) to allow arbitrary overrides.

```
spec:
  deploymentPlan:
    extraMounts:
      configMaps:
        - ex-aao-jaas-config 
env:
    - name: DEBUG_ARGS
      value: -Dhawtio.realm=my-hawtio-realm
```

gtully:
I have raised https://github.com/apache/activemq-artemis/pull/4274 to have an easier option to append arguments, JAVA_ARGS_APPEND

ruromero:
@gtully Unit tests and IT on Microshift are good

Issue #362, opened 2022-10-28
upgrade has issue with existing services that don't have owner reference set


gtully:
Earlier versions of the operator did not always set the owner reference on services, on upgrade this leads to a clash when the current operator attempts to create what it needs on reconcile, it will not update b/c the missing owner reference means it does not get into the deployed list. This happens when console.Expose is true.

Errors of the form:
```
INFO    package k8s_actions    Creating new *v1.Service    {"ActiveMQArtemis Name": "amq-wconsj-0-svc", "Namespace": "amq"}
ERROR    package k8s_actions    Failed to create new *v1.Service    {"ActiveMQArtemis Name": "amq-wconsj-0-svc", "Namespace": "amq", "error": "Service \"amq-wconsj-0-svc\" is invalid: spec.clusterIPs: Invalid value: []string{\"172.30.231.228\"}: failed to allocate IP 172.30.231.228: provided IP is already allocated"}
 manager github.com/artemiscloud/activemq-artemis-operator/controllers(*ActiveMQArtemisReconcilerImpl).createRequestedResource
```

We should look for and claim any existing services that will clash on create.

PR #363, opened 2022-10-28
[#362] - check for console.expose services that don't have an owner r…

gtully:
…eference and reclaim


gaohoward (approved):

brusdev (approved):
I got only one test failure executing the test suite on minikube but it isn't related to this fix, see https://github.com/artemiscloud/activemq-artemis-operator/issues/366

PR #370, opened 2022-11-02
clarify valueFrom

pwright:
maybe this is better?


gtully:
This comment comes from the upstream source v1 , I guess our yaml is auto generated... so my assumption is that any change here wold be overwritten if we regenerate.

pwright:
@gtully fair enough, can you point me in the direction of 'upstream source v1'?

pwright:
closed in favor of https://github.com/kubernetes/kubernetes/pull/114041

PR #372, opened 2022-11-03
Simplify provided examples

michalxo:
Simplified structure and file names. Removed redundant files. 
For information like possible values of key X, we have documentation.


brusdev:
are you removing all ssl examples on purpose?

michalxo:
> are you removing all ssl examples on purpose?

Yes, but not all. There is still `artemis_persistence_tls.yaml`, which is easy to change to "cluster" :)

michalxo:
I don't think we should add any new examples in this PR. It's about not losing value and make simpler.

pwright (commented):
looks good, great to make things simpler. Note: I haven't tested so not commenting on technical change

examples/artemis/artemis_persistence_tls.yaml
```diff
@@ -9,7 +9,7 @@ spec:
...
     messageMigration: true
   acceptors:
-  - name: amqp
-    protocols: amqp
-    port: 5672
-    sslEnabled: true
+    - name: amqp
+      protocols: amqp
+      port: 5672
+      sslEnabled: true
```
brusdev, line 15: do you remove the newline at end of file on purpose?

examples/address/address_topic_auto_remove.yaml
```diff
@@ -6,5 +6,5 @@ spec:
   addressName: emptyAddress
   queueName: ""
```
brusdev, line 7:
```suggestion
```
queueName is not required

examples/address/address_topic_auto_remove.yaml
```diff
@@ -6,5 +6,5 @@ spec:
...
   routingType: multicast
-  removeFromBrokerOnDelete: true
+  removeFromBrokerOnDelete: true  # optional
```
brusdev, line 9:
```suggestion
  removeFromBrokerOnDelete: true
```
The `# optional` comment it is not clear what it means, I would remove it
What about renaming the filename to `address_topic.yaml`?

brusdev (commented):
What about renaming `address_queue_anycast.yam` to `address_queue.yaml`?
What about renaming `address_topic_auto_remove.yaml` to `address_topic.yaml`?

examples/address/address_topic_auto_remove.yaml
michalxo: Removed.

examples/artemis/artemis_persistence_tls.yaml
michalxo: No, that was an IDE issue.

michalxo:
Updated examples and changed cryptic names of deployed resources. All examples except persistence_tls works fine.
`artemis_persistence_tls.yaml` fails to get mount (maybe problem only on my cluster)

brusdev:
> Updated examples and changed cryptic names of deployed resources. All examples except persistence_tls works fine. `artemis_persistence_tls.yaml` fails to get mount (maybe problem only on my cluster)

Yes, persistence_tls needs a secret not created by the example itself. Maybe we should drop this example for now. We could add it back when the operator will support the [cert-manager](https://cert-manager.io/).

gtully (commented):
From my perspective, less is more with examples - and having them verified as part of the test suite is a priority. Great to have some focus on these and this is a great step in the right direction

examples/artemis/artemis_persistence.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: broker.amq.io/v1beta1
 kind: ActiveMQArtemis
 metadata:
-  name: ex-aao
+  name: artemis-broker-tls
 spec:
   deploymentPlan:
     size: 1
```
brusdev, line 7: messageMigration requires deploymentPlan.size > 1

michalxo:
Removed tls example. Renamed `cluster_with_resources` to single broker and naturally gave `persistence` example `cluster_persistence`. I think that `clicked` together nicely now.

examples/artemis/artemis_persistence.yaml
michalxo: I left there messageMigration, increased size to 2 and renamed file to be cluster. :-)

examples/README.md
```diff
@@ -2,16 +2,17 @@
...
 ```bash
-kubectl create -f examples/artemis-basic-deployment.yaml -n activemq-artemis-operator
+kubectl create -f examples/artemis/artemis_single.yaml -n <namespace>
 ```
 
 See https://artemiscloud.io/ for tutorials and more information about Artemis.
 
+For security enabled (TLS) connections example, please follow guide [Secured connection with ArtemisCloud Operator](https://artemiscloud.io/docs/tutorials/ssl_broker_setup).
+
+[ArtemisCloud Operator Security Configuration Example](https://github.com/artemiscloud/artemiscloud-examples/tree/main/operator/keycloak) provides example of JAAS configuration with Keycloak.
```
michalxo, line 18: @howardgao @brusdev @gtully can I link this item here? Is it ok?

brusdev: I would move this link to the doc

PR #385, opened 2022-11-17
[#384] Adding containerPorts to statefulset to support prometheus service monitor

### LINKED ISSUE #384: Adding containerPorts to statefulset to support prometheus service monitor
**Describe the bug**
The prometheus service monitor need to reference the container port to collect metrics.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:


controllers/activemqartemis_reconciler.go
```diff
@@ -846,6 +848,9 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) configureConsoleExposure(custom
 			}
 		}
 	}
+	if len(currentStatefulSet.Spec.Template.Spec.Containers) > 0 {
+		currentStatefulSet.Spec.Template.Spec.Containers[0].Ports = containerPorts
```
gtully, line 852: I think this may step on:
	containerPorts := MakeContainerPorts(customResource)
when the jolokia agent is in play.

We may need to add this logic to make container ports or append

controllers/activemqartemis_controller_test.go
```diff
@@ -799,6 +799,37 @@ var _ = Describe("artemis controller", func() {
...
+				Namespace: defaultNamespace,
+			}
+			currentSS := &appsv1.StatefulSet{}
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, ssKey, currentSS)).Should(Succeed())
+				g.Expect(len(currentSS.Spec.Template.Spec.Containers[0].Ports)).To(Equal(2))
+				g.Expect(currentSS.Spec.Template.Spec.Containers[0].Ports[0].Name).To(Equal("wconsj-0"))
+				g.Expect(currentSS.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort).To(Equal(int32(8161)))
+				g.Expect(currentSS.Spec.Template.Spec.Containers[0].Ports[0].Protocol).To(Equal(corev1.ProtocolTCP))
+				g.Expect(currentSS.Spec.Template.Spec.Containers[0].Ports[1].Name).To(Equal("wconsj-1"))
```
brusdev, line 825: I'm a bit confused, why do we need more than one port 8161 in the broker container spec?

gaohoward: It depends on deploymentPlan: size. If size is 2, there will be two pods available for collecting metrics from. So we need to expose all the ports in the spec so that prometheus can collect each of them.

controllers/activemqartemis_reconciler.go
gaohoward: yes good point. Will update to append and include it in test.

controllers/activemqartemis_controller_test.go
brusdev: For Hawtio Online we only added one port (`jolokia`) for all pods of a statefulset to support discovery, If multiple ports are defined in the broker container spec then each pods of the statefulset have multiple ports. I don't understand why the prometheus service monitor should require to define multiple ports. Is that a requirement from the prometheus doc?

gaohoward: The service monitor collects metrics from a list of container ports. Not sure if *jolokia* port does the same thing but that's requested by some user. See https://issues.redhat.com/browse/ENTMQBR-5228

brusdev: I reviewed [ENTMQBR-5228](https://issues.redhat.com/browse/ENTMQBR-5228) and I'm not able to find where it is requested to add multiple ports for the console in each broker container.

See the example in the prometheus-operator getting-started guide:
https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md

One port defined in the statefulset for all 3 replicas
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-app
spec:
  replicas: 3
  selector:
    matchLabels:
      app: example-app
  template:
    metadata:
      labels:
        app: example-app
    spec:
      containers:
      - name: example-app
        image: fabxc/instrumented_app
        ports:
        - name: web
          containerPort: 8080
```
```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: example-app
  labels:
    team: frontend
spec:
  selector:
    matchLabels:
      app: example-app
  podMetricsEndpoints:
  - port: web
```

gtully: I came across the same doc and you bet me to the comment. I bet the service monitor queries the stateful set metadata to find the existing set of service endpoints that host that port. Ie, it knows about a stateful set.

we man need a round trip test to validate fully, but it seems a single port is sufficient, the service exposes it as necessary based on the replicas.

controllers/activemqartemis_reconciler.go
```diff
@@ -1546,13 +1554,20 @@ func MakeContainerPorts(cr *brokerv1beta1.ActiveMQArtemis) []corev1.ContainerPor
...
 	if cr.Spec.DeploymentPlan.JolokiaAgentEnabled {
 		jolokiaContainerPort := corev1.ContainerPort{
-
 			Name:          "jolokia",
 			ContainerPort: 8778,
 			Protocol:      "TCP",
 		}
 		containerPorts = append(containerPorts, jolokiaContainerPort)
 	}
+	if cr.Spec.Console.Expose {
```
brusdev, line 1563: It could be good to define the console container port even if we don't need to expose the console (only an app running in the Kubernetes cluster should be able to access a pod port), WDYT?

brusdev: Yes, a PodMonitor CR should only requires should a container port, see [Using PodMonitors](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#using-podmonitors):
```
Instead of a ServiceMonitor, we can use a PodMonitor which doesn't require the creation of a Kubernetes Service. In practice, the spec.selector label tells Prometheus which Pods should be scraped.
```

gaohoward: ok, I'll just remove the condition check. Thanks Dom.

controllers/activemqartemis_reconciler.go
```diff
@@ -818,6 +818,9 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) configureConsoleExposure(custom
 		Name:      customResource.Name,
 		Namespace: customResource.Namespace,
 	}
+	commonPortName := "wconsj"
+	secondPort := int32(8261)
+	portNumber := int32(8161)
```
brusdev, line 823: if the port name with the ordinal will be deprecated should the port 8161 have the name `wconsj`?
```suggestion
	commonPortName := "wconsj"
	secondPort := int32(8161)
	portNumber := int32(8162)
```

brusdev (approved):

brusdev:
@gtully @RoddieKieley this PR LGTM, I'm going to merge it before EOD if you have no concerns

brusdev:
@howardgao I squashed the commits before merging

PR #394, opened 2022-11-28
[#393] Support multiple security CRs applied to one broker deployment

### LINKED ISSUE #393: Support multiple security custom resources applied to one broker deployment
When multiple security CRs applied to the same broker, they are combined instead of replace each other.

**Additional context**

gaohoward:


controllers/activemqartemis_controller.go
```diff
@@ -46,57 +46,118 @@ import (
...
 }
 
+func (r *ActiveMQArtemisReconciler) IsSecurityApplied(artemis *brokerv1beta1.ActiveMQArtemis, handler common.ActiveMQArtemisConfigHandler) bool {
+	//looking for the secret <cr-name>-applied-security-secret (owned by broker CR)
+	secretKey := types.NamespacedName{
+		Name:      artemis.Name + "-applied-security",
+		Namespace: artemis.Namespace,
+	}
+	appliedSecSecret := corev1.Secret{}
+	err := r.Client.Get(context.TODO(), secretKey, &appliedSecSecret)
```
gtully, line 125: This looks like we are again adding state to the reconcile loop, I don't think this is wise. The reconcile process should not be conditional on the exist of a secret, this is essentially two sources of truth. What will make them converge and or stop them diverging. I think we have to make it such that the process of applying is fast and idempotent.

gaohoward: Yeah that one need some more thought. I'm still trying to figure that out.

brusdev:
@howardgao are you still working on it?

gtully:
I think we should close this PR and work on a new onboarding or app-provisioning CR, that can capture the desired address and permissions (roles) requirements of an app, and work out how that can be mapped to a broker with some labels and reconciled via the content of properties files on the broker.

This sort if thing:
```
{
 - name: app-a1 

 - producer on [WorkQueue] with role [RoleA]
 - producer on [BroadCastChannel] with role [RoleA]

 - <competing> <durable> <max-back-log> consumer on [WorkQueue] with role [RoleA]
 - <competing> <durable> consumer on [BroadCastChannel] with role [RoleA, PartitionOne]
}
```
The challenge is capturing the exact verbs to expose, but we can start with this (real) use case.

Issue #392, opened 2022-11-24
Use shortNames for custom resource


michalxo:
It might be cool and easier to use shortNames in all ActiveMQArtemis Custom Resource/s
Example
```yaml
names:
    plural: crontabs
    singular: crontab
    kind: CronTab
    shortNames:
    - ct
```

And use it like 
`kubectl get ct -n namespace-name`

```yaml
  group: broker.amq.io
  names:
    kind: ActiveMQArtemis
    listKind: ActiveMQArtemisList
    plural: activemqartemises
    singular: activemqartemis
    shortNames:
    - aa
 ///
    - aaa (address)
    - aas (security)
    - aasd (for scaledown resource)
```

For more information see [shortNames link](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) .

michalxo:
Maybe having following might be better?
`ac` - ArtemisCloud
`aca` - ACAddress
`acs` - ACSecurity
`acd` - ACScaleDown

brusdev:
I repeated the test using CRDs with different groups. In this case, Kubernetes accepts CRDs with conflicting shortnames. This means CRD full names can be used also if shortnames conflict but conflicting shortnames are only enabled for the first CRD created.

brusdev:
Included in 1.0.12

PR #395, opened 2022-11-28
Introduce shortnames into CR/Ds

michalxo:
I have tried to implement short names as described in #392 
Currently I did `aa` and `ac`. Easy to pick one or completely different after some broader discussion.

### DIFF
--- a/api/v1alpha1/activemqartemissecurity_types.go
+++ b/api/v1alpha1/activemqartemissecurity_types.go
@@ -197,6 +197,7 @@ type ActiveMQArtemisSecurityStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:resource:path=activemqartemissecurities,shortName=aas
 
 // ActiveMQArtemisSecurity is the Schema for the activemqartemissecurities API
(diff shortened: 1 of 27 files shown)


brusdev (commented):
@michalxo nice improvement, I would only use `ac` shortnames

brusdev:
I'll close this PR because the current version of Kubernetes doesn't provide good support for shorthame conflicts causing the impossibility of using conflicted CRDs.

brusdev:
I would propose to keep only shortnames starting with `ac`

michalxo:
As in voting `ac` prefix won, I have updated this PR and should be ready to merge.

brusdev:
could you update the bundle/manifests too?

gtully:
I like the idea of short names, and am in favour of aa for activemqartemis, but i think the other CRD's don't need this love. I have added deprecation to artemissecurty in the past week due to it needing restarts and think the limitations of addresses (single queue, using jmx etc) will force it to be deprecated also in favour of brokerProperties or config map references.
in short, just aa for ActivemqArtemisCRs makes sense to me.

brusdev (approved):

Issue #396, opened 2022-11-28
ActiveMQArtemisAddress does not persist after a restart on AMQ operator pods and the broker pods


brusdev:
In v1.0.7, when "persistenceEnabled: false" and create a broker pod and a address, it works (verified at the webconsole). However, when the operator pod and broker pod restarted, the address is gone (oc get activemqartemisaddress shows the address existed, but the webconsole does not).

brusdev:
Resolved in 1.0.8

PR #397, opened 2022-11-28
[#396] Fix address reconciliation after restart

brusdev:


gtully:
I was wondering if we would find a sufficient reason to refactor this state from the address controller.... but this fix looks trivial enough to just do. Keep an eye out for an opportunity to remove the need for that secret, the lscr

gtully (approved):

gaohoward (approved):

PR #411, opened 2022-12-07
[#410] ConfigurationManaged property in address CR should default to true

### LINKED ISSUE #410: ConfigurationManaged property in address CR should default to true
**Describe the bug**
When deploying a custom resource for an address the configurationManaged property is set to false. This is incorrect since as a result the deployed address is treaded differently by the operator from any address defined in broker.xml. There should be no difference in the way addresses are handled by the operator.
The property configurationManaged should be set to true by default.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:


controllers/activemqartemisaddress_controller.go
```diff
@@ -235,49 +235,43 @@ func createAddressResource(a *jc.JkInfo, addressRes *brokerv1beta1.ActiveMQArtem
...
-			if nil != err {
-				if mgmt.GetCreationError(respData) == mgmt.QUEUE_ALREADY_EXISTS {
-					glog.Info("The queue already exists, updating", "queue", queueCfg)
-					respData, err := a.Artemis.UpdateQueue(queueCfg)
-					if err != nil {
-						glog.Error(err, "Failed to update queue", "details", respData)
-					}
-					return err
+		} else if addressRes.Spec.QueueConfiguration.ConfigurationManaged == nil {
+			addressRes.Spec.QueueConfiguration.ConfigurationManaged = &defaultConfigurationManaged
```
gtully, line 250: I am missing where we apply a non nil value from addressRes.Spec.QueueConfiguration.ConfigurationManaged

controllers/activemqartemisaddress_controller_test.go
```diff
@@ -1101,3 +1176,64 @@ func CheckQueueExistInPod(brokerCrName string, podName string, queueName string,
...
+	Eventually(func(g Gomega) {
+		execReq := restClient.
+			Post().
+			Namespace(namespace).
+			Resource("pods").
+			Name(podName).
+			SubResource("exec").
+			VersionedParams(&corev1.PodExecOptions{
+				Container: brokerCrName + "-container",
+				Command:   []string{"curl", "-H", "Origin:", "http://" + podName + ":8161", "-u", "anyuser:anypasswd", "http://" + podName + ":8161/console/jolokia/read/org.apache.activemq.artemis:broker=\"amq-broker\",address=\"" + addressName + "\",component=addresses,queue=\"" + queueName + "\",routing-type=\"" + routingType + "\",subcomponent=queues/" + attrName}, //curl -H "Origin: http://localhost:8161" -u guest:guest http://localhost:8161/console/jolokia/list
```
gtully, line 1212: maybe pull the comment over this Command or remove - it is way too long a line!

brusdev: An alternative  to avoid to run the command from the pod is using the jolokia client, i.e.
https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemisaddress_controller_test.go#L533
```
						jolokia := jolokia.GetJolokia(pod.Status.PodIP, "8161", "/console/jolokia", "", "", "http")
						data, err := jolokia.Read("org.apache.activemq.artemis:broker=\"amq-broker\",address=\"" + addressName + "\",component=addresses,queue=\"" + queueName + "\",routing-type=\"" + routingType + "\",subcomponent=queues/" + attrName")

```

gaohoward: yep didn't noticed the comment is added to the same line. 
I'll try to use the jolokia client as it's merged and convenient.

controllers/activemqartemisaddress_controller.go
gaohoward: Yes if the configurationManaged is not nil, we don't need handle it, it will just be passed to jolokia to create the queue.

gtully (approved):

PR #412, opened 2022-12-07
update go version to reflect current state

rhmkrutov:
Problem:
go.mod references go 1.16
go 1.16.15 fails to build this with non-descriptive errors (primarily, controller-gen fails)

to avoid this in future, let's keep up with proper go.mod versioning as this is the primary source of truth for versioning purposes.


gtully:
looks like this commit needs the result of go  mod tidy to complete it, see the failed check

mkrutov:
Oh yeah, I missed the notification that this failed the checks. Updated as per suggested command (compatibility with 1.16 is broken *anyways* so no need to keep it, two options suggested by `go mod tidy` were either -compat=1.16 or =1.17.)

brusdev:
@mkrutov thanks for your contribution, could you rebase your branch using the main branch?

andytaylor:
@mkrutov Could you also update the build docs please?

brusdev (approved):

Issue #413, opened 2022-12-08
Track failure to apply broker properties in CR condition


gtully:
the CR.brokerProperties allows the broker to be configured, with 2.17.0 there is some feedback in the Status attribute of the server. This needs to be surfaced in a CR condition such that it is possible to see when config is applied and if it applied without any error.

PR #414, opened 2022-12-08
#413 - reflect broker config props apply errors in cr status condition

gtully:


gtully:
a question? the new condition is applied with error. and it is true once the properties are in sync. I wonder if we want to surface this condition as ready false? I have not done that.

controllers/activemqartemis_reconciler.go
```diff
@@ -2622,6 +2630,13 @@ func UpdateBrokerPropertiesStatus(cr *brokerv1beta1.ActiveMQArtemis, client rtcl
 				Reason:  brokerv1beta1.ConfigAppliedConditionOutOfSyncReason,
 				Message: brokerv1beta1.ConfigAppliedConditionOutOfSyncMessage,
 			}
+		case inSyncApplyError:
+			condition = metav1.Condition{
+				Type:    brokerv1beta1.ConfigAppliedConditionType,
+				Status:  metav1.ConditionTrue,
```
brusdev, line 2636: Should the `status` be false?

gtully: that is the question. I think not, b/c the condition is about being in sync. but I think it could go either way... the conditions are  accumulative, so ready would be false b/c this condition is false. I can go either way on this... having a bla=bla property that has been ignored, from a CR status perspective, think it is minor.

gtully: I think you are right. The genesis of the condition is the delay in a config mount projection, has the broker seen the projection update (with a mutable) config map. that can take up to 2mins. But I guess users don't care about that really, it is fine to be unknown, then true or false based on the actual content being valid rather than just being applied.
you win! thanks for the feedback.

gaohoward (approved):

Issue #420, opened 2022-12-15
To allow configure log level for the operator via configmap


gaohoward:
**Is your feature request related to a problem? Please describe.**
When deploy the operator manually user can directly edit the yaml to set the log level. However
when operator is deployed with olm, that wouldn't work. A configmap should be provided as
a way to solve this. The operator at its startup will look for this configmap to init the log level.

**Describe the solution you'd like**

**Describe alternatives you've considered**
Optional. A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
Optional. Add any other context or screenshots about the feature request here.

gtully:
I think the config map should be of the general cmd_line_args_append format. and support any argument addition. A map of the yaml "args"

gtully:
the ARGS env is great.

PR #421, opened 2022-12-15
[#420] To allow configure log level for the operator

gaohoward:


main.go
```diff
@@ -95,6 +98,49 @@ func init() {
...
+	}
+	oprNamespace, err := sdkk8sutil.GetOperatorNamespace()
+	if err != nil {
+		log.Error(err, "failed to get operator namespace")
+		os.Exit(1)
+	}
+
+	oprCfg, _ := configmaps.RetrieveConfigMap(oprNamespace, "artemiscloud-operator-config", clnt)
+	if oprCfg != nil && len(oprCfg.Data) > 0 {
+		if level, ok := oprCfg.Data["zap-log-level"]; ok {
```
brusdev, line 121: If we use a generic key like `log-level` it will work also if the `--zap-log-level` option will change

main.go
```diff
@@ -95,6 +98,49 @@ func init() {
...
+		log.Error(err, "can't create client from config")
+		os.Exit(1)
+	}
+	oprNamespace, err := sdkk8sutil.GetOperatorNamespace()
+	if err != nil {
+		log.Error(err, "failed to get operator namespace")
+		os.Exit(1)
+	}
+
+	oprCfg, _ := configmaps.RetrieveConfigMap(oprNamespace, "artemiscloud-operator-config", clnt)
```
brusdev, line 119: the operator deployment name is `activemq-artemis-controller-manager` maybe the configmap name could be `activemq-artemis-controller-config` or `activemq-artemis-operator-config`

controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -0,0 +1,124 @@
...
+				Eventually(func(g Gomega) {
+					oprLog, err := GetOperatorLog(defaultNamespace)
+					g.Expect(err).To(BeNil())
+					g.Expect(*oprLog).To(ContainSubstring("INFO"))
+					g.Expect(*oprLog).To(ContainSubstring("DEBUG"))
+					g.Expect(*oprLog).NotTo(ContainSubstring("ERROR"))
+				}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
+				By("Uninstall existing operator")
+				uninstallOperator(false)
```
brusdev, line 84: I suppose the users will only restart the operator, maybe deleting the relative pod. I was wondering if the test should work at the same way, WDYT?

Makefile
```diff
@@ -129,6 +129,11 @@ test-mk-do: manifests generate fmt vet envtest generate-deploy ## Run tests agai
 test-mk-do-fast: manifests generate fmt vet envtest generate-deploy ## Run tests against minikube with deployed operator(do) and exclude slow
 	DEPLOY_OPERATOR=true USE_EXISTING_CLUSTER=true ENABLE_WEBHOOKS=false go test  ./controllers/... -test.timeout=40m -ginkgo.slow-spec-threshold=30s -ginkgo.fail-fast -ginkgo.v -ginkgo.label-filter="do && !slow"
 
+## It requires a registry to push the image for testing
+## example: make OPERATOR_IMAGE_REPO=quay.io/hgao/operator OPERATOR_VERSION=1.0 test-mk-doh
+test-mk-doh: manifests generate fmt vet envtest generate-deploy docker-build docker-push ## Run tests against minikube with deployed operator image built from repo head(doh)
+	DEPLOY_OPERATOR=true USE_EXISTING_CLUSTER=true ENABLE_WEBHOOKS=false go test  ./controllers/... -test.timeout=40m -ginkgo.slow-spec-threshold=30s -ginkgo.fail-fast -ginkgo.v -ginkgo.label-filter="doh"
+
```
brusdev, line 136: if we change the test restarting the operator we don't need to deploy it, i.e.
```
// Restart the controller manager
shutdownControllerManager()
createControllerManagerForSuite()
```

main.go
gaohoward: yeah log-level seems better.

main.go
gaohoward: +1 for activemq-artemis-operator-config

controllers/activemqartemis_controller_deploy_operator_test.go
gaohoward: Here the uninstallOperator is just to clean up the previous deploy and then testing a fresh install and see if the config can be picked up.

Makefile
gaohoward: I need to test a real operator deployed into k8s/openshift. The shutdownControllerManager/createControllerManager cannot serve the purpose, it runs a local controller manager which is not exactly build from operator's main.go.

brusdev: I see, why do wee need to add `test-mk-doh` could we reuse `test-mk-do`?

controllers/suite_test.go
```diff
@@ -468,7 +471,7 @@ func installYamlResource(resPath string) error {
 	cobj := robj.(client.Object)
 	cobj.SetNamespace(defaultNamespace)
 
-	if oprImg := os.Getenv("OPERATOR_IMAGE"); oprImg != "" {
+	if oprImg := os.Getenv("IMG"); oprImg != "" {
```
brusdev, line 474: Should `IMG` be `OPERATOR_IMAGE_REPO`?
```suggestion
	if oprImg := os.Getenv("OPERATOR_IMAGE_REPO"); oprImg != "" {
```

Makefile
gaohoward: test-mk-doh will build the operator from your local repo, push it to a repo and deploy the operator. test-mk-do just use the operator latest tag on artemiscloud repo. In my case I nee to build the operator to include my fix and test it.
so I added 'test-mk-doh' for that matter, and also it will take longer time to build and test.

controllers/suite_test.go
gaohoward: We should use IMG as in the Makefile we define the IMG as
IMG ?= $(OPERATOR_IMAGE_REPO):$(OPERATOR_VERSION)
So we running the test you need to define OPERATOR_IMAGE_REPO and OPERATOR_VERSION

Makefile
gtully: that test-mk-do probably should do that. 
I am using it via:
```
make OPERATOR_IMAGE_REPO=localhost/am OPERATOR_VERSION=v1.0 build
podman-remote build -t am:v1.0 .
make OPERATOR_IMAGE_REPO=localhost/am OPERATOR_VERSION=v1.0 test-mk-do
```

in this way, the deployed yamls would reference the am:v1.0 image and not the latest.
the podman-remote build gets the image loaded into my local podman registry on the minikube vm

gaohoward: does it work on openshift?

gtully: with a different registry or with openshift access to the local podman registry it should.
I guess the build and push steps will be necessary to automate when it is not in code.

main.go
```diff
@@ -95,6 +98,49 @@ func init() {
...
+	}
+	oprNamespace, err := sdkk8sutil.GetOperatorNamespace()
+	if err != nil {
+		log.Error(err, "failed to get operator namespace")
+		os.Exit(1)
+	}
+
+	oprCfg, _ := configmaps.RetrieveConfigMap(oprNamespace, "activemq-artemis-operator-config", clnt)
+	if oprCfg != nil && len(oprCfg.Data) > 0 {
+		if level, ok := oprCfg.Data["log-level"]; ok {
```
gtully, line 121: I think maybe this should be generic "args", not specific to logging config - just any append-args

main.go
```diff
@@ -108,9 +154,8 @@ func main() {
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
```
gtully, line 156: then here we would bind flags with the args from the config map.

main.go
gaohoward: agreed. will change accordingly.

Makefile
gaohoward: ok, so the target seems redundant. I'll remove it.

main.go
gaohoward: Don't need touch the BindFlags which is just add those logger options to command line.

gaohoward:
> how does this new config map compare with https://github.com/artemiscloud/activemq-artemis-operator/blob/main/deploy/install/100_operator_config.yaml - is there some overlap?
> 
> am also wondering about a single source of truth!

I don't know about this configmap. Will look into it.

gaohoward:
> > how does this new config map compare with https://github.com/artemiscloud/activemq-artemis-operator/blob/main/deploy/install/100_operator_config.yaml - is there some overlap?
> > am also wondering about a single source of truth!
> 
> I don't know about this configmap. Will look into it.

Here is the definition of this configmap https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/config/v1alpha1#ControllerManagerConfigurationSpec

The configmap configures the operator however it seems doesn't have log options.

controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -0,0 +1,153 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+var _ = Describe("artemis controller", Label("doh"), func() {
```
brusdev, line 38: I see `doh` label works but  `test-mk-do` has the filter `do`
```suggestion
var _ = Describe("artemis controller", Label("do"), func() {
```

gaohoward: ah, forgot that. Thanks!

main.go
brusdev: generic "args" is a nice idea my only concern is related to the data structure of the configmap,  the current one has 1 level so all data keys should be an opertator arg but if we add another level using the `args` data key in the future we could resuse this config map for other settings

current configmap data structure with one level:
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: activemq-artemis-operator-config
immutable: false
data:
  --zap-log-level: debug
  --zap-encoder: console
  --zap-time-encoding: iso8601
  --leader-elect:
```

alternative configmap data structure with 2 levels:
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: activemq-artemis-operator-config
immutable: false
data:
  args: |
    --zap-log-level=debug
    --zap-encoder=console
    --zap-time-encoding=iso8601
    --leader-elect
```

brusdev:
@gaohoward could you add the documentation/example for this new feature?

gaohoward:
> @gaohoward could you add the documentation/example for this new feature?

sure will do.

main.go
gaohoward: I think for now one-level is ok, and it's simple.

gaohoward: @brusdev Gary mentioned that we may consider using this configmap for that purpose instead of creating a new configmap. https://github.com/artemiscloud/activemq-artemis-operator/blob/main/deploy/install/100_operator_config.yaml , this is generated from kustomize and I think it may be possible to add a 'args' entry into it.

gaohoward: but the problem is that how user can pre-create the configmap when the operatorHub will install it again.

brusdev:
An alternative approach to support configuration for operators installed using OLM is using [Subscription Config](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md) to define a list of Environment Variables that must exist in all containers in the Pod created by OLM, i.e.
```
kind: Subscription
metadata:
  name: prometheus
spec:
  package: prometheus
  channel: alpha
  config:
    env:
    - name: ARGS
      value: "--zap-log-level=debug"
```

The OpenShift UI already support the [Subscription Config](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md), for further details see http://openshift.github.io/openshift-origin-design/designs/administrator/olm/env-toleration-config-on-olm-subscription/

gaohoward:
@brusdev @gtully Updated PR as discussed.

docs/help/operator.md
```diff
@@ -84,6 +84,34 @@ in your Custom Resource), you need to have two PVs available. By default, each b
 If you specify persistenceEnabled=false in your Custom Resource, the deployed brokers uses ephemeral storage. Ephemeral 
 storage means that that every time you restart the broker Pods, any existing data is lost.
 
+## Configuring logging for the Operator
+
+This section describes how to configure logging for the operator.
+
+The operator image is using zap logger for logging. You can define logging options in the operator yaml before
+deploying the operator as shown in https://github.com/artemiscloud/activemq-artemis-operator/blob/main/deploy/install/110_operator.yaml#L24. This works for manual installing of the operator.
```
brusdev, line 92: The operator image is using zap logger for logging. You can set the zap log level editing the container args defined in the operator deployment, i.e.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
  name: activemq-artemis-controller-manager
  spec:
      containers:
      - args:
        - --zap-log-level=error
...
```

You can use edit the operator deployment using the Kubernetes dashboard or the Kubernetes command-line tool, i.e.
```
`sed 's/--zap-log-level=.*/--zap-log-level=error/' deploy/install/110_operator.yaml | kubectl apply -f -`
```

main.go
```diff
@@ -95,6 +96,37 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func SetupLogging(opts *zap.Options) {
+	if logLevelArg, ok := common.FindArgFromEnv("ARGS", "--zap-log-level="); ok {
```
brusdev, line 100: I don't think we need to validate arguments, we could just append/replace all them. If I pass a bad zap log level the operator process already fails, i.e.
```
spec:
      - args:
        - --zap-log-level=brusdev
```
```
Executing entrypoint
exec /home/activemq-artemis-operator/bin/activemq-artemis-operator
invalid value "brusdev" for flag -zap-log-level: invalid log level "brusdev"
Usage of /home/activemq-artemis-operator/bin/activemq-artemis-operator:
  -health-probe-bind-address string
...
```
It should be up to `flag.Parse()` validate args.

docs/help/operator.md
gaohoward: +1

main.go
gaohoward: +1

main.go
```diff
@@ -108,9 +133,8 @@ func main() {
 		Development: true,
 	}
 	opts.BindFlags(flag.CommandLine)
-	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	SetupLogging(&opts)
```
brusdev, line 137:
```suggestion
	os.Args = append(os.Args, strings.Split(os.Getenv("ARGS"), " ")...)
	
	flag.Parse()

	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
```

If an option is appended multiple times only the last option value is applied, so we can simplify the `ARGS` parsing/appending.

gaohoward: +1, will update my branch. thx.

brusdev (approved):

PR #423, opened 2022-12-15
Operator auth with serviceAccount and split mgmt and app realms

### LINKED ISSUE #417: Operator should authenticate using ServiceAccount token
The operator should leverage the [KubernetesLoginModule](https://github.com/apache/activemq-artemis/blob/main/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/KubernetesLoginModule.java) to authenticate to the broker instead of using the user/password

A new login.config should be used to extend the current support for propertiesLoginModule to configure the KubernetesLoginModule with admin role for the Operator service account.
It would be better if management and application users could be configured separately.

**Additional context**

https://github.com/apache/activemq-artemis/pull/4289

ruromero:
Fix https://github.com/artemiscloud/activemq-artemis-operator/issues/417

Depends on
* https://github.com/artemiscloud/activemq-artemis-management/pull/10
* https://github.com/artemiscloud/activemq-artemis-operator/pull/403

Based on the broker version, the operator will authenticate using the serviceAccount or the username/password. The operator will also configure different realms for management and application users.


ruromero:
The PR is missing the decision of whether to authenticate using the service account or the username/password

controllers/activemqartemis_controller_test.go
```diff
@@ -220,7 +224,8 @@ var _ = Describe("artemis controller", func() {
 				g.Expect(ownerFound).To(BeTrue())
 
 				g.Expect(k8sClient.Get(ctx, ssKey, createdSs)).Should(Succeed())
-
+				g.Expect(k8sClient.Get(ctx, saKey, &corev1.ServiceAccount{})).Should(Succeed())
+				g.Expect(k8sClient.Get(ctx, crbKey, &rbacv1.ClusterRoleBinding{}))
```
gaohoward, line 228: Should(Succeed()) is missing here.

controllers/activemqartemis_controller_test.go
```diff
@@ -3871,252 +3876,341 @@ var _ = Describe("artemis controller", func() {
...
 
+			if test.shouldCreate {
+				Expect(k8sClient.Delete(ctx, resource)).Should(Succeed())
+				Eventually(func(g Gomega) {
+					key := types.NamespacedName{
+						Name:      resource.GetName(),
+						Namespace: resource.GetNamespace(),
+					}
+					g.Expect(k8sClient.Get(ctx, key, test.buildResource())).To(BeFalse())
+				})
```
gaohoward, line 3935: Here Eventually(func(g Gomega){...}), should add timeout, interval).Should(Succeed())

controllers/activemqartemis_controller_test.go
```diff
@@ -3871,252 +3876,341 @@ var _ = Describe("artemis controller", func() {
...
-				return err == nil
-			}, timeout, interval).Should(BeTrue())
-
-			By("By creating a new crd")
-			crd := generateArtemisSpec(defaultNamespace)
-			crd.Spec.DeploymentPlan.ExtraMounts.ConfigMaps = []string{loggingConfigMapName}
-			Expect(k8sClient.Create(ctx, &crd)).Should(Succeed())
+			Eventually(func(g Gomega) {
+				g.Expect(getPersistedVersionedCrd(crd.Name, defaultNamespace, createdCrd)).To(BeFalse())
+			})
```
gaohoward, line 3941: missing timeout,interval).Should(Succeed())

ruromero:
Thanks @gaohoward for the missing bits.

gtully:
this is great :-)

I am wondering about the two realms, it would be ideal to be able to use the service account for broker to broker comms, like a federation connection, but that will use the "activemq" realm. Afaik, the security manager uses a single realm "activemq" by default. It is configurable, but there is only one.

with the security.properties approach - is there a merge option? That will bring its own potential problems too, but I think it may be necessary. ie: into the activemq realm we add the service account login module for use inter-broker and by management

the other thought, separate, is that we do this just for the jolokia agent and always configure that for the operator. In that way we are locking down the broker and we deprecate the deployment of the web console war.

config/rbac/role.yaml
```diff
@@ -1,4 +1,35 @@
 
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
brusdev, line 4: Is the cluster role required also when the operator is installed in a single namespace?

ruromero: Yes. The tokenreview granted to the broker SA is a cluster resource and requires a cluster role.

ruromero:
> this is great :-)
> 
> I am wondering about the two realms, it would be ideal to be able to use the service account for broker to broker comms, like a federation connection, but that will use the "activemq" realm. Afaik, the security manager uses a single realm "activemq" by default. It is configurable, but there is only one.
> 

Isn't federation configured with a specific credentials? https://access.redhat.com/documentation/en-us/red_hat_amq/2020.q4/html/configuring_amq_broker/assembly-br-configuring-addresses-and-queues_configuring#proc-br-configuring-upstream-address-federation_configuring
I don't see how to use here the jaas configuration

> with the security.properties approach - is there a merge option? That will bring its own potential problems too, but I think it may be necessary. ie: into the activemq realm we add the service account login module for use inter-broker and by management

Yes, this security.properties file adds an additional login.config [1] file that will be appended to the first one available on the default location [0]. This allows users to customize the activemq realm and also provide additional management configurations.
> 
> the other thought, separate, is that we do this just for the jolokia agent and always configure that for the operator. In that way we are locking down the broker and we deprecate the deployment of the web console war.

I think, the jolokia agent is the only one using the management realm.

gtully:
looking in more detail at this, I think we need a feature flag, otherwise all existing deployments will be restarted with the extra volume mount needing a stateful set update/refresh.

and thinking more, maybe the flag is the presence of a -jaas-config extra mount. When external jaas config is specified the operator manages its own for the control plane, ie: the "management" realm

gtully:
and the management realm config should be in a secret rather than a config map.

controllers/activemqartemis_reconciler.go
```diff
@@ -2173,6 +2206,96 @@ func brokerPropertiesData(props []string) map[string]string {
...
+		fmt.Fprintln(login_config, "    org.apache.activemq.jaas.kubernetes.role=\"k8s-roles.properties\"")
+		fmt.Fprintf(login_config, "    baseDir=\"%v\"\n", getJaasManagementPath(customResource))
+		fmt.Fprintln(login_config, "  ;")
+		fmt.Fprintln(login_config, "")
+
+		k8s_roles_properties := newPropsWithHeader()
+		bytes, err := sdkk8sutil.GetOperatorNamespace()
+		namespace := "local"
+		if err != nil {
+			clog.V(5).Info("Using --localOnly without --namespace, but unable to determine namespace")
```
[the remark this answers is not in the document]
ruromero: This is just a placeholder for parsing the `k8s-roles.properties` file when not running on k8s in case the `sdkk8sutil.GetOperatorNamespace()` fails to retrieve the ns. That would happen only if it is running locally and not connected to a k8s cluster but that case is very unlikely to happen so I can just replace it by `""` to avoid misunderstandings.

brusdev:
Incorporated by #1038

Issue #438, opened 2023-01-09
use a secret for mutable brokerProperties


gtully:
configuration may be sensitive, by using a secret it is possible to make it confidential
with the change to mutable in #323 we can also move to using a secret.

PR #439, opened 2023-01-09
#438 use a secret for mutable broker properties

gtully:


brusdev:
> I was wondering a little if we loose anything with visibility here, but the data the populates the secret is in the CR, so we can always view it there. The secret can protect it when it is deployed

This PR make sense to me because the Kubernetes RBAC is very flexible, so you could have monitor users with no access to the CRDs and secrets data

controllers/activemqartemis_controller_test.go
```diff
@@ -3888,7 +3887,7 @@ var _ = Describe("artemis controller", func() {
 			Expect(createdCrd.Name).Should(Equal(crd.ObjectMeta.Name))
 
 			By("By finding a new config map with broker props")
-			configMap := &corev1.ConfigMap{}
+			configMap := &corev1.Secret{}
```
brusdev, line 3890: could you rename the configMap variables to align them with theirs new type?

controllers/activemqartemis_reconciler.go
```diff
@@ -1668,11 +1671,16 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
...
 
 	reqLogger.Info("Checking out extraMounts", "extra config", customResource.Spec.DeploymentPlan.ExtraMounts)
-	brokerPropertiesConfigMapName := reconciler.addConfigMapForBrokerProperties(customResource)
-	configMapsToCreate := []string{brokerPropertiesConfigMapName}
-	configMapsToCreate = append(configMapsToCreate, customResource.Spec.DeploymentPlan.ExtraMounts.ConfigMaps...)
 
-	extraVolumes, extraVolumeMounts := createExtraConfigmapsAndSecrets(container, configMapsToCreate, customResource.Spec.DeploymentPlan.ExtraMounts.Secrets)
+	configMapsToCreate := customResource.Spec.DeploymentPlan.ExtraMounts.ConfigMaps
+	secretsToCreate := customResource.Spec.DeploymentPlan.ExtraMounts.Secrets
+	resourceName, isSecret := reconciler.addResourceForBrokerProperties(customResource, namer)
```
brusdev, line 1677: why do we need `isSecret`?

controllers/activemqartemis_controller_test.go
gtully: yep, i'll have that in a later PR, I left them intentionally unchanged to make the diff visible

controllers/activemqartemis_reconciler.go
gtully: it is for migration, an existing CR can still have a configMap, that will remain till there is some mod, the mount point is then different

brusdev (approved):

PR #443, opened 2023-01-11
[#442] Allow configuring PodDisruptionBudget for broker deployment

### LINKED ISSUE #442: Allow configuring PodDisruptionBudget for broker deployment

**Additional context**

gaohoward:


docs/help/operator.md
```diff
@@ -735,3 +735,30 @@ spec:
...
+spec:
+  deploymentPlan:
+    size: 2
+    image: placeholder
+    podDisruptionBudget:
+      minAvailable: 1
+```
+
+When deploying the above custom resource the operator will create a PodDisruptionBudget
+object with the **minAvailable** set to 1. The operator also sets the proper selector
```
gtully, line 763: is the use case here that by default a deployment could be scaled to zero and we want to avoid that?

controllers/activemqartemis_controller_test.go
```diff
@@ -121,6 +122,84 @@ var _ = Describe("artemis controller", func() {
...
+				candidate.Spec.DeploymentPlan.PodDisruptionBudget = &pdb
+				candidate.Spec.DeploymentPlan.Size = 2
+			})
+
+			pdbKey := types.NamespacedName{
+				Name:      brokerCr.Name + "-pdb",
+				Namespace: defaultNamespace,
+			}
+
+			testTimeout := timeout
```
gtully, line 142: is this necessary, we expect the test to succeed once the policy is deployed, so a longer timeout is not a problem.

Maybe we need an init in the test suite and set testTimeout based on USE_EXISTING_CLUSTER, I guess that would sort it in a general sense and the entire suite could revert to testTimeout. but in this case I don't see a problem with using the existingClusterTimeout as we don't wait for that unless there is an error

controllers/activemqartemis_reconciler.go
```diff
@@ -320,6 +322,38 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessDeploymentPlan(customRes
...
+	if customResource.Spec.DeploymentPlan.PodDisruptionBudget != nil {
+		reconciler.applyPodDisruptionBudget(customResource, client, currentStatefulSet)
+	}
+}
+
+func (reconciler *ActiveMQArtemisReconcilerImpl) applyPodDisruptionBudget(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, currentStatefulSet *appsv1.StatefulSet) {
+	pdbSpec := customResource.Spec.DeploymentPlan.PodDisruptionBudget.DeepCopy()
+
+	if pdbSpec.Selector != nil {
+		clog.Info("User specified selector in PodDisruptionBudget will be ignored", "selector", pdbSpec.Selector)
```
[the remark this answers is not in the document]
gaohoward: yes that can be done.

docs/help/operator.md
gaohoward: The podDisruptionBudget is optional and it seems doesn't prevent the voluntary scale down (if user choose to do so by updating the CR).

controllers/activemqartemis_controller_test.go
gaohoward: ok, we can just use a longer timeout.

docs/help/operator.md
gtully: that makes sense, the user can still scale to 0 but kube won't do that via eviction if the budget disallows

gtully (approved):

PR #446, opened 2023-01-12
[#445] Add document on enabling metrics plugin

### LINKED ISSUE #445: Add document on enabling metrics plugin

gaohoward:


docs/help/operator.md
```diff
@@ -735,3 +735,45 @@ spec:
       secrets:
       - "my-logging-config"
 ```
+
+## Enable broker's metrics plugin
+
+The ActiveMQ Artemis Broker comes with a metrics plugin to expose metrics data. By default it is disabled.
+It can be enabled in the custom resource so the operator makes necessary configuration of the broker
+statefulset to enable tools such as Prometheus and Grafana can collect data within the kubernetes environment.
```
brusdev, line 743:
```suggestion
The ActiveMQ Artemis Broker comes with a metrics plugin to expose metrics data. The metrics data can be collected by tools as Prometheus and visualized by tools as Grafana.
By default, the metric plugin is didabled, to instruct the Operator to enable metrics for each broker Pod in a deployment, you must set the value of the `deploymentPlan.enableMetricsPlugin` property to true in the Custom Resource (CR) instance used to create the deployment. 
```

brusdev (approved):

Issue #448, opened 2023-01-13
support ordinal prefix in brokerProperties


gtully:
With N brokers deployed, it would be great to have a broker feature enable on just one of them, federation is one example.

Supporting a "ordinal-N." prefix in brokerProperties could allow ordinal specific configuration.

Routing is an existing use case where a shared config can make use of the STATEFUL_SET_ORDINAL env var to have ordinals behave differently. 
This proposal is a more general solution where arbitrary properties can be applied to an ordinal.

For example, the following client id auto shared config over two brokers:
```
crd.Spec.BrokerProperties = []string{
		"connectionRouters.autoShard.keyType=CLIENT_ID",
		"connectionRouters.autoShard.localTargetFilter=NULL|${STATEFUL_SET_ORDINAL}|-${STATEFUL_SET_ORDINAL}",
		"connectionRouters.autoShard.policyConfiguration=CONSISTENT_HASH_MODULO",
		"connectionRouters.autoShard.policyConfiguration.properties.MODULO=2",
```

could be rewritten as:

```
crd.Spec.BrokerProperties = []string{
		"connectionRouters.autoShard.keyType=CLIENT_ID",
		"connectionRouters.autoShard.policyConfiguration=CONSISTENT_HASH_MODULO",
		"connectionRouters.autoShard.policyConfiguration.properties.MODULO=2",
		"ordinal-0.connectionRouters.autoShard.localTargetFilter=NULL|0|-0",
		"ordinal-1.connectionRouters.autoShard.localTargetFilter=NULL|1|-1",
```
where it is more explicit that just two brokers are being configured, ie: the MODULO=2 is important!

While the stateful set template will always be a common base config and environment, this feature allows independence of broker configuration for static deployments.

PR #449, opened 2023-01-13
#448 - support ordinal-N prefix in brokerProperties for pod specific …

gtully:
…broker behaviours


brusdev:
@gtully nice improvement!!!

WDYT about the following alternative prefixes?
- broker-N
- $broker-N
- $N

I would like `broker-N`, i.e. `broker-0.connectionRouters.autoShard.localTargetFilter=NULL|0|-0`

gtully:
I like it, the ordinal-N is a bit clunky - however it is indeed what it is tied to, the stateful set ordinal, and it is unlikely to ever be a broker configuration attribute.

But broker-N. is indeed better. I don't want the use $ as that is the default for env var expansion.
thanks for the fast feedback :-) I will make it broker-N

gtully:
broker-<stateful set ordinal> to be more exact

controllers/activemqartemis_controller_test.go
```diff
@@ -6309,6 +6307,151 @@ var _ = Describe("artemis controller", func() {
 		}, timeout, interval).Should(BeTrue())
 	})
 
+	It("ordinal broker properties", func() {
+		ctx := context.Background()
+		crd := generateArtemisSpec(defaultNamespace)
+		crd.Spec.DeploymentPlan.Size = 2
+		crd.Spec.BrokerProperties = []string{
+			"name=BROKER_ORDINAL_0",
```
gaohoward, line 6315: Is this 'name' only overrides pod-0 or all the broker-n (except broker-1)? I think some doc need here to explain the behavior. It looks like if you scale up to 3 the 3rd doesn't use this name. (I saw some log like File /amq/extra/secrets/artemis-broker-props/broker-2 does not exist), maybe it's a bug.

gtully: properties without a prefix end up in <mount point>/broker.properties in the normal way. Properties with a prefix end up in <mount-point>/broker-N/broker-N.broker.properties and each broker looks in mount-point and mount-point/broker-${STATEFUL_SET_ORDINAL} - such that it only picks up the prefixed properties that match.

let me check with a test for scale up, I think the primary use case is static pods, however any pod that does not have extra prefixed props, will get the normal config.

gtully: and I will add some doc, in the help/operator.md - there is currently no brokerProperties section, I will fix that. thanks

gtully: it is better now

brusdev (approved):

gaohoward (approved):

Issue #455, opened 2023-01-16
Optmize deployment structure


brusdev:
There are 2 minor annoying things related to the deploy folder:
- the number prefix slows cmd line with tab (it is hard to remember the prefix number, you need to ls folder)
- the install folder name conflicts with install_opr.sh script slowing cmd line with tab

Proposal:
- remove the number prefix from the yaml files
- rename the deploy/install folder to deploy/resources

brusdev:
This improvement will be included in 1.0.8

PR #456, opened 2023-01-16
[#455] Optimize deploy folder

brusdev:
Changes:
- remove the number prefix from the yaml files
- rename the deploy/install folder to deploy/resources


gaohoward (approved):

tlbueno:
I like the idea to have the numbering in the file because I can easily see what is the order expected, in my head it is logical. Maybe the tab completion can be a bit slow down, but in bash double tabs will list the files even though any char is provided, like oc apply -f [tab][tab] , so that may reduce that slowdown on typing.

Some people can get confused with the files start by number but I personally have the inverse issue about get some confusion. I always need to cat the install script to remember the ordering. 
Also, install script is nice but to me as most of time I don't use it for different reasons, like, CRDS are already installed and I don´t want to keep apply them, or user don´t have the permissions to install the CRDs, or sometimes part of the file is already applied, etc.

Anyway I think that keep some consistence for our users has its importance but it should not block is to change something, lots of projects deprecates things, change how it works, etc things usually get deprecated or changed sometimes with breaking changes. I think it is just a matter to provide the  information about it to the users imo. 
At the same time I got a bit impressed when I saw artemis and artemis operator backwards compatibility. It is nice but I guess that can provide some challenges on dev side.

So, I think that we need more discussion with the users the may have a better view or different thoughts. For example, maybe in general users install the operator using OLM instead of use the files.

gtully:
For me the logical ordering is not necessary, it is only the grouping by kind/type that is important.
The numbering implies (once you ignore the first jumps)  an ordering that will be ignored by kube; kube it is totally async. Unless there is a pause get/list loop to ensure one piece of desired state is visible before the other the ordering is at best advice.
If a resource is invalid because it has a dependency, then that should be or will be explicit in a failed apply.
For that reason alone I think it just confuses. I am all for removing the number prefix.

gtully (approved):

michalxo:
For my usage it is also very useful hint about which files have been deployed and/or which are missing from deployment. Instead of reading/decrypting all file names, just peek at numbers and if nothing is missing anything you have it all nicely deployed. I also do not use `install_opr.sh` script at all (`kubectl apply -f crd, install folders`.

Imo, later it will be also for users easier to refer to these files by numbers, than full names, as we don't often update these files. Probably just few times and once deployment is fine, we won't touch it.

brusdev:
> For my usage it is also very useful hint about which files have been deployed and/or which are missing from deployment. Instead of reading/decrypting all file names, just peek at numbers and if nothing is missing anything you have it all nicely deployed. I also do not use `install_opr.sh` script at all (`kubectl apply -f install` folders).

How could you miss something from the deployment?
How could you peek at numbers to check if nothing is missing?

michalxo:
> How could you miss something from the deployment?
Automation can miss, when editing file (as new file) and making sure it is applied (and later removed).

>  How could you peek at numbers to check if nothing is missing?
When I am looking at logs, I can quickly check (without reading full file names) just file numbers.

Something like this for example (default files in this case)
```
2023-01-17T15:54:00.181Z INFO  [ArtemisCloudClusterOperator:78] Deploying Artemis Cluster Operator in namespace tls-tests
2023-01-17T15:54:00.182Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/crds/010_crd_artemis.yaml
2023-01-17T15:54:00.530Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/crds/020_crd_artemis_security.yaml
2023-01-17T15:54:00.578Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/crds/030_crd_artemis_address.yaml
2023-01-17T15:54:00.611Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/crds/040_crd_artemis_scaledown.yaml
2023-01-17T15:54:00.646Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/050_service_account.yaml
2023-01-17T15:54:00.668Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/080_election_role.yaml
2023-01-17T15:54:00.694Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/090_election_role_binding.yaml
2023-01-17T15:54:00.713Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/100_operator_config.yaml
2023-01-17T15:54:00.724Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/110_operator.yaml
2023-01-17T15:54:00.903Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/060_namespace_role.yaml
2023-01-17T15:54:00.924Z DEBUG [ArtemisCloudClusterOperator:82] [tls-tests] Deploying file /home/mtoth/work/repos/claire/artemis/install/070_namespace_role_binding.yaml
2023-01-17T15:54:10.954Z INFO  [ArtemisCloudClusterOperator:92] [tls-tests] Cluster operator activemq-artemis-controller-manager successfully deployed!
```

brusdev:
I see benefits and drawbacks related to the number prefix of yamls. In the absence of a stronger argument, I removed the number prefix to restore previous naming convection of yamls. Thank you all for your contributions.

PR #458, opened 2023-01-17
Default to Deployment.Size 1

ruromero:


gtully:
I think this is a very sensible default, but there must be some doc impact here? docs/help/...

controllers/activemqartemis_reconciler.go
```diff
@@ -2574,20 +2559,24 @@ func getValidCondition(cr *brokerv1beta1.ActiveMQArtemis) metav1.Condition {
 }
 
 func getDeploymentCondition(cr *brokerv1beta1.ActiveMQArtemis, podStatus olm.DeploymentStatus) metav1.Condition {
-	if cr.Spec.DeploymentPlan.Size == 0 {
+	deploymentSize := DefaultDeploymentSize
+	if cr.Spec.DeploymentPlan.Size != nil {
+		deploymentSize = *cr.Spec.DeploymentPlan.Size
+	}
```
brusdev, line 2565:
```suggestion
	deploymentSize := getDeploymentSize(cr)
```
Could `getDeploymentSize` be used here?

ruromero: Indeed, this is a refactor leftover. Thanks for the catch.

examples/artemis/artemis_single.yaml
```diff
@@ -2,7 +2,3 @@ apiVersion: broker.amq.io/v1beta1
 kind: ActiveMQArtemis
 metadata:
   name: artemis-broker
-spec:
-  deploymentPlan:
-    size: 1
```
brusdev, line 7: Should the size field be preserved  at least in 1 example?

brusdev (approved):

examples/artemis/artemis_single.yaml
ruromero: added a new example with `size: 0`

gtully (approved):

gtully:
done a rebase

gtully:
need to investigate a test, seems we are doing extra reconciles, don't know the root cause yet.

PR #464, opened 2023-01-19
[#359] further tests to validate reference to existing config and spl…

gtully:
…it console realm

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -418,6 +418,19 @@ func UpdateCRStatus(cr *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, n
 		return err
 	}
 
+	if len(cr.Status.ExternalConfigs) != len(current.Status.ExternalConfigs) {
+		return resources.UpdateStatus(client, cr)
+	}
+	if len(cr.Status.ExternalConfigs) >= 0 {
+		for _, cfg := range cr.Status.ExternalConfigs {
+			for _, curCfg := range current.Status.ExternalConfigs {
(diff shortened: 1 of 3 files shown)


gaohoward (approved):

controllers/activemqartemis_controller.go
```diff
@@ -461,6 +461,19 @@ func UpdateCRStatus(cr *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, n
...
+	if len(cr.Status.ExternalConfigs) >= 0 {
+		for _, cfg := range cr.Status.ExternalConfigs {
+			for _, curCfg := range current.Status.ExternalConfigs {
+				if curCfg.Name == cfg.Name && curCfg.ResourceVersion != cfg.ResourceVersion {
+					return resources.UpdateStatus(client, cr)
+				}
+			}
+		}
+	}
+
```
brusdev, line 476: @gtully could you move this change in a separate commit?

gtully: yep, either that or verify the external config status in one of those tests. I omitted doing that. let me revisit in the next day or so. thanks for the feedback.

gtully:
I am getting an error in make test-mk that I need to track down.

controllers/activemqartemis_controller.go
gtully: I have added a test to validate the need for that fix. thanks!

gtully:
Dom found the problem with the tests, the mad naming of crs fell down when replaced chars left a nothing between dashes... and that was an invalid field in the jolokia policy. Time to name the test crs after their line numbers!

brusdev:
@gtully I created the #477 PR to improve the suite test output and change the resources names.

brusdev (approved):

PR #465, opened 2023-01-20
Requeue if has extramounts instead of Watch resources

### LINKED ISSUE #457: extra mount watch is too simplistic and potentially costly
To track the status of referenced resources in #429 a new watch was added on config maps and secrets.

https://github.com/artemiscloud/activemq-artemis-operator/commit/c6f69135df6c6b583c5bcc4d7c683f0b2930b95c#diff-4b39fd9a9f5927e7fa6934d6892919f2ef47e81568ec6908feafcac119d52347

The filter tracks the extramounts on our CRs, even with no extramounts, we still have the watch events.

it would be better if we could have a watch per cr extra mount, 100s of watches rather than 10000s of watch events.

We would need to find a good stateless way to map watchers to CRs.extramounts such that reconcile is idempotent

ruromero:
Fix #457

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -30,12 +30,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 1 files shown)


brusdev:
Can you add the the related issue number to the commit message, i.e. `[#457] Requeue if has extraMounts instead of Watch resources` ?

ruromero:
> Requeue if has extraMounts instead of Watch resources

done

Issue #452, opened 2023-01-15
Artemis broker instance fails to start due to an invalid "broker.properties" file path in JAVA_OPTS


vkwiley:
**Describe the bug**
Artemis broker instance fails to start due to an invalid broker.properties file path

**To Reproduce**
Deploy ActiveMQArtemis object with Operator versions starting from 1.0.6

**Platform (please complete the following information):**
 - AWS EKS (vanilla EKS image)

**Additional context**
Logs:

2023-01-15 04:52:36,075 ERROR [org.apache.activemq.artemis.core.server] AMQ224097: Failed to start server: java.io.FileNotFoundException: /amq/extra/configmaps/rex-artemis-backup-props (Is a directory)

Faulty code is probably [this line](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemis_reconciler.go#L1886) (missing /broker.properties)

brusdev:
@vkwiley could you provide an ActiveMQArtemis CR to reproduce this issue?

gtully:
hmm, this is due to brokerProperties being mutable and the broker accepting a -properties arg that is a directory.
the change is: https://github.com/artemiscloud/activemq-artemis-operator/commit/8f8624a97d0c8fe55465464bf540fac9eee5bdbf#diff-32e4c985bb650a7028cf649306e140d1cf176b91d4f36dbfbce53083d0f63bd6R1675

it looks like the broker image does not have the corresponding change to support a directory rather than a file.
the change is in 2.27.1

https://github.com/apache/activemq-artemis/commit/7bc3b0280946811a65e6019f0e1ccba98097bf47

I think the fix is to use both JAVA_OPTS and JAVA_ARGS_APPEND - JAVA_ARGS_APPEND is only in 2.27.1 and can override the JAVA_OPTS to provide the required directory configuration.

https://github.com/apache/activemq-artemis/blob/2.27.1/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis#L117

vkwiley:
@brusdev  here is the manifest
`apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: artemis-broker
  namespace: qa
spec:
  acceptors:
  - name: artemis
    port: 61616
    protocols: core,amqp,stomp
  adminPassword: XXXXXX
  adminUser: XXXXXXX
  deploymentPlan:
    affinity:
[...]

PR #469, opened 2023-01-24
[#452] Support folder brokerProperties mounts

### LINKED ISSUE #457: extra mount watch is too simplistic and potentially costly

ruromero:
Fix #457

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -1912,17 +1912,7 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
 
 	clog.Info("Total volumes ", "volumes", podSpec.Volumes)
 
-	// this depends on init container passing --java-opts to artemis create via launch.sh *and* it
-	// not getting munged on the way. We CreateOrAppend to any value from spec.Env
-	var mountPoint = secretPathBase
-	if !isSecret {
-		mountPoint = cfgMapPathBase
-	}
-	javaOpts := corev1.EnvVar{
-		Name:  "JAVA_OPTS",
-		Value: brokerPropertiesConfigSystemPropValue(mountPoint, resourceName, brokerPropertiesMapData),
-	}
-	environments.CreateOrAppend(podSpec.InitContainers, &javaOpts)
+	configureBrokerPropertiesMount(podSpec, resourceName, isSecret, brokerPropertiesMapData)
 
 	var initArgs []string = []string{"-c"}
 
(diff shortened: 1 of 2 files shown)


gtully (approved):
looks good, thanks!

controllers/activemqartemis_reconciler.go
```diff
@@ -1989,12 +1979,44 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
...
+		Value: brokerPropertiesConfigFileSystemPropValue(mountPoint, resourceName, brokerPropertiesMapData),
+	}
+	environments.CreateOrAppend(podSpec.InitContainers, &javaOpts)
+
+	// older versions won't use with JAVA_ARGS_APPEND and these will prevale over JAVA_OPTS
+	javaArgsAppend := corev1.EnvVar{
+		Name:  "JAVA_ARGS_APPEND",
+		Value: brokerPropertiesConfigFolderSystemPropValue(mountPoint, resourceName, brokerPropertiesMapData),
+	}
+	environments.CreateOrAppend(podSpec.Containers, &javaArgsAppend)
```
brusdev, line 2002: Why do we need `brokerPropertiesConfigFolderSystemPropValue`? Does `brokerPropertiesConfigFileSystemPropValue` work with older and newer versions?

gtully: 2.27.1 introduced support for treating the brokerProperties url ending in a / as a directory. That version also introduced JAVA_ARGS_APPEND.
So the 'folder' via the new env var is what we want, it gives us freedom to ignore the name or the properties file. But for existing deployments that just mount a single immutable properties file, we need to respect the use of the file url.

gtully: yes, brokerPropertiesConfigFileSystemPropValue works with old and new.

brusdev: I see, I was wondering that a single implementati on could be more easy to maintain if both implementations are equivalent, anyway I have no strong opinioni on this, It LGTM thanks for this fix

brusdev (approved):

brusdev:
I'll revert this PR because `JAVA_ARGS_APPEND` is not expanded causing the following warning in the broker log:
```
WARN  [org.apache.activemq.artemis.core.server] AMQ222219: File /amq/extra/secrets/aa-controller6472a-props/broker-${STATEFUL_SET_ORDINAL} does not exist
```

Execute `ordinal broker properties` to reproduce this issue.

brusdev:
According to the [EnvVar v1 core reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#envvar-v1-core):
```
Variable references $(VAR_NAME) are expanded using the previous defined environment variables in the container and any service environment variables. If a variable cannot be resolved, the reference in the input string will be unchanged. The $(VAR_NAME) syntax can be escaped with a double $$, ie: $$(VAR_NAME). Escaped references will never be expanded, regardless of whether the variable exists or not. Defaults to "".
```

We have 2 options to fix this issue:

1. Change the artemis script adding `JAVA_ARGS_APPEND=`eval "echo $JAVA_ARGS_APPEND"``
2. Add `{name: PODNAME, valueFrom: {fieldRef: {fieldPath: metadata.name}}}` to the container env and replace `${STATEFUL_SET_ORDINAL}` with `$(PODNAME)`

gtully:
new PR https://github.com/artemiscloud/activemq-artemis-operator/pull/482

Issue #495, opened 2023-02-03
provide some validation of the contents of a login.config key in the -jaas-config suffix secret extra mount


gtully:
The format of a jaas login config is error prone, it would be nice if the operator could catch trivial errors like missing semi colons etc.
In that way, if the auth on the broker is broken due to and invalid login.config update, the status can reflect that.
For new CR's, validation can avoid some simple mistakes. It can never be perfect because the referenced login modules won't be exercised, but catching syntax errors could be a real help.

gtully:
The validation uses a regex - the regex can be configured or disabled via the operator manager env var: JAAS_CONFIG_SYNTAX_MATCH_REGEX

brusdev:
That regex is incredible!!! Are you [Neo](https://en.wikipedia.org/wiki/Neo_(The_Matrix)) ?

PR #496, opened 2023-02-03
[#495] regex login config secret validation that will trap syntax err…

gtully:
…ors and report in the cr.Status

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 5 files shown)


gtully:
I am thinking this validation, being general, it may not cover every case, it may make sense to have a operator level env var or argument that can suppress it. --disable-jaas-config-syntax-validation or some such thing?

gtully:
second thought - store the regexp in an env var, that can be enhanced or can be set to "" to disable. in that way, it is still trivial to disable but it is also possible to improve.

ruromero:
> I am thinking this validation, being general, it may not cover every case, it may make sense to have a operator level env var or argument that can suppress it. --disable-jaas-config-syntax-validation or some such thing?

Why not `strict-validation` and then don't make the jaas-validation cause the resource to not be ready? With this you can have the following conditions:
- jaas-validation: False 
- ready: True

gtully:
not sure on the value of valid false, I think the real use case here is if there is something valid in the login.confg and the regex is giving a false negative, in that case the regex needs to be improved or disabled, but continually reporting the false negative is of little use.
I am going with an env var to override - JAAS_CONFIG_SYNTAX_MATCH_REGEX="" will disable or can indeed improve it

Issue #488, opened 2023-02-01
ActiveMQ Artemis Operator should validate Secrets validity / existance when enableSSL is true


mkrutov:
**Is your feature request related to a problem? Please describe.**
enabling ssl creates amq-broker-secret-* with empty data. Users are supposed to populate these secrets, but by-defalut they are empty and Broker enters a crash loop because of that. 

**Describe the solution you'd like**
Operator should validate that Data field in Secret is not empty prior to updating/rolling out broker

gtully:
I think the first problem is that we should not create that secret.
then we validate that it exists and that it has the keys we want.

gtully:
I would like to combine this validation with https://github.com/artemiscloud/activemq-artemis-operator/pull/475 such that we don't have to verify two code paths. We can do ssl pass through in a consistent way, only one way!

gtully:
of note is that we no longer create the secret, it is up to the user to create and populate it. we do now validate the required keys and presence. the credentials secret should not be owned by the operator

mkrutov:
What is the new expected behaviour of Broker in this case (non-existant secret, enableSSL=yes)?

gtully:
see the test: the CR is invalid until the secret is created, the cr.status reflects that state in its valid condition false

https://github.com/artemiscloud/activemq-artemis-operator/commit/4a7c80abf34c6b98ef34ea0e276bfeaf0def5779#diff-d78c1465cfcf5a1ea5e8e62663d05d50bc3d12175de63981c05441f8b31636dfR1565

PR #498, opened 2023-02-13
[#488] implement validation of secrets for console sslEnabled

gtully:


controllers/activemqartemis_controller.go
```diff
@@ -259,85 +307,83 @@ func validateBrokerVersion(customResource *brokerv1beta1.ActiveMQArtemis) *metav
...
-		}
+		found := retrieveResource(cm, customResource.Namespace, &configMap, client, scheme)
 		if !found {
 			return &metav1.Condition{
 				Type:    common.ValidConditionType,
 				Status:  metav1.ConditionFalse,
 				Reason:  common.ValidConditionMissingResourcesReason,
-				Message: fmt.Sprintf("Missing required configMap %v", cm),
-			}, nil
+				Message: fmt.Sprintf("%v missing required configMap %v", ContextMessage, cm),
```
brusdev, line 324: A log line starting with a variable part its hard to search in the sources. Quotes or square brackets could help if there is no way to rephrase it

controllers/activemqartemis_controller.go
```diff
@@ -259,85 +307,83 @@ func validateBrokerVersion(customResource *brokerv1beta1.ActiveMQArtemis) *metav
...
-			Condition = AssertConfigMapContainsKey(configMap, LoggingConfigKey)
+			Condition = AssertConfigMapContainsKey(configMap, LoggingConfigKey, ContextMessage)
 			instanceCounts[loggingConfigSuffix]++
 		} else if strings.HasSuffix(cm, jaasConfigSuffix) {
 			Condition = &metav1.Condition{
 				Type:    common.ValidConditionType,
 				Status:  metav1.ConditionFalse,
 				Reason:  common.ValidConditionFailedExtraMountReason,
-				Message: fmt.Sprintf("extramount %v with suffix %v must be a secret", cm, jaasConfigSuffix),
+				Message: fmt.Sprintf("%v entry %v with suffix %v must be a secret", ContextMessage, cm, jaasConfigSuffix),
```
brusdev, line 335: A log line starting with a variable part its hard to search in the sources. Quotes or square brackets could help if there is no way to rephrase it

brusdev (commented):
Nice wok!!!

controllers/activemqartemis_controller.go
gtully: that arg is the context message, which is the prefix based on the caller - extra mounts secrets or config maps. In practice it won't be a variable, it will be one of the CR attributes.

gtully: maybe I should have used %s?

gtully: that would be more clear, it is not any var

gtully: let me fix that, thanks

gtully: done!

brusdev (approved):

PR #504, opened 2023-02-14
[#499] Allow SSL secret to be reused in the CR

### LINKED ISSUE #499: Allow SSL secret to be reused in the CR
**Describe the bug**
It's a common case to use only one certificate for all exposed endpoints by the broker.

For example, when applying a new Custom Resource and exposing the console using SSL, if you want to use the same certificate for both acceptor and the console, you need to create different secrets.

**To Reproduce**
2. '...'

 - OS: [e.g. Fedora]
 - Version [e.g. 32]
 - Go [e.g. 1.13.4

**Expected behavior**

**Screenshots**

**Additional context**

gaohoward:


controllers/activemqartemis_reconciler.go
```diff
@@ -1514,15 +1549,24 @@ func MakeVolumes(customResource *brokerv1beta1.ActiveMQArtemis, namer Namers) []
...
 	}
 
 	if customResource.Spec.Console.SSLEnabled {
 		clog.V(1).Info("Make volumes for ssl console exposure on k8s")
 		secretName := namer.SecretsConsoleNameBuilder.Name()
-		volume := volumes.MakeVolume(secretName)
-		volumeDefinitions = append(volumeDefinitions, volume)
+		if "" != customResource.Spec.Console.SSLSecret {
+			secretName = customResource.Spec.Console.SSLSecret
+		}
```
brusdev, line 1564: How is this related to #499?

controllers/activemqartemis_controller_test.go
```diff
@@ -126,6 +126,266 @@ var _ = Describe("artemis controller", func() {
...
+					Name:      serviceName,
+					Namespace: defaultNamespace,
+				}
+				Eventually(func(g Gomega) {
+					g.Expect(k8sClient.Get(ctx, serviceKey, &service)).To(Succeed())
+				}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
+				if isOpenshift {
+
+					By("check route is created for console")
```
brusdev, line 271: We could just validate that the pod is ready, why do we need to validate ingress or route resources?

controllers/activemqartemis_reconciler.go
```diff
@@ -1262,6 +1259,40 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessResources(customResource
...
+
+		if !reflect.DeepEqual(deployedRoute.Spec.TLS, requestedRoute.Spec.TLS) {
+			reqLogger.V(1).Info("Route tls not equal", "deployed", deployedRoute.Spec.TLS, "requested", requestedRoute.Spec.TLS)
+			return false
+		}
+		reqLogger.V(1).Info("routes are equal")
+		return true
+	})
+
```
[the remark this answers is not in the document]
gaohoward: This is to improve reconcile of route resources. The route has some other fields that are auto filled by the openshift, next time when it reconcile again the "deployed" route will be "different" from the expected route object, causing unnecessary resource updates. (I think the same is true with ingress, which is why we added a comparator for ingress too).

controllers/activemqartemis_reconciler.go
```diff
@@ -1501,8 +1533,11 @@ func MakeVolumes(customResource *brokerv1beta1.ActiveMQArtemis, namer Namers) []
 		if "" != acceptor.SSLSecret {
 			secretName = acceptor.SSLSecret
 		}
-		volume := volumes.MakeVolume(secretName)
-		volumeDefinitions = append(volumeDefinitions, volume)
+		if _, ok := secretVolumes[secretName]; !ok {
+			volume := volumes.MakeVolume(secretName)
+			volumeDefinitions = append(volumeDefinitions, volume)
+			secretVolumes[secretName] = secretName
+		}
```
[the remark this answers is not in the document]
gaohoward: yeah tha't a good idea , will do.

controllers/activemqartemis_reconciler.go
gaohoward: I believe this is a bug (it should have been there, otherwise the user specified secret will be ignored).

controllers/activemqartemis_controller_test.go
gaohoward: I do think that's necessary to check further the ingress or routes are created. Pod is ready only means pod is up it doesn't mean we fixed the shared secret problem.

controllers/activemqartemis_reconciler.go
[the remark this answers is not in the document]
brusdev: Ingress comparator was added because the default comparator doesn't include it while it already includes a comparator for Route, for further details see https://github.com/RHsyseng/operator-utils/blob/v1.4.10/pkg/resource/compare/defaults.go#L98

I see the test `console and acceptor share one secret` passing also without this comparator.

The default Route comparator seems similar to the one added, what does the custom comparator do that the default doesn't?

pkg/resources/k8s_actions.go
```diff
@@ -109,7 +109,8 @@ func Update(client client.Client, clientObject client.Object) error {
 				// "StatefulSet.apps is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden"}
 				reqLogger.V(1).Info("Deleting on failed updating "+objectTypeString, "obj", clientObject, "Forbidden", err)
 				err = Delete(client, clientObject)
-				break
+			} else {
+				reqLogger.Error(err, "Got status error")
 			}
```
brusdev, line 114: I see the test `console and acceptor share one secret` passing also without this change, why is this change required?

gaohoward: It's a mistake in the code. I corrected it.

brusdev (approved):

Issue #505, opened 2023-02-14
Operator upgarde doesn't work as expected


brusdev:
**Describe the bug**
The broker container image is updated also if the `spec.upgardes.enabled` property is false.
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata: 
  name: ex-aao
spec: 
  deploymentPlan: 
    image: placeholder
    size: 1
  upgrades: 
    enabled: false
    minor: false
```

**To Reproduce**
1. Install the operator version v1.0.7
2. Deploy a basic ActiveMQArtemis CR
3. Check the broker container image is quay.io/artemiscloud/activemq-artemis-broker-kubernetes:1.0.11
4. Update the operator to version v1.0.10
3. Check the broker container image is quay.io/artemiscloud/activemq-artemis-broker-kubernetes:1.0.14

**Expected behavior**
The broker container image should not change if the `spec.upgardes.enabled` property is false.

gtully:
I think that is an invalid CR, that CR tells me that the operator decides what version to deploy via the 'placeholder' and that upgrades are expected.
If upgrades are not required, then either the image tag is provided in the init and broker image fields, or some version information is provided in the version field.
Part of the reason why I deprecated the Upgrades field is that it has the wrong default. There are patches and cve updates in openshift that are outside of the control of the operator. OLM handles that at the bundle level. 
It can never mean "do not restart the broker", which is a side effect of upgrades that some folks want to avoid. 
(now, there is a case to make for restarts being the norm, which is fine too but separate.)

I think it is valid to not want upgrades and that can be configured in the floating version string, x.y or x, or x.y.z to lock it down. The point of the version string is to lock it down. But that is limited to the version information that the operator has, the window of supported versions. A CR can become invalid if a version grows out of range with time. 

The problem I see with turning temporal information like the value of the place holder today, into some persisted ( desired) thing that we need to respect in the future, is that the only good place for that information is the CR.
We could update the status, but that is not fully in our control.

I think the image attributes are the first port of call here, those sha's are locked in stone, that is the only way to avoid upgrades completely. 
Then the version string can come next, but that will be bound by the operator bundle, patches and cve updates will still be in the mix. It is directive to upgrade or not within the bounds of supported versions. In the absence of a version, upgrades are assumed.

brusdev:
I think that asking users to set the image fields we are asking users to do an operator task.
[...]

PR #506, opened 2023-02-14
[#505] Fix ActiveMQ Artemis upgrades

brusdev:


gtully:
I had previously deprecated the upgrades attribute thinking that a version string would be sufficient. It may be that there is a need for a postfix to lock it down, x.y.z-final or some thing like that. Maybe the boolean is still necessary if a postfix is ugly or not sufficient.
see: https://github.com/artemiscloud/activemq-artemis-operator/commit/1cb7aacca805f5efdbc8b75b803c132fb6c6cc90#diff-1160c62501def02fcbb55235053bd8eb8a3c6b8e715b3350b4c844d6c3f1a85eR352

gtully:
recalling some more thinking, x.y.z can only go so far, cve or patches that come via new bundles and via operator updates will leave version information out of date. Tags can move. The only way to avoid image update/rollout in those cases is to specify the image tag in the CR. Maybe we need to make it easier to identify the image that is in play such that the CR can be updated with the correct information with by the user.
That could be a use case for putting it in the status, but it would be the current image, not the desired image that is in the status.

gtully:
actually, maybe the status contains:
observed image: ...
upgradeable = no (if image is specified) |major|minor (depending on the presence and specificity of the  version string).
I wonder if it would make a good Condition?

brusdev:
Regardless of this PR I think that the CR status should include the current ActiveMQ Artemis version. The images are a bit raw for a user.

brusdev:
Closed, see the issue #505 for further details.

Issue #516, opened 2023-02-20
add the observed image urls and version information to the status 


gtully:
To avoid updates, either the image urls or the version string must be specified in the CR. 
To make it trivial, it would be helpful if the current observed values were present in the Status sub resources.
Then the process of locking down a broker container would be a case of cut and paste.

PR #521, opened 2023-02-22
[#516] reflect upgrade version and image urls in the status, validate…

gtully:
… and simplify logic around version matching and upgrades


pkg/utils/common/common.go
```diff
@@ -135,53 +135,62 @@ func ResolveWatchNamespaceForManager(oprNamespace string, watchNamespace string)
...
+				} else if *minor == versions[i].Minor {
+					if patch == nil {
+						break
+					} else if *patch == versions[i].Patch {
+						break
+					}
+				}
+			}
+		}
 	}
```
brusdev, line 166: The semver range could simplify the search
```suggestion
	// Build matchDesiredRange from desired
	matchDesiredRange := semver.MustParseRange(">=0.0.0")
	if desired != "" {
		var parseErr error

		// Add the wildcard suffix if the semantic version is incomplete
		if strings.Count(desired, ".") < 2 && !strings.HasSuffix(desired, ".x") {
			matchDesiredRange, parseErr = semver.ParseRange(desired + ".x")
		} else {
			matchDesiredRange, parseErr = semver.ParseRange(desired)
		}

		if parseErr != nil {
			return nil
		}
	}

	for i := len(versions) - 1; i >= 0; i-- {
		if matchDesiredRange(versions[i]) {
			return &versions[i]
		}
	}
```

gtully: am not convinced, I see it as a match/pruning process rather than a range issue. Maybe if we supported ranges in Version it would make sense. 
But thanks for the suggestion. We may use it in the future.

gtully: or are you thinking that our version list may contain patch and descriptive semantic versions in the future. like 2.20.7-pre-01 or some other semver format and we are future proofing?

brusdev: I suggested this change only because it seems we are doing the same thing of blang semver range but I see your point, I don't have a strong opinion on that so feel free to reject this suggestion.

gtully: thanks for the feedback, much appreciated.

api/v1beta1/activemqartemis_types.go
```diff
@@ -575,16 +575,42 @@ type ActiveMQArtemisStatus struct {
...
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="BrokerVersion",xDescriptors="urn:alm:descriptor:text"
+	BrokerVersion string `json:"brokerVersion,omitempty"`
+
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Image URI",xDescriptors="urn:alm:descriptor:org.w3:link"
+	Image string `json:"image,omitempty"`
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="InitImage URI",xDescriptors="urn:alm:descriptor:org.w3:link"
+	InitImage string `json:"initImage,omitempty"`
+
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Upgrade Status"
+	Upgrade UpgradeStatus `json:"upgrade,omitempty"`
```
brusdev, line 596: Should Upgrade be a property of ActiveMQArtemisStatus?

gtully: maybe, it has Status as the postfix. I think that would be better, it is not just a function of Version. thanks!

Issue #124, opened 2022-01-10
Configuring autoscaling of broker pods.


sachinsaju:
**Is your feature request related to a problem? Please describe.**
We can manually scale up and scale down artemis brokers via ActiveMQArtemis custom resource. But when I add horizontal pod autoscaling to artemis statefulset, initially brokers willl be scaled and after the scaling ,the operator will scale down to the number specified in ActiveMQArtemis CR. So HPA tries to bring up more pod and ActiveMQArtemis CR tries to bring down pods one after another.

**Describe the solution you'd like**
I would like to have Kubernetes handle autoscaling the artemis pods based on resources and Artemis operator manage the message migration when pod goes down. We can either have some fields in ActiveMQArtemis CR to mention HPA details or have a different custom resource for autoscaling.

**Additional context**
I didn't find any detail on HPA autoscaling in the documentation. If I miss something, please let me know.

brusdev:
@sachinsaju could you explain why the horizontal pod autoscaling is required in your use case?

gtully:
I think we should support the scale sub resource. Particularly for the scale to zero use case, where this is a single broker that can be auto scaled between 0 and 1 with the same PV.
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

gtully:
the HPA use case seems more involved, with metrics etc, and to make it work there would need to be a cluster service that can expose all brokers. As a general rule, partitioning is necessary to have brokers work better at scale with point to point messaging. For multicast, there can be benefits with bridges/federation rather than a plain cluster. this may be non trivial to configure.

To enable the possibility of HPA we need to come up with a selector path, should it contain the ArtemisCR name?

https://book.kubebuilder.io/reference/generating-crd.html?highlight=subre#scale

sachinsaju:
@brusdev There is chance of large number of active queues and messages potentially in my use case. Therefore, If pods can be scaled up or down based on load, it would be helpful . In that way , I can ensure better load balancing to artemis pods(thereby increasing overall performance of apps). HPA is a common feature in most of the k8s tool installations. We can avoid manual scale up or scale down.

gtully:
https://github.com/artemiscloud/activemq-artemis-operator/pull/526 via support for a scale subresource will get the ball rolling on this capability.

PR #526, opened 2023-02-28
[#124] support scale subresource for scale to zero

gtully:


api/v1beta1/activemqartemis_types.go
```diff
@@ -574,6 +574,9 @@ type ActiveMQArtemisStatus struct {
 	//we probably use Deployments as operatorHub shows invalid field podStatus
 	//see 3scale https://github.com/3scale/3scale-operator/blob/8abbabd926616b98db0e7e736e68e5ceba90ed9d/apis/apps/v1alpha1/apimanager_types.go#L87
 
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Deployment Plan Size"
```
brusdev, line 577: Are you not adding the OLM Descriptor on purpose?
```suggestion
	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Deployment Plan Size",xDescriptors="urn:alm:descriptor:text"
```

gtully: I was not really sure why I would add text to this field, I think it will be fine with the default representation of an integer, possibly as a string.

brusdev (approved):

PR #536, opened 2023-03-07
[#535] strip white space around property key-value entries to match j…

gtully:
…ava properties loader

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -10,6 +10,7 @@ import (
 	"hash/adler32"
 	osruntime "runtime"
 	"sort"
+	"unicode"
 
 	"github.com/blang/semver/v4"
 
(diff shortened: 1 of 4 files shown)


controllers/activemqartemis_reconciler.go
```diff
@@ -2943,7 +2944,12 @@ func alder32FromData(data []byte) string {
 }
 
 func appendNonEmpty(propsKvs []string, data []byte) []string {
-	keyValue := strings.TrimSpace(string(data))
+	keyValue := strings.Map(func(r rune) rune {
+		if unicode.IsSpace(r) {
+			return -1
+		}
+		return r
+	}, string(data))
```
brusdev, line 2952: What about if the property value includes whitespaces?

gtully: that will be broken, good catch. I will need to just trim space around the '='
Actually this points to a bigger problem, the alder calculation does not account for the numerous escapes that a java properties file will support. with brokerProperties this is less of a problem b/c the subset is minimal, but with jaas user and role properties there may be more need for escapes etc. in the future. We may need to keep this in mind.

gtully: fixed up the doc to point out that the format is actually strings or lines for a java properties file. so the relevant characters that are significant to java properties encoding (space,colon and equals) need to be escaped. We now process those escapes when calculating the checksum.

brusdev (approved):

PR #558, opened 2023-04-04
[NO-ISSUE] Fix script issue with deploy/cluster_wide_install_opr.sh

gaohoward:
The script uses KUBE_CLI for client tool (oc or kubectl). However there are several places where the kubectl is used directly instead of using $KUBE_CLI


brusdev (approved):

brusdev:
@howardgao good catch!!!

Issue #565, opened 2023-04-19
Jolokia client should use pod's hostname instead of ip address


gaohoward:
Currently jolokia client uses pod ip to contact broker for managment operations.
This doesn't work when broker web console is configured with ssl + sniHost check.
The pod IP is dynamically allocated and is not suitable for such a case. Instead it should
be using host name.

moritzschmitz-oviva:
What is the workaround here? For me this means the operator cannot do anything on the broker.

gaohoward:
In next release the sniHostCheck will be disabled by default. See this commit https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/commit/ba5d069c5e1927a69f6c132c24f3e0332fbbf317
right now you can manually do it using custom init image.

PR #566, opened 2023-04-19
[#565] Jolokia client should use pod's hostname instead of ip address

gaohoward:


pkg/utils/jolokia_client/jolokia_client.go
```diff
@@ -80,7 +83,16 @@ func GetBrokers(resource types.NamespacedName, ssInfos []ss.StatefulSetInfo, cli
 					jolokiaUser, jolokiaPassword, jolokiaProtocol := resolveJolokiaRequestParams(resource.Namespace, client, client.Scheme(), jolokiaSecretName, &containers, podNamespacedName, statefulset, info.Labels)
 
 					reqLogger.Info("New Jolokia with ", "User: ", jolokiaUser, "Protocol: ", jolokiaProtocol, "broker ip", pod.Status.PodIP)
-					artemis := mgmt.GetArtemis(pod.Status.PodIP, "8161", "amq-broker", jolokiaUser, jolokiaPassword, jolokiaProtocol)
+
+					hostToUse, err := resolvePodHostName(pod.Status.PodIP, pod.Name)
+					if err != nil {
+						reqLogger.V(1).Info("failed dns lookup, use IP address", "podIp", pod.Status.PodIP, "error", err)
+						hostToUse = pod.Status.PodIP
+					}
```
brusdev, line 91: According to the [Kubernetes DNS documentation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields) we should be able to build the hostToUse from the pod hostname and subdomain, i.e.
```suggestion
					hostToUse := pod.Spec.Hostname + pod.Spec.Subdomain + pod.Namespace
```

gaohoward: Yes we can have the hostName and subdomain, but I couldn't find where we can get the domain name (default is .cluster.local). The hostname we currently using takes the form <pod-hostname>.<subdomain>.<namespace>.<domain-name>.  Using DNS lookup I think we can get them all.

brusdev: I think that we don't need the domain-name inside the cluster

gaohoward: Yes but having it there does no harm. If we don't need the domain we need to change the launch.sh script and drain.sh script because they are using `'hostname -f` to get the full name.

gaohoward: another point is that we should trust DNS look up, instead of manually assemble the host name. It's safe for us.

gaohoward: just thought that in the cert the SAN names requires the domain name.

brusdev: > Yes but having it there does no harm. If we don't need the domain we need to change the launch.sh script and drain.sh script because they are using `'hostname -f` to get the full name.

I think that having the domain in the hostname could increase the difficulty on creating a certificate, so if not required I would remove it from the hostname but I don't have a strong opinion on that.

> another point is that we should trust DNS look up, instead of manually assemble the host name. It's safe for us.

How could we be sure that `hostname -f` and `resolvePodHostName` will return the same result?
If we build it with the same Kubernetes rules we should be sure that those matches and could provide instructions to users how to create their certificates.

> just thought that in the cert the SAN names requires the domain name.

If we use `<pod-hostname>.<subdomain>` as hostname then `*.<subdomain>` should work
and if we use `<pod-hostname>.<subdomain>.<namespace>.<cluster-domain>` as hostname then `*.<subdomain>.<namespace>.<cluster-domain>` should work

gaohoward: Re: How could we be sure that hostname -f and resolvePodHostName will return the same result?
If we build it with the same Kubernetes rules we should be sure that those matches and could provide instructions to users how to create their certificates.

We know hostname -f will return one entry and resolvePosHostName will pick up from the DNS lookup list the one matches it. 

I think we should use the full qualified host name, even if the short one works.

test/resources/broker_sni.ks
gtully, line 1: would it be possible to generate this cert on the fly, or use a kube api to generate it? committed certs always get stale. I guess at a min, add a readme with the commands to create/recreate

gaohoward: I believe there are some modules that provide apis for generate the cert (self-signed at least). Need to explore a bit.

gtully:
It looks like getting the cluster domain has to be done via dns lookup, and if we have a default to cluster.local our envTests can still work with the test-proxy to get the fqdn into the cluster.
In that way we can avoid all references to IP's in for the jolokia and artemis client.

This sort of thing, to resolve the cluster domain once:
```
var ClusterDomain *string

func getClusterDomain() string {
	if ClusterDomain == nil {
		apiSvc := "kubernetes.default.svc"
		cname, err := net.LookupCNAME(apiSvc)
		if err == nil {
			clusterDomain := strings.TrimPrefix(cname, apiSvc)
			clusterDomain = strings.TrimSuffix(clusterDomain, ".")
			ClusterDomain = &clusterDomain
		} else {
			defaultClusterDomain := "cluster.local"
			ClusterDomain = &defaultClusterDomain
		}
	}

	return *ClusterDomain
}
```

And them make up fqdn's from the cr name, following our naming convention for the headless service we always create.

this means that a jolokia client will need a refrernce to the CR, and that will be sufficient.
We may still need the containers for the env vars etc until we get service account auth up and running.

to get the jolokias for a cr, something like:
```
func GetBrokersFromDNS(crName string, namespace string, size int32, client rtclient.Client) []*JkInfo {
	var artemisArray []*JkInfo = nil
	var i int32 = 0

	clusterDomain := getClusterDomain()
	for i = 0; i < size; i++ {
		// from NewHeadlessServiceForCR2 and
		// https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records
		ordinalFqdn := fmt.Sprintf("%s-ss-%d.%s-hdls-svc.%s.svc.%s", crName, i, crName, namespace, clusterDomain)
		jkInfo := JkInfo{
			Artemis: mgmt.GetArtemisForRBACAgent(ordinalFqdn),
			IP:      ordinalFqdn,
			Ordinal: strconv.FormatInt(int64(i), 10),
		}
		artemisArray = append(artemisArray, &jkInfo)
	}
	return artemisArray
}
```

gaohoward:
> var ClusterDomain *string
> 
> func getClusterDomain() string {
> 	if ClusterDomain == nil {
> 		apiSvc := "kubernetes.default.svc"
> 		cname, err := net.LookupCNAME(apiSvc)
> 		if err == nil {
> 			clusterDomain := strings.TrimPrefix(cname, apiSvc)
> 			clusterDomain = strings.TrimSuffix(clusterDomain, ".")
> 			ClusterDomain = &clusterDomain
> 		} else {
> 			defaultClusterDomain := "cluster.local"
> 			ClusterDomain = &defaultClusterDomain
> 		}
> 	}
> 
> 	return *ClusterDomain
> }

Thanks Gary, that's a better solution.

pkg/utils/jolokia_client/jolokia_client.go
```diff
@@ -39,60 +41,72 @@ type JkInfo struct {
...
+		// https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records
+		ordinalFqdn := fmt.Sprintf("%s-ss-%d.%s-hdls-svc.%s.svc.%s", crName, i, crName, namespace, clusterDomain)
+
+		pod := &corev1.Pod{}
+		podNamespacedName := types.NamespacedName{
+			Name:      namer.CrToSS(crName) + "-" + strconv.Itoa(int(i)),
+			Namespace: namespace,
+		}
+
+		reqLogger.Info("Trying finding pod " + pod.Name)
```
gtully, line 82: looks good,
I think reducing the verbosity of the logging will help, v(2) or something... really we only need to log when we have a problem.

gtully (approved):

brusdev (approved):

brusdev (commented):
I got the following test failure:
```
  [FAILED] in [It] - /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:4776 @ 06/14/23 14:11:29.429
  << Timeline

  [FAILED] Timed out after 180.000s.
  Expected success, but got an error:
      <*errors.errorString | 0xc001697810>: {
          s: "Assertion in callback at /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:4771 failed:\nExpected\n    <v1.ConditionStatus>: Unknown\nto equal\n    <v1.ConditionStatus>: False",
      }
      Assertion in callback at /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:4771 failed:
      Expected
          <v1.ConditionStatus>: Unknown
      to equal
          <v1.ConditionStatus>: False
  In [It] at: /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:4776 @ 06/14/23 14:11:29.429
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Summarizing 1 Failure:
  [FAIL] artemis controller BrokerProperties [It] expect error message on invalid property value
  /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemis_controller_test.go:4776

Ran 65 of 149 Specs in 591.674 seconds
FAIL! -- 64 Passed | 1 Failed | 0 Pending | 84 Skipped
```

brusdev:
Never mind, the test failure was due to an environmental issue, I execute make test-mk-v without any failures

brusdev:
I execute also `make test-mk-do-v` without any failures

brusdev:
Great work!!!

PR #568, opened 2023-04-19
[NO-ISSUE] Security Controller uses Scaldown Logger

gaohoward:
It should use its own logger


brusdev:
The fix for this issue will be included in v1.0.12

PR #573, opened 2023-04-24
[NO-ISSUE] cleanup taint in test after it's completed.

gaohoward:


controllers/activemqartemisscaledown_controller_test.go
```diff
@@ -287,6 +287,15 @@ var _ = Describe("Scale down controller", func() {
...
 
 			Expect(k8sClient.Delete(ctx, createdCrd)).Should(Succeed())
+
+			//clean up taint
+			nodes := &corev1.NodeList{}
+			Expect(k8sClient.List(ctx, nodes, &client.ListOptions{})).Should(Succeed())
+			Expect(len(nodes.Items) > 0).Should(BeTrue())
+
+			node := nodes.Items[0]
+			node.Spec.Taints = nil
```
gtully, line 297: there is a revert taint section above, https://github.com/artemiscloud/activemq-artemis-operator/blob/d2270c75113d3e58dba07178b01c93454ace3767/controllers/activemqartemisscaledown_controller_test.go#L262

should that use an nill in place of the emtpy array?

It seems as if the earlier work should be sufficient

gaohoward: hmm, I didn't notice that. re-running the test again it seems the taint is properly cleaned up.
I'll close that. Thanks!

PR #587, opened 2023-05-08
[NO-ISSUE] Improve cert gen in tests

gaohoward:
- Generate tls secret without using static keystore/trustore files
- Reuse the tls secret generation code in tests


gtully:
we need to find a place to include the license of the dependency, see the binary distro requirement at: https://pkg.go.dev/software.sslmate.com/src/go-pkcs12?tab=licenses

gaohoward:
@gtully @brusdev 
How about this location: https://github.com/artemiscloud/activemq-artemis-operator/pull/587/commits/f40a98b90ee66e7ac9dc64f73423fb24ce123ba2
I see [k8s](https://github.com/kubernetes/kubernetes/tree/master/LICENSES) has a folder LICENSES however I don't know whether *vendor* or *third_party* is a proper category for us. I use *dependencies* instead. wdyt?

brusdev:
I would move the `licenses` folder in the `deploy` folder because we should include it in the distribution package. ActiveMQ Artemis does a similar thing, see https://github.com/apache/activemq-artemis/tree/main/artemis-distribution/src/main/resources/licenses/bin

gaohoward:
> I would move the `licenses` folder in the `deploy` folder because we should include it in the distribution package. ActiveMQ Artemis does a similar thing, see https://github.com/apache/activemq-artemis/tree/main/artemis-distribution/src/main/resources/licenses/bin
good point. I'll move the folder into deploy. thanks!

controllers/common_util_test.go
```diff
@@ -569,3 +576,117 @@ func GetOperatorLog(ns string) (*string, error) {
...
+			Country:       []string{"US"},
+			Province:      []string{"MA"},
+			Locality:      []string{"Westford"},
+			StreetAddress: []string{"Sunset Blvd"},
+			PostalCode:    []string{"12345"},
+			CommonName:    "artemiscloud",
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().AddDate(10, 0, 0),
+		IsCA:      true,
```
brusdev, line 617: Those settings don't match the previous broker.ks:
```
-dname "CN=ArtemisCloud Broker, OU=Broker, O=ArtemisCloud" -validity 36500 -ext bc=ca:false -ext eku=sA -ext "san=dns:*.apps.artemiscloud.io,dns:*.tests.artemiscloud.io"
```

gaohoward:
> I would create an issue to have a tracker for the new license.

@brusdev PR updated. Pls review again. thx.

deploy/LICENSES/dependencies/software.sslmate.com/src/go-pkcs12/LICENSE
brusdev, line 1: @howardgao I see 2 main differences with ActiveMQArtemis licenses:

1. the folder name is lowercase, i.e. `license`
2. the license file name include the dependency name so no sub folders are required, i.e. [licenses/LICENSE-angular-cookies.txt](https://github.com/apache/activemq-artemis/blob/main/artemis-distribution/src/main/resources/licenses/bin/licenses/LICENSE-angular-cookies.txt)

brusdev (approved):
Nice work, this will speed up creating new test with certificates!!!

Issue #589, opened 2023-05-09
PodDisruptionBudget is not part of the deployed resources list


gaohoward:
**Describe the bug**
Operator doesn't respect the existing PodDisruptionBudget object because it's not on the list of deployed resources and tries to create new one every time it goes through reconciliation process. it results in following error log:
```


2023-03-23T06:25:14.272Z	INFO	controller_v1beta1activemqartemis	Creating 	{"ActiveMQArtemis Name": "broker-app", "kind ": "v1.PodDisruptionBudget", "named ": "broker-app-pdb"}
2023-03-23T06:25:14.272Z	INFO	package k8s_actions	Creating new *v1.PodDisruptionBudget	{"ActiveMQArtemis Name": "broker-app-pdb", "Namespace": "broker-version-test"}
2023-03-23T06:25:14.280Z	ERROR	package k8s_actions	Failed to create new *v1.PodDisruptionBudget	{"ActiveMQArtemis Name": "broker-app-pdb", "Namespace": "broker-version-test", "error": "poddisruptionbudgets.policy \"broker-app-pdb\" already exists"}
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).createRequestedResource
	/workspace/app/controllers/activemqartemis_reconciler.go:1163
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).createResource
	/workspace/app/controllers/activemqartemis_reconciler.go:1146
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).ProcessResources
	/workspace/app/controllers/activemqartemis_reconciler.go:1126
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).Process
	/workspace/app/controllers/activemqartemis_reconciler.go:158
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconciler).Reconcile
	/workspace/app/controllers/activemqartemis_controller.go:163
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/workspace/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
[...]

PR #590, opened 2023-05-09
[#589] PodDisruptionBudget is not part of the deployed resources list

gaohoward:


controllers/activemqartemis_controller.go
```diff
@@ -121,7 +121,7 @@ type ActiveMQArtemisReconciler struct {
 //+kubebuilder:rbac:groups=monitoring.coreos.com,namespace=activemq-artemis-operator,resources=servicemonitors,verbs=get;create
 //+kubebuilder:rbac:groups=apps,namespace=activemq-artemis-operator,resources=deployments/finalizers,verbs=update
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,namespace=activemq-artemis-operator,resources=roles;rolebindings,verbs=create;get;delete
-//+kubebuilder:rbac:groups=policy,namespace=activemq-artemis-operator,resources=poddisruptionbudgets,verbs=create;get;delete
+//+kubebuilder:rbac:groups=policy,namespace=activemq-artemis-operator,resources=poddisruptionbudgets,verbs=create;get;delete;list;watch
```
gtully, line 124: I see that we need the list permission, not sure that we need the watch? We won't watch for changes, but we do want to be able to update

controllers/activemqartemis_controller_test.go
```diff
@@ -442,10 +442,26 @@ var _ = Describe("artemis controller", func() {
 			pdbObject := policyv1.PodDisruptionBudget{}
 			Eventually(func(g Gomega) {
 				g.Expect(k8sClient.Get(ctx, pdbKey, &pdbObject)).Should(Succeed())
+
+				resMap, err := getDeployedResources(createdCr, k8sClient)
```
gtully, line 446: I think this use of getDeployedResources is too low level. Ideally these tests just work on the kube api.
A test that verifies that we can update the PDB when the CR updates may be at a better level of abstraction. Without the fix, an update will always fail b/c we see it as a create in error. 
A test that creates a CR with PDB, then updates the CR with a change that is relevant to the PDB and verifies that the PDB resource version is updated (!= to the original) will cover this nicely.

deploy/cluster_role.yaml
```diff
@@ -174,6 +174,8 @@ rules:
   - create
   - delete
   - get
+  - list
+  - watch
```
gtully, line 178: not seeing the need for watch!

gtully (changes requested):

controllers/activemqartemis_controller.go
gaohoward: Yes that's strange if I don't add the `watch` permission, the operator seems work but the log will show some warning message regarding this permission periodically.

controllers/activemqartemis_controller_test.go
gaohoward: yeah make sense. will add it.

deploy/cluster_role.yaml
gaohoward: Not sure why as replied above, I need more investigation.

gaohoward: yes I think we need the watch permission. If I remove the watch, the pdb created the first time are not quickly sync'ed in the cache so the controller take a long time to know that there is already a pdb created (so the controller will try a few times to create it and get 'already exists' error in each following reconcile loop).

controllers/activemqartemis_controller_test.go
```diff
@@ -485,6 +486,57 @@ var _ = Describe("artemis controller", func() {
...
+
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, pdbKey, &pdbObject)).Should(Succeed())
+				g.Expect(*pdbObject.Spec.MinAvailable).To(BeEquivalentTo(intstr.FromInt(2)))
+				g.Expect(prevResourceVersion).NotTo(Equal(pdbObject.ResourceVersion))
+			}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
+			CleanResource(brokerCr, brokerCr.Name, defaultNamespace)
+		})
 	})
```
gtully, line 540: nice test!

deploy/cluster_role.yaml
```diff
@@ -174,6 +174,9 @@ rules:
   - create
   - delete
   - get
+  - list
+  - update
+  - watch
```
gtully, line 179: still don't see the need for the watch permission, is there a way to remove?

gtully (approved):

deploy/cluster_role.yaml
gaohoward: We are not directly use it. It is used for cache update I think.

controllers/activemqartemis_controller.go
```diff
@@ -121,7 +121,7 @@ type ActiveMQArtemisReconciler struct {
 //+kubebuilder:rbac:groups=monitoring.coreos.com,namespace=activemq-artemis-operator,resources=servicemonitors,verbs=get;create
 //+kubebuilder:rbac:groups=apps,namespace=activemq-artemis-operator,resources=deployments/finalizers,verbs=update
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,namespace=activemq-artemis-operator,resources=roles;rolebindings,verbs=create;get;delete
-//+kubebuilder:rbac:groups=policy,namespace=activemq-artemis-operator,resources=poddisruptionbudgets,verbs=create;get;delete
+//+kubebuilder:rbac:groups=policy,namespace=activemq-artemis-operator,resources=poddisruptionbudgets,verbs=create;get;delete;list;update;watch
```
gtully, line 124: this watch permission. I don't think that is necessary. Is it? we should not watch it b/c we already watch the CR.

gaohoward: this watch permission is needed, but not directly. If the permission is absent, the operator got issues with client cache updates. (i.e. when you create the resource the cache won't get updated quickly). (same thing with other resources like services which we don't directly watch them).

Issue #616, opened 2023-06-15
Warning: would violate PodSecurity "restricted:v1.24"


brusdev:
I got the following warning installing the operator on OpenShift 4.12.7 with the script `./deploy/install_opr.sh`:
```
Warning: would violate PodSecurity "restricted:v1.24": unrestricted capabilities (container "manager" must set securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container "manager" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

brusdev:
This bug will be fixed in v1.0.12

brusdev:
The fix for this issue will be reverted in v1.0.14 becaus it is causing #625

PR #617, opened 2023-06-15
[#616] Restrict manager container security context

brusdev:
Drop all cpabilities and set seccomp profile to RuntimeDefault to avoid the warning: would violate PodSecurity "restricted:v1.24"


brusdev:
I executed make test-mk-v and make test-mk-do-v without failures.

Issue #619, opened 2023-06-19
update operator-utils dependency to v1.4.12


gtully:
the logging of the equals methods for the map comparator is gated with debug level in .12 which tidies up our logs.
https://github.com/RHsyseng/operator-utils/releases/tag/v1.4.12

brusdev:
@gtully I'm going to revert the commit related to this issue that upgrades github.com/RHsyseng/operator-utils to v1.4.12 because it is causing #643. I created the issue for github.com/RHsyseng/operator-utils: https://github.com/RHsyseng/operator-utils/issues/110

brusdev:
The fix for https://github.com/RHsyseng/operator-utils/issues/110 was included in github.com/RHsyseng/operator-utils to v1.4.12.1

@gtully do we still need to update github.com/RHsyseng/operator-utils after #649 ?

gtully:
no, we are sorted with that filter.

PR #644, opened 2023-07-10
Revert "[#619] debug log for not equals in map comparator - dep update"

brusdev:
Reverts artemiscloud/activemq-artemis-operator#620


gtully (approved):
Good catch. We may need to provide our own compatator default func to avoid that logging

Issue #655, opened 2023-07-17
Failure to create route or ingress is not visible in the CR.Status


gtully:
If resource create or update or delete errors out we don't report the error in the CR status conditions.
The deployed condition should be unknown if some resource CRUD operation fails.

Example scenario:
`Failed to create new *v1.Route	{"ActiveMQArtemis Name": "cfg-broker-23a-amq-acceptor-original-0-svc-rte", "Namespace": "brkconfig-tests-47fbfc", "error": "Route.route.openshift.io \"cfg-broker-23a-amq-acceptor-original-0-svc-rte\" is invalid: spec.host: Invalid value: \"cfg-broker-23a-amq-acceptor-original-0-svc-rte-brkconfig-tests-47fbfc.apps.bbl711v2.broker.app-services-dev.net\": must be no more than 63 characters"}
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).createRequestedResource
	/workspace/app/controllers/activemqartemis_reconciler.go:1163
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).createResource
	/workspace/app/controllers/activemqartemis_reconciler.go:1146
github.com/artemiscloud/activemq-artemis-operator/controllers.(*ActiveMQArtemisReconcilerImpl).ProcessResources`

gtully:
DeployedStatus will capture these errors with Unknown for reason: ResourceError
here there are two Spec.Acceptors with names too long that error out.

`{Deployed Unknown 0 2023-07-20 17:16:36 +0100 IST ResourceError ["failed to create new *v1.Ingress, Ingress.extensions \"aa-controller542a-my-extra-long-acceptor-0-svc-ing\" is invalid: spec.rules[0].http.paths[0].backend.service.port.name: Invalid value: \"my-extra-long-acceptor-0\": must be no more than 15 characters" "failed to create new *v1.Ingress, Ingress.extensions \"aa-controller542a-my-clashing-acceptor-0-svc-ing\" is invalid: spec.rules[0].http.paths[0].backend.service.port.name: Invalid value: \"my-clashing-acceptor-0\": must be no more than 15 characters"]}`

PR #656, opened 2023-07-17
#655 reflect CRUD resource errors in deployed status condition

gtully:


api/v1beta1/activemqartemis_types.go
```diff
@@ -685,6 +685,7 @@ const (
 	DeployedConditionNotReadyReason         = "PodsNotReady"
 	DeployedConditionZeroSizeReason         = "ZeroSizeDeployment"
 	DeployedConditionValidationFailedReason = "ValidationFailed"
+	DeployedConditionResourceErrorReason    = "ResourceError"
```
gtully, line 688: would ResourceCRUDerror be better?

controllers/activemqartemis_controller.go
```diff
@@ -160,12 +160,12 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
 
 	if valid, result = validate(customResource, r.Client, r.Scheme, *namer); valid {
 
-		reconciler.Process(customResource, *namer, r.Client, r.Scheme)
+		err = reconciler.Process(customResource, *namer, r.Client, r.Scheme)
```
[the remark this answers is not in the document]
brusdev: If we want different type of conditions we could add new error implementations and the caller can then use a [type assertion](https://go.dev/doc/go_spec.html#Type_assertions) to identify the source error type.

controllers/activemqartemis_reconciler.go
```diff
@@ -1152,22 +1154,37 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessResources(customResource
...
 	//empty the collected objects
 	reconciler.requestedResources = nil
 
-	return nil
+	if len(compositeError) == 0 {
+		return nil
+	} else {
+		// maybe errors.Join in go1.20
+		// using %q(uote) to keep errors separate
+		return fmt.Errorf("%q", compositeError)
```
brusdev, line 1177: The function [errors.Join](https://pkg.go.dev/errors@master#Join) could be used to create an error that wraps the given errors. The error formats as the concatenation of the strings obtained by calling the Error method of each element of errs, with a newline between each string.

gtully: not sure about having new lines in a condition message, but also join is in go1.20

api/v1beta1/activemqartemis_types.go
brusdev: I vote for `ResourceError` or `DeploymentResourceError`

brusdev (commented):
This is a great enhancement!!! It will allows to check deployment resource errors from the CR.

brusdev (approved):

controllers/activemqartemis_reconciler.go
brusdev: A newline between each string could be a good separator but I don't know if the Kubernetes/OpenShift console renders the new lines in a condition message.

api/v1beta1/activemqartemis_types.go
gtully: now thinking: KindCRUDError or is that more incomprehensible? it is more accurate

gtully: resource means cpu/memory etc in kube, a type is a kind, and we are doing CRUD as part of the reconcile loop.

controllers/activemqartemis_controller.go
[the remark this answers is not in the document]
gtully: in this case, the message is the error, the crud errored. I don't think we want conditions per error. These errors should only be transient or real and rare.

api/v1beta1/activemqartemis_types.go
gtully: hmm, the crud acronym is too much, going to revert to the original, resource is more general than CPU and RAM and is suitable.

Issue #657, opened 2023-07-18
add validation to ensure no duplicate ports in the Spec.Acceptor list


gtully:
If a CR contains duplicate ports in the acceptors list in error, the broker won't start with bind - address in use errors, and will result in a crash back off loop and the CR.Status Deployed false condition. 
We can short circuit this with a quick sanity validation on the list and report as a Valid condition false instead and reject the deployment.

gtully:
`something like: {[{Valid False 1 2023-07-18 12:40:34 +0100 IST DuplicateAcceptorPort .Spec.Acceptors "second" and "first" contain a duplicate port 61636} {Deployed False 0 2023-07-18 12:40:34 +0100 IST ValidationFailed } {Ready False 0 2023-07-18 12:40:34 +0100 IST WaitingForAllConditions Some conditions are not met}]`

PR #660, opened 2023-07-20
[#657] requeue on resource error, tidy requeue logic, allow verbose s…

gtully:
…etting for tests via env

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -137,61 +137,56 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
 
 	customResource := &brokerv1beta1.ActiveMQArtemis{}
 
-	// Fetch the ActiveMQArtemis instance
-	// When first creating this will have err == nil
-	// When deleting after creation this will have err NotFound
-	// When deleting before creation reconcile won't be called
-	err := r.Get(context.TODO(), request.NamespacedName, customResource)
+	result := ctrl.Result{}
 
+	err := r.Get(context.TODO(), request.NamespacedName, customResource)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			reqLogger.V(1).Info("ActiveMQArtemis Controller Reconcile encountered a IsNotFound, for request NamespacedName " + request.NamespacedName.String())
-			return ctrl.Result{}, nil
+			return result, nil
 		}
-		reqLogger.Error(err, "unable to retrieve the ActiveMQArtemis for request "+request.Namespace+"/"+request.Name)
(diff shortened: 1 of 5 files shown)


controllers/activemqartemis_controller.go
```diff
@@ -137,61 +137,56 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
...
+	if requeueRequest {
+		reqLogger.V(1).Info("requeue reconcile")
+		result = ctrl.Result{RequeueAfter: common.GetReconcileResyncPeriod()}
+	}
+
 	return result, err
 }
 
-func validate(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, scheme *runtime.Scheme, namer Namers) (bool, ctrl.Result) {
+func validate(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, scheme *runtime.Scheme, namer Namers) (valid bool, retry bool) {
```
brusdev, line 189: Where are you using the named return arguments?

controllers/activemqartemis_controller.go
```diff
@@ -137,61 +137,56 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
...
-
-	if err != nil {
-		if apierrors.IsConflict(err) {
-			reqLogger.V(1).Info("unable to update ActiveMQArtemis status", "Request Namespace", request.Namespace, "Request Name", request.Name, "error", err)
-			err = nil // we don't want the controller event loop reporting this as an error
-		} else {
-			reqLogger.V(1).Error(err, "unable to update ActiveMQArtemis status", "Request Namespace", request.Namespace, "Request Name", request.Name)
-		}
-		return ctrl.Result{RequeueAfter: common.GetReconcileResyncPeriod()}, err
+	crStatusUpdateErr := UpdateCRStatus(customResource, r.Client, request.NamespacedName)
```
brusdev, line 168: Did you remove the dump of the err from the log on purpose?

brusdev (approved):

controllers/activemqartemis_controller.go
gtully: yes, it was a duplicate, already logged by resources.UpdateStatus, good watching!

controllers/activemqartemis_controller.go
gtully: I was only using one, the retry. Can remove the name from the first bool, will do that. thanks

gtully: I did not use a naked return, but I think it helps to use the named argument in this case.

gtully:
@brusdev thanks for the fast feedback :-)

Issue #623, opened 2023-06-26
Expose console using cert manager


gaohoward:
**Is your feature request related to a problem? Please describe.**
When cert-manager is installed, users should be able to use cert-manager to manage certs for console expose with ssl enabled.

**Describe the solution you'd like**

**Describe alternatives you've considered**
Optional. A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
Optional. Add any other context or screenshots about the feature request here.

PR #666, opened 2023-07-31
[#623] Expose console using cert manager

gaohoward:


rvais (commented):
So far I can't see anything that would be obviously wrong. It's a draft so I expect more changes.  Too much to keep everything in my head, especially since it's written in go-lang.

bundle/manifests/broker.amq.io_activemqartemises.yaml
```diff
@@ -1970,8 +1974,6 @@ spec:
                         type: integer
                       grpc:
                         description: GRPC specifies an action involving a GRPC port.
```
rvais, line 1976: What is this related to?

gaohoward: This new field is auto generated because of our dependency upgrade (which is needed by the cert manager package dependency)

gaohoward:
I'm closing this as it's included in another PR https://github.com/artemiscloud/activemq-artemis-operator/pull/673

Issue #588, opened 2023-05-08
Support custom annotations and labels for all resources created by the operator


brusdev:
The operator supports custom annotations and labels for the pods related to the ActiveMQArtemis CRs by setting the `spec.deploymentPlan.annotations` and `spec.deploymentPlan.labels` fields.

Prefixes could be use in the keys to apply annotations and labels to a specific resource type, i.e. the prefix  `io.artemiscloud.ingress.annotation/` could be used to apply annotations to ingress resources:
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: artemis-broker-resources
spec:
  deploymentPlan:
    labels:
      io.artemiscloud.ingress.annotation/foo: too
```

gtully:
would it just be? 
labels:
      io.artemiscloud.ingress/foo: too

where label or annotation is implicit. First forward slash delimits.

brusdev:
I like the char `@` to identify the item name +1

brusdev:
I was thinking that maybe the ActiveMQArtemis CR should have the field `labels` in acceptor, connector and console too

joshdreagan:
One additional suggestion would be to add the ability to use templates in the annotations. Doing so would allow for easy integration with skupper. 

For example:

```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: artemis-broker
spec:
  acceptors:
    - name: amqp
      port: 5672
      protocols: amqp 
      annotations:
        - skupper.io/address: "artemis-broker-amqp-${STATEFUL_SET_ORDINAL}-svc-dc1"
        - skupper.io/port: "5672"
        - skupper.io/proxy: "tcp"
```

Possibly also allow for specifying annotations that only apply to certain members of the StatefulSet.

gtully:
@davidlanouette thanks for your PR, I used your tests as a start point, I think the Template idea in #737 will resolve your use case, that was the intent. It keeps the existing pod labels and annotations intact, those are just for the PodTemplate in the SS.

gtully:
there is support for templates in the keys and values using the variables from ingressHost, see: https://github.com/artemiscloud/activemq-artemis-operator/blob/23deb38c3686bafadf9d07664b781a83868f3a9a/deploy/crds/broker_activemqartemis_crd.yaml#L60

this should support the scupper use case @joshdreagan

gtully:
[...]

PR #672, opened 2023-08-15
Draft: [481] Custom Labels for Services and Pods

davidlanouette:
Apply custom labels to most resources that are created by the operator.

Not all resources have labels, so some - like Containers, EnvVar's, etc.

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -8,8 +8,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"hash/adler32"
+	"os"
+	"reflect"
 	osruntime "runtime"
 	"sort"
+	"strconv"
+	"strings"
 	"unicode"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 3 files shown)


davidlanouette:
This needs additional tests to make sure all resources have the labels applied correctly.  But, it's in generally good shape to review.

brusdev:
@davidlanouette thanks for your PR but there is also the issue #588 related to custom labels. The solution discussed in the issue #588 is that the ActiveMQArtemis CR should have the field labels in acceptor, connector and console too.

davidlanouette:
@brusdev Thanks for the note.  Looking at that issue, it looks like it's _mostly_ a duplicate of 481.  I'll comment on that issue.

Regardless, I'll add see about incorporating the comments in that issue here.

brusdev:
@davidlanouette I think that we can close this this PR because the solution implemented by #737 already allows [Custom Labels and Annotations on supporting resources; Services, Ingress, Secrets etc.](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/docs/help/operator.md#custom-labels-and-annotations-on-supporting-resources-services-ingress-secrets-etc). Feel free to reopen it if I'm missing something.

PR #686, opened 2023-09-20
Fix Scorecard Test

### LINKED ISSUE #684: Fix Scorecard Test
The scorecard test is currently failing, and it's preventing us from proceeding with the testing process. We need to address this issue to ensure the successful execution of all scorecard tests.

Msarawan:


install.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
gaohoward, line 1: @Msarawan I think the name of the script should be more specific, like "install_olm.sh" so that it would be easy to get the idea what this script is about. Also the script is better to be moved into the **hack** dir.

gaohoward (changes requested):

install.sh
Msarawan: modified the changes.

brusdev (approved):
@Msarawan nice work!

Msarawan:
> @Msarawan nice work!

Thank You! @brusdev

gaohoward (approved):

PR #690, opened 2023-09-23
remove metadata.application from CRs

pwright:


gaohoward:
@pwright It looks like your comment message didn't pass the ci check. You need either use this format:
[NO-ISSUE] commit message
for commits not related to any issue, or
[# nnn] commit message
for commits releated to an issue.

Issue #702, opened 2023-10-06
Remove ControllerManagerConfig


brusdev:
The ComponentConfig has been deprecated in the Controller-Runtime since its version 0.15.0. [More info](https://github.com/kubernetes-sigs/controller-runtime/issues/895) Moreover, it has undergone breaking changes and is no longer functioning as intended. As a result, Kubebuilder, which heavily relies on the Controller Runtime, has also deprecated this feature, no longer guaranteeing its functionality from version 3.11.0 onwards. You can find additional details on this issue [here](https://github.com/kubernetes-sigs/controller-runtime/issues/2370).

Please, be aware that it will force Kubebuilder remove this option soon in future release.

PR #703, opened 2023-10-06
[#702] Remove ControllerManagerConfig

brusdev:
The ComponentConfig has been deprecated in the Controller-Runtime since its version 0.15.0. [More info](https://github.com/kubernetes-sigs/controller-runtime/issues/895) Moreover, it has undergone breaking changes and is no longer functioning as intended. As a result, Kubebuilder, which heavily relies on the Controller Runtime, has also deprecated this feature, no longer guaranteeing its functionality from version 3.11.0 onwards. You can find additional details on this issue [here](https://github.com/kubernetes-sigs/controller-runtime/issues/2370).

Please, be aware that it will force Kubebuilder remove this option soon in future release.


gtully (commented):
great. I found my self looking for how to configure QPS for the api client rate limit and came across this and found it to be of no use :-) and now find it is deprecated - good news.

brusdev:
@gtully an alternative to ComponentConfig could be the env `ARGS`. This alternative works with OLM deplyments too, see https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md

I used that mechanism in #701.

PR #705, opened 2023-10-09
[#704] Improve operator logging

### LINKED ISSUE #704: Improve operator logging

Currently the operator logs a lot informations at the info level. Which is too verbose and difficult to read
Also as more reconcile loops are called the log message accumulates.
Loggers should be created from a single source so to make the logger configuration easier.

* All loggers come from ctrl.Log which is initiated/configured in main.go
* Each logger is given a proper name - each controller has a logger with its owned CRD name as its name.
   all util like methods use package specific logger from ctrl.Log.WithName()
* Log messages are at level 1, 2 or 3, no 0-level log 
* when level set to INFO, Except during startup, only error messages(if any) are shown in the log

**Additional context**

gaohoward:
* All loggers come from ctrl.Log which is initiated/configured in main.go
* Each logger is given a proper name - each controller has a logger with its owned CRD name as its name. all util like methods use package specific logger from ctrl.Log.WithName()
* Log messages are at level 1, 2 or 3, no 0-level log
* when level set to INFO, Except during startup, only error messages(if any) are shown in the log

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -38,14 +38,13 @@ import (
 
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/resources"
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/utils/namer"
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 
 	brokerv1beta1 "github.com/artemiscloud/activemq-artemis-operator/api/v1beta1"
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/utils/common"
 	"github.com/artemiscloud/activemq-artemis-operator/pkg/utils/selectors"
 )
-
-var clog = ctrl.Log.WithName("controller_v1beta1activemqartemis")
 
 var namespaceToConfigHandler = make(map[types.NamespacedName]common.ActiveMQArtemisConfigHandler)
 
@@ -69,7 +68,7 @@ func (r *ActiveMQArtemisReconciler) UpdatePodForSecurity(securityHandlerNamespac
 			candidate.Name = artemis.Name
 			candidate.Namespace = artemis.Namespace
 			if handler.IsApplicableFor(candidate) {
-				clog.V(1).Info("force reconcile for security", "handler", securityHandlerNamespacedName, "CR", candidate)
+				r.log.V(1).Info("force reconcile for security", "handler", securityHandlerNamespacedName, "CR", candidate)
 				r.events <- event.GenericEvent{Object: &existingCrs.Items[index]}
 			}
 		}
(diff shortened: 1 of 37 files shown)


controllers/activemqartemis_controller.go
```diff
@@ -311,41 +319,6 @@ func validatePodDisruption(customResource *brokerv1beta1.ActiveMQArtemis) *metav
 	return nil
 }
 
-func validateBrokerVersion(customResource *brokerv1beta1.ActiveMQArtemis) *metav1.Condition {
```
brusdev, line 314: `validateBrokerVersion` seems only used in `controllers/activemqartemis_controller.go`, why was it moved?

controllers/activemqartemis_controller.go
```diff
@@ -163,15 +171,15 @@ func (r *ActiveMQArtemisReconciler) Reconcile(ctx context.Context, request ctrl.
 		}
 	}
 
-	ProcessStatus(customResource, r.Client, request.NamespacedName, *namer, err)
+	common.ProcessStatus(customResource, r.Client, request.NamespacedName, *namer, err)
```
brusdev, line 174: ProcessStatus seems only used in controllers/activemqartemis_controller.go, why was it moved?

controllers/activemqartemis_reconciler.go
```diff
@@ -2302,7 +2182,7 @@ func createExtraConfigmapsAndSecretsVolumeMounts(brokerContainer *corev1.Contain
 	if len(secrets) > 0 {
 		for _, secret := range secrets {
 			if secret == "" {
-				clog.Info("No Secret name specified, ignore", "Secret", secret)
+				r.log.V(2).Info("No Secret name specified, ignore", "Secret", secret)
```
brusdev, line 2185: V(2).Info means that users need to set `--zap-log-level` >= 2 because `--zap-log-level=debug` only prints Info levels <= 1. Is that on purpose?

brusdev:
@howardgao this is a great work that will improve the user experience!!!

controllers/activemqartemis_controller.go
gaohoward: this is also to move those utility func to common to reuse not live in any controllers.

controllers/activemqartemis_reconciler.go
gaohoward: yes I was thinking about the rules at the moment. Making it level 2 because it prints the whole secret object.
rather that a simple type var. (same rule apply to printing a whole statefulset, deployment etc).
But we need a more comprehensive set of logging rules to follow. I'll be thinking of it.

gaohoward:
As discussed we will limit the logging just up to 1 for now. make Error default and use INFO and V(1).INFO for debug and trace equivalent logging. I'll update the PR following that decision. (feel free to argue against that)

I'll also think of some rules for logging. :)

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
```diff
@@ -2570,7 +2570,7 @@ spec:
             spec:
               containers:
               - args:
-                - --zap-log-level=info
+                - --zap-log-level=error
```
brusdev, line 2573: @gaohoward @gtully we have already discussed this offline but I still think we should use info level by default. If it is too  much verbose we could move some log messages to debug level `logger.V(1).Info(...`

gaohoward: Yeah It would be more conventional to set info as default level. This also make it easy to print the start up logs in normal info format without having to use a separate logger (the setup logger). This is because the error log doesn't print simple message lines, it also print the stack trace even a nil err is passed in, polluting the log output.

gaohoward: I'll change it back to info if no objection. @brusdev @gtully

brusdev (approved):

Issue #708, opened 2023-10-10
jaas-config validation regexp needs to take account of quoted property values and space around =


gtully:
The validation introduced via #495 needs to be more lenient on the values of properties. They can be any content once quoted. In addition, space around = needs to be treated as insignificant.
Thus can avoid the need to disable validation using the operator env var - JAAS_CONFIG_SYNTAX_MATCH_REGEX="" 

examples of problematic config:

roleSearchMatching="(&(objectCategory=group)(groupType:1.1.111.111111.1.1.111:=1111111111)(member:1.1.111.111111.1.1.1111:={0})(sAMAccountName=AAA AAA AAA*))"
roleSp =true
roleSp= true

PR #709, opened 2023-10-10
[#708] support quoted keys and space around = for login.config valida…

gtully:
…tion regexp

### DIFF
--- a/pkg/utils/common/common.go
+++ b/pkg/utils/common/common.go
@@ -21,7 +21,7 @@ const (
 	DEFAULT_RESYNC_PERIOD  = 30 * time.Second
 	// comments push this over the edge a little when dealing with white space
 	// as en env var it can be disabled by setting to "" or can be improved!
-	JaasConfigSyntaxMatchRegExDefault = `^(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s*{(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s+(?i:required|optional|sufficient|requisite)+(?:\s*\S+=\S+\s*)*\s*;)+(\s*|(?://.*)|(?s:/\*.*\*/))*}\s*;)+\s*\z`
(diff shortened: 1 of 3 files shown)


pkg/utils/common/common.go
```diff
@@ -21,7 +21,7 @@ const (
 	DEFAULT_RESYNC_PERIOD  = 30 * time.Second
 	// comments push this over the edge a little when dealing with white space
 	// as en env var it can be disabled by setting to "" or can be improved!
-	JaasConfigSyntaxMatchRegExDefault = `^(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s*{(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s+(?i:required|optional|sufficient|requisite)+(?:\s*\S+=\S+\s*)*\s*;)+(\s*|(?://.*)|(?s:/\*.*\*/))*}\s*;)+\s*\z`
+	JaasConfigSyntaxMatchRegExDefault = `^(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s*{(?:(\s*|(?://.*)|(?s:/\*.*\*/))*\S+\s+(?i:required|optional|sufficient|requisite)+(?:\s*\S+\s*=((\s*\S+\s*)|("[^"]*")))*\s*;)+(\s*|(?://.*)|(?s:/\*.*\*/))*}\s*;)+\s*\z`
```
gaohoward, line 24: This regex: (?i:required|optional|sufficient|requisite)+
the '+' suggest one or more such words, should it be just once? with that the following could be legal
'a_realm { a_good_login_module optionalsufficient noOp=true; };'
?

gtully: good catch, that + needs to be removed. Will add a test and proceed. thanks!

gtully: sorted

gaohoward (approved):

Issue #711, opened 2023-10-12
use of init container can break brokerProperties system property set via JAVA_OPTS in profile, using JDK_JAVA_OPTIONS would just fix that 


gtully:
If an init container replaces the artemis.profile, any modification via JAVA_OPTS are lost, leading to a loss of the operator supplied system property and missing properties.

```- lastTransitionTime: '2023-09-29T14:46:47Z'
 message: 'Status out of sync - missing status entry for keys: [broker.properties]'
 reason: OutOfSync
 status: Unknown
 type: BrokerPropertiesApplied```

Using JDK_JAVA_OPTIONS env var to directly configure the jvm will avoid this problem of inter-dependence on the run scripts etc.

PR #712, opened 2023-10-12
[#711] use JKD_JAVA_OPTIONS in place of JAVA_OPTS for operator jvm co…

gtully:
…nfig

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -1764,17 +1764,29 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
 
 	clog.V(1).Info("Total volumes ", "volumes", podSpec.Volumes)
 
-	// this depends on init container passing --java-opts to artemis create via launch.sh *and* it
-	// not getting munged on the way. We CreateOrAppend to any value from spec.Env
 	var mountPoint = secretPathBase
 	if !isSecret {
 		mountPoint = cfgMapPathBase
 	}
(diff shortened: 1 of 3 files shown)


controllers/activemqartemis_reconciler.go
```diff
@@ -1764,17 +1764,29 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) NewPodTemplateSpecForCR(customR
...
 		mountPoint = cfgMapPathBase
 	}
-	javaOpts := corev1.EnvVar{
-		Name:  "JAVA_OPTS",
-		Value: brokerPropertiesConfigSystemPropValue(mountPoint, brokerPropertiesResourceName, brokerPropertiesMapData),
+	brokerPropsValue := brokerPropertiesConfigSystemPropValue(mountPoint, brokerPropertiesResourceName, brokerPropertiesMapData)
+
+	// only use init container JAVA_OPTS on existing deployments and migrate to JDK_JAVA_OPTIONS for independence
+	// from init containers and broker run scripts
+	if environments.Retrieve(podSpec.InitContainers, "JAVA_OPTS") != nil {
```
brusdev, line 1775: is this check required to avoid broker pod restarts after the operator upgrade?

brusdev (approved):

controllers/activemqartemis_reconciler.go
gtully: exactly. if there is any use of JAVA_OPTS, presumably they work and lets piggy back on that and not force a restart by a env change to the SS. 
If not, this is a new CR so we can go with the new approach.

Issue #715, opened 2023-10-16
Better support for Istio


johgoe:
**Is your feature request related to a problem? Please describe.**
Istio Headless services requires all ports declared in the headless service and a naming convention or appProtocol instead.
see:
https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services

**Describe the solution you'd like**
the service my-cluster-hdls-svc should contain a istio compatible declaration of ports e.g.

```
    - name: console-jolokia
      port: 8161
      protocol: TCP
      targetPort: 8161
      appProtocol: http
    - name: all
      port: 61616
      protocol: TCP
      targetPort: 61616
      acceptors: http
```
or alternative

**Describe alternatives you've considered**
```
    - name: http-console-jolokia
      port: 8161
      protocol: TCP
      targetPort: 8161
    - name: tcp-all
      port: 61616
      protocol: TCP
      targetPort: 6161
```
**Additional context**

The issue does not occurs for acceptors but only for the default ports because it's not possible to choose any name there.
Maybe acceptors should also be able to set the appProtocol field.

Is  there any reason why acceptors will produce a new service per node instead of adding the additional port to the headless service?

brusdev:
This enhancement will be included in 1.0.16

PR #720, opened 2023-10-22
Add protocol to console default port to fix issue with istio

johgoe:
Add protocol to console default port to fix issue with istio when reading metrics 

see #715


brusdev (approved):

brusdev:
@johgoe thanks for your contribution

PR #730, opened 2023-11-07
 [694] - Enable metrics to be scraped on operator

### LINKED ISSUE #694: Support for operator metrics

It is not documented how to expose metrics from the operator itself.

The Artemis operator is built on-top on the operator-sdk which in turns uses the kube-builder library.
kube-builder has default list of metrics that are available in a Prometheus format.
See for reference: https://book.kubebuilder.io/reference/metrics-reference

When the operator starts up, it starts a metrics endpoint that can be scraped.

To allow metrics to be scraped, a k8s service and some annotations on the k8s operator deployment are required.
Along with supporting documentation

robert-mcnamara:
Fixes [694](https://github.com/artemiscloud/activemq-artemis-operator/issues/694)

### Type of change
Enhancement

### Description
- Add service to expose metrics port
- Add Prometheus annotations
- Documentation updated

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -120,6 +120,9 @@ spec:
           initialDelaySeconds: 15
           periodSeconds: 20
         name: manager
+        ports:
+        - name: http-metrics
+          containerPort: 8383
         readinessProbe:
           httpGet:
             path: /readyz
--- a/docs/help/operator.md
+++ b/docs/help/operator.md
@@ -983,6 +983,24 @@ spec:
 ```
 For a complete example please refer to this [artemiscloud example](https://github.com/artemiscloud/artemiscloud-examples/tree/main/operator/prometheus).
 
+## Enabling Operator Metrics
+
+The operator exposes a port called **http-metrics** for Prometheus to monitor.
+The following is a sample Prometheus PodMonitor resource
+
+```yaml
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: activemq-artemis-controller-manager
+spec:
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  podMetricsEndpoints:
+  - port: http-metrics
+```
+
 ## Configuring PodDisruptionBudget for broker deployment
 
 The ActiveMQArtemis custom resource offers a PodDisruptionBudget option


brusdev:
@robert-mcnamara in this way the users have to change the operator deployment, an alternative could be to define the port `http-metrics` in the operator deployment template and add to the doc how to create a PodMonitor CR for the operator pod, i.e.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
  name: activemq-artemis-controller-manager
spec:
  template:
    spec:
      containers:
        ports:
        - name: http-metrics
          containerPort: 8383
...
```

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: example-app
  labels:
    team: frontend
spec:
  selector:
    matchLabels:
      app: example-app
  podMetricsEndpoints:
  - port: http-metrics
```

WDYT?

robert-mcnamara:
> @robert-mcnamara in this way the users have to change the operator deployment, an alternative could be to define the port `http-metrics` in the operator deployment template and add to the doc how to create a PodMonitor CR for the operator pod, i.e.
> 
> ```
> apiVersion: apps/v1
> kind: Deployment
> metadata:
>   labels:
>     control-plane: controller-manager
>   name: activemq-artemis-controller-manager
> spec:
>   template:
>     spec:
>       containers:
>         ports:
>         - name: http-metrics
>           containerPort: 8383
> ...
> ```
> 
> ```
> apiVersion: monitoring.coreos.com/v1
> kind: PodMonitor
> metadata:
>   name: example-app
>   labels:
>     team: frontend
> spec:
>   selector:
>     matchLabels:
>       control-plane: controller-manager
>   podMetricsEndpoints:
>   - port: http-metrics
> ```
> 
> WDYT?

Thanks for the feedback.

Is the concern the addition of the service? With the alternative it would require an update to the existing deployment and still require an addition resource of kind PodMonitor, so to me the impact is same we have just replaced the service with the PodMonitor.

It also prescribes the deployment of Prometheus via operator versus auto-discovery.

I did notice the artemis cloud docs already give an example of monitoring the brokers via a ServiceMonitor kind. 
https://artemiscloud.io/docs/help/operator/#monitor-broker-metrics-by-using-prometheus 

So consistency might be the best approach here?

brusdev:
@robert-mcnamara I have no concerns for the new operator service and I agree with you on monitoring the operator via a ServiceMonitor kind for consistency with  the brokers.

My only concerns is for the new annotations, because the users would have to change the operator deployment to disable the auto-discovery:
```
      annotations:
        prometheus.io/port: "8383"
        prometheus.io/scrape: "true"
        prometheus.io/path: "/metrics"
```

brusdev:
@robert-mcnamara are you still working on this PR?

brusdev (approved):

brusdev:
@robert-mcnamara I manually merged this PR to fix CI check failure by executing `make generate-deploy bundle` to regenerate static manifests, thanks for your contribution!!!

Issue #717, opened 2023-10-17
Large messages management issues for AMQP and STOMP


pabateman:
**Is your feature request related to a problem? Please describe.**

According to [ActiveMQ Artemis documentation](https://activemq.apache.org/components/artemis/documentation/latest/large-messages.html#configuring-amqp-acceptor) `-1` is valid `amqpMinLargeMessageSize` value, but [in code](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemis_reconciler.go#L654) this even won't be considered.

Also it would be nice to have same parameter for STOMP. This parameter [described in documentation](https://activemq.apache.org/components/artemis/documentation/latest/stomp.html#handling-of-large-messages-with-stomp) too.

**Describe the solution you'd like**

- Extend value range for `amqpMinLargeMessageSize`
- Add `stompMinLargeMissageSize` for [AcceptorType](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/api/v1beta1/activemqartemis_types.go#L418)

I can contribute if you don't mind.

**Additional context**

I noticed that you use OperatorSDK for CRD generating, that uses kubebuilder under the hood.  Kubebuilder supports [validation annotations](https://book.kubebuilder.io/reference/markers/crd-validation), which enrich CRD spec with value validation. It would be more friendly if the user became aware of the incorrect specification at the stage of applying the resource in the API.

brusdev:
@pabateman thanks, contributes are always welcome!
I agree with you, the current value range for `amqpMinLargeMessageSize` is bug ([code](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemis_reconciler.go#L654)).
You don't need to add new field to the AcceptorType for new extra parameters, you could use broker properties, i.e.
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: artemis-stomp
spec:
  acceptors:
    - name: my-stomp
      protocols: STOMP
      port: 61613
  brokerProperties:
    - "acceptorConfigurations.my-stomp.extraParams.stompMinLargeMissageSize=..."
```

pabateman:
Sorry, a lot of work.

Thanks for the answer! 

> I agree with you, the current value range for amqpMinLargeMessageSize is bug ([code](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemis_reconciler.go#L654)).

[...]

PR #731, opened 2023-11-07
Allow -1 as valid amqpMinLargeMessage value

pabateman:
Extends `acceptor.AMQPMinLargeMessageSize` condition because `acceptor.AMQPMinLargeMessageSize > -1` is not valid condition, since if not `amqpMinLargeMessageSize` is set, then 0 is be the default value.

Ref #717


pabateman:
> @pabateman your PR LGTM, could you add the issue number to your commit message to fix the [PR check failure](https://github.com/artemiscloud/activemq-artemis-operator/actions/runs/6784361444/job/18467812394?pr=731)?
> 
> i.e.
> 
> ```
> [#717] allow -1 as valid amqpMinLargeMessage value
> ```

Sure!

brusdev:
@pabateman thanks for your contribution!

Issue #614, opened 2023-06-14
Support custom ingress hosts


brusdev:
Add the field `spec.ingressHost` to customize the host for Ingress and Route resources of the broker. The field `spec.ingressHost` also supports the following variables:
- $(CR_NAME) - the name of the custom resource
- $(CR_NAMESPACE) - the namespace of the custom resource
- $(BROKER_ORDINAL) - the StatefulSet ordinal index of the broker pod
- $(ITEM_NAME) - the name of the item (acceptor's name, connector's name or console)
- $(RES_TYPE) - the type of the resource (ing for Ingress and rte for Route)
- $(INGRESS_DOMAIN) - the value of the field `spec.ingressDomain`

The following ActiveMQArtemis CR generates the following acceptor, connector and console Ingress resources:
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata: 
  name: ex-aoo
spec: 
  deploymentPlan: 
    size: 2
  acceptors: 
  - name: my-acceptor
    port: 61617
    sslEnabled: true
    expose: true
    ingressHost: $(CR_NAME)-$(ITEM_NAME)-$(BROKER_ORDINAL).$(INGRESS_DOMAIN)
  connectors: 
  - name: my-connector
    port: 61617
    sslEnabled: true
    expose: true
    ingressHost: $(CR_NAME)-$(ITEM_NAME)-$(BROKER_ORDINAL).$(INGRESS_DOMAIN)
  console: 
    name: my-console
    sslEnabled: true
    expose: true
    ingressHost: $(CR_NAME)-$(ITEM_NAME)-$(BROKER_ORDINAL).$(INGRESS_DOMAIN)
  ingressDomain: my-domain.com
```
```
ex-aoo-my-acceptor-0.my-domain.com
ex-aoo-my-acceptor-1.my-domain.com
```
```
ex-aoo-my-connector-0.my-domain.com
ex-aoo-my-connector-1.my-domain.com
```
```
ex-aoo-my-console-0.my-domain.com
ex-aoo-my-console-1.my-domain.com
```

brusdev:
This feature will be included in v1.0.12

PR #737, opened 2023-11-10
[#588] support template for custom labels and annotations that can ma…

gtully:
…tch any supporting resource managed by the operator


docs/help/operator.md
```diff
@@ -737,6 +737,28 @@ spec:
       promethes-prop: "somevalue"
 ```
 
+### Custom Labels and Annotations on supporting resources; Services, Ingress, Secrets etc.
+It is possible to configure a Template(s) for resources that are managed by the operator.
+The Template contains a Labels and Annotations with an optional Selector. If the selector is empty
+the template matches all resources. Othewise it can be used to restrict what is matched.
+Note: the relevant variables supported by `ingressHost` can be referenced in keys and values for both labels and annotations.
```
gaohoward, line 744: IngressHost seems confusing as there is nowhere else it is mentioned.

gaohoward: It it would be nice to document supported resource types for labels and annotations. (i.e. services, secrets, ingresses/routes)

gaohoward (approved):
It looks good to me in general.

docs/help/operator.md
gtully: thanks, I wonder if it is trivial to link to that?

The reference to ingressHost is via the CRD: https://github.com/artemiscloud/activemq-artemis-operator/blob/main/deploy/crds/broker_activemqartemis_crd.yaml#L59

api/v1beta1/activemqartemis_types.go
```diff
@@ -370,6 +370,33 @@ type DeploymentPlanType struct {
 	// Specifies the Revision History Limit of the statefulset
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Revision History Limit"
 	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
+	// Specifies the template for various resources that the operator controls
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resource Templates"
+	Templates []TemplateType `json:"templates,omitempty"`
```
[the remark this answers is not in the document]
gtully: I must have been thinking that too based on the Display Name - I can revert to that. Thanks!

docs/help/operator.md
gaohoward: yeah, that'll be great

pkg/resources/ingresses/ingress.go
```diff
@@ -82,7 +80,10 @@ func NewIngressForCRWithSSL(existing *netv1.Ingress, namespacedName types.Namesp
 
 	desired.Spec.Rules[0].Host = host
 	if sslEnabled {
-		desired.Annotations = map[string]string{"nginx.ingress.kubernetes.io/ssl-passthrough": "true"}
+		if desired.Annotations == nil {
+			desired.Annotations = make(map[string]string)
+		}
+		desired.Annotations["nginx.ingress.kubernetes.io/ssl-passthrough"] = "true"
```
brusdev, line 86: The annotation `nginx.ingress.kubernetes.io/ssl-passthrough` is specific for nginx ingress controller and the new templates field could be used to set this annotation. What about removing the annotation `nginx.ingress.kubernetes.io/ssl-passthrough` and adding to the doc how to set it using the new templates field?

api/v1beta1/activemqartemis_types.go
```diff
@@ -370,6 +370,33 @@ type DeploymentPlanType struct {
...
+
+// match criteria to restrict the selection of resources
+type ResourceMatch struct {
+	// APIGroup is the group version string of resources to match
+	APIGroup *string `json:"apiGroup,omitempty"`
+	// Kind is the type of resources to match
+	Kind *string `json:"kind,omitempty"`
+	// Name is the pattern matcher for the Name of resources to match
+	Name *string `json:"name,omitempty"`
 }
```
brusdev, line 400: [OPTIONAL] looking the kubernetes api, I found something similar fileds/structs in the [StatefulSetSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#statefulsetspec-v1-apps)
```
selector
LabelSelector
...
template
PodTemplateSpec
...
```

If my understanding is correct (BIG IF 😅) `TemplateType` could also be `ResourceTemplateSpec` and `ResourceMatch` could also be `ResourceSelector`, WDYT?

brusdev:
@gtully great work!!! The full test-suite passes without failures.

api/v1beta1/activemqartemis_types.go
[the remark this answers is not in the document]
gtully: now also wondering if this should be party of the Spec or part of the Spec.DeploymentPlan, is there a good heuristic that is in play to help?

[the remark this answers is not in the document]
gtully: Think it should be at the Spec level b/c acceptors at that level are the main driver of services and ingress etc.

controllers/activemqartemis_reconciler_test.go
```diff
@@ -510,12 +513,241 @@ func TestNewPodTemplateSpecForCR_IncludesDebugArgs(t *testing.T) {
...
+		if secret, ok := resource.(*v1.Secret); ok {
+			assert.True(t, len(secret.Labels) >= 1)
+			assert.Equal(t, secret.Labels["myKey"], "myValue")
+			assert.Equal(t, secret.Labels["mySecretKey"], "mySecretValue")
+			secretFound = true
+		}
+
+		if service, ok := resource.(*v1.Service); ok {
+			assert.True(t, len(service.Labels) >= 1)
+			assert.Equal(t, service.Labels["myKey"], "myValue")
```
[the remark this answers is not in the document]
gtully: yes, I can add an assertion for that. thanks

api/v1beta1/activemqartemis_types.go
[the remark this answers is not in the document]
brusdev: I see a lot of stuff related to the broker config at the Spec level and a lot of stuff related to the Kubernetes resources config at the `Spec.DeploymentPlan` level, if `Templates` is for kubernetes resources could also make sense keep them at the `Spec.DeploymentPlan` level.

controllers/activemqartemis_reconciler_test.go
```diff
@@ -510,12 +513,241 @@ func TestNewPodTemplateSpecForCR_IncludesDebugArgs(t *testing.T) {
...
+	)
+
+	namer := MakeNamers(cr)
+
+	newSS, _ := reconciler.ProcessStatefulSet(cr, *namer, nil)
+	reconciler.ProcessDeploymentPlan(cr, *namer, nil, nil, newSS)
+
+	var secretFound = false
+	for _, resource := range reconciler.requestedResources {
+		if secret, ok := resource.(*v1.Secret); ok {
```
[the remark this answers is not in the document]
gtully: sure, thanks!

pkg/resources/ingresses/ingress.go
gtully: I was thinking that could be a follow up commit, it would be a good example of usage. I like the idea. Thanks.

api/v1beta1/activemqartemis_types.go
gtully: for sure Type does not bring much value, I was blindly following AddressType/AcceptorType etc... let me revisit. Thanks!

gtully: same holds for the Spec postfix, am thinking TemplateType could also be ResourceTemplate

docs/help/operator.md
gtully: https://github.com/artemiscloud/activemq-artemis-operator/issues/614 I will link to the issue

api/v1beta1/activemqartemis_types.go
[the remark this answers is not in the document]
gtully: interesting, worth more consideration then.. the original may have been correct. I think your split of concerns may be closer to what we have, I was equating plan to the stateful set, but it is more general than that.
I will revisit and probably put it back in the Plan.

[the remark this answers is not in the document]
gtully: the resourceTemplate is for sure about kube resources

brusdev (approved):

api/v1beta1/activemqartemis_types.go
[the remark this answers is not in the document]
gtully: peeking some more, it is not clear cut and the exposure defs are the spec level, I think on balance it is best at the Spec level too. will leave as it is now.

PR #743, opened 2023-11-22
[#741] Make deployment to be compatible with restricted level Pod Security Admission

### LINKED ISSUE #741: Support deployment in namespaces with stricted Pod Security Admission
**Describe the bug**
The Pod Security Admission will be enforced upon Openshift and the operator should make sure it conforms to the restricted security level.

gaohoward:


gtully:
is there any scenario where the broker will need some capabilities? I am wondering if this needs to be something we can configure rather than something that is hardcoded. Would extending our CR with PVC[] and Service[] leaves us open to requiring more capabilities I wonder?

gaohoward:
> is there any scenario where the broker will need some capabilities? I am wondering if this needs to be something we can configure rather than something that is hardcoded. Would extending our CR with PVC[] and Service[] leave us open to requiring more capabilities I wonder?

I don't think broker need any special capabilities by default. But if it does, user can configure the SecurityContext to add those capabilities. If those added capabilities conflicts with the default PSA/SCC policy in the target namespace, the user needs to modify/create the target namespace with required policies.

brusdev (approved):
Restricting special capabilities by default if those are not required is the safest option. Users can use the new template patches to customize them.

brusdev:
@howardgao I incorporated the commits from this PR in #801 to test the operator deployment in a restricted namespace. I removed the changes related to the security contexts of the broker pods because they were causing a conflict with the commits from #757 and I split the test to be consistent.

gaohoward:
@brusdev thanks. I'll rebase the PR.

PR #749, opened 2023-11-27
[#748] adding a single yaml for installing operator

### LINKED ISSUE #748: adding a single yaml for installing operator
It would be convenient for users/developers to use a single yaml to install the operator and add it to the releases as an artifact.

**Additional context**

gaohoward:


brusdev:
@howardgao providing a single yaml for installling the operator is a very good feature. Strmizi also has a similar feature, it has an endpoint do dynamically create  a single yaml, see https://strimzi.io/quickstarts/:
```
kubectl create -f 'https://strimzi.io/install/latest?namespace=kafka' -n kafka
```
The latest endpoint is defined in https://github.com/strimzi/strimzi.github.io/blob/main/.functions/namespace/namespace.js

A static single yaml has the following drawbacks:
- another static file to check and update
- we would need one for namespaced installations and another for cluster installations

gtully:
i may have this wrong, but was the default from the sdk to create a single yaml and we had some scripts to split it up? It looks kike we now have a script to put it back together?

brusdev:
By default, operator-sdk init scaffolds a cluster-scoped operator and only provides the Makefile target `deploy` to deploy the operator to a K8s cluster, see https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/

I see your points, a single yaml for cluster-scoped operator makes sense.

brusdev (approved):

gaohoward:
> i may have this wrong, but was the default from the sdk to create a single yaml and we had some scripts to split it up? It looks kike we now have a script to put it back together?

Yes it basically comes from the sdk make target, we only need to output it to a single static file for easy access. (for local dev you can just use make deploy to deploy a cluster-scoped opr in one go)

Issue #680, opened 2023-09-12
Provide an easy way to provide extra broker properties - extra mount -bp


gtully:
The solution at https://github.com/artemiscloud/activemq-artemis-operator/issues/598#issuecomment-1570443852
provides a file with extra properties and augments the broker env to reference the file extra mount.

This capability could be encapsulated as a -bp prefix on the resource, having the operator do the necessary updates to the env etc. Only the -bp is relevant, meaning treat as an additional broker properties and track it in the CR Status. 

extraMounts:
      secrets:
        - address-settings-bp

PR #755, opened 2023-12-11
[#680] support -bp extra mount secret

gtully:


gaohoward:
looks good. :)

controllers/activemqartemis_controller.go
```diff
@@ -719,11 +719,11 @@ type inSyncApplyError struct {
 	detail map[string]string
 }
 
-const inSyncWithErrorCause = "some properties resulted in error on pod %s"
+const inSyncWithErrorCause = "some properties from %v resulted in error on pod %s"
```
brusdev, line 722: why '%v' and not `%s`?

docs/help/operator.md
```diff
@@ -801,6 +801,12 @@ spec:
     - globalMaxSize=512m
 ```
 
+## Providing additional brokerProperties configuration from a secret
+It is possible to replace the use of the activemqartemisaddresses CRD and much of the activemqartemissecurities CRD with configuration via broker properties. This can necessitate a large amount of configuration in the CR.brokerPropertis field.
+In order to provide a way to split or orgainse these properties by file or by secret, an extra mount can be used to provide a secret that will be treated as an additional source of broker properties configuration.
+Using an **extraMounts** secret with a suffix "-bp" will cause the operator to auto mount the secret and make the broker aware of it's location. In addition the CR.Status.Condition[BrokerPropertiesApplied] will reflect the content of this secret.
+Broker properties are applied in order, starting with the CR.brokerProperties and then with the "-bp" auto mounts in turn. Keys (or property files) from secrets are applied in alphabetical order.
```
brusdev, line 808: Does a test that checks `Keys (or property files) from secrets are applied in alphabetical order` already exist?

brusdev (approved):
I love convention over configuration +1000

docs/help/operator.md
gtully: yep, in the broker: https://github.com/apache/activemq-artemis/commit/7bc3b0280946811a65e6019f0e1ccba98097bf47#diff-c1a1b087900107abca96cc3dd98ef776ee58ebe4182024ac380ad4d06ba9f281

controllers/activemqartemis_controller.go
gtully: was thinking of putting in the whole projection but settled on just name, %v works well on everything.

Issue #764, opened 2023-12-22
Check version when images are specified


brusdev:
The status of the validation condition is unknown when version and images are specified at the same time while the operator could compare the version specified with the version of the running brokers to check if the version specified match the broker version included in the image specified.

brusdev:
This enhancement will be included in 1.0.17

PR #765, opened 2023-12-22
[#764] Check version when images are specified

brusdev:


api/v1beta1/activemqartemis_types.go
```diff
@@ -741,4 +741,8 @@ const (
 	ConfigAppliedConditionUnknownReason                   = "UnableToRetrieveStatus"
 	ConfigAppliedConditionOutOfSyncReason                 = "OutOfSync"
 	ConfigAppliedConditionNoJolokiaClientsAvailableReason = "NoJolokiaClientsAvailable"
+
+	BrokerVersionAlignedConditionType           = "BrokerVersionAligned"
+	BrokerVersionAlignedConditionMatchReason    = "VersionMatch"
```
gtully, line 746: wondering if we need a reason when the condition is true, the condition type, BrokerVersionAligned is informative when true. But I see we do have a precedent for this with other positive reason strings.

gtully (approved):

api/v1beta1/activemqartemis_types.go
brusdev: I agree with you for "normal-true" conditions the reason is useless but the k8s status client fails when the reason is empty, i.e.:
```
... status.conditions[x].reason in body should be at least 1 chars long", Reason: "Invalid", Details: *(*"k8s.io/apimachinery/pkg/apis/meta/v1.StatusDetails")(0xc00055d0e0), Code: 422}}
```

gaohoward (approved):
LGTM

Issue #503, opened 2023-02-14
support headless services, listeners are set up for each IP:Port pair in the headless service.


ramu11:
proposal to add the port 7800 to the headless service makes sense to me:
https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services

refer [491](https://github.com/artemiscloud/activemq-artemis-operator/discussions/491)

brusdev:
This enhancement will be included in 1.0.17

PR #766, opened 2024-01-02
[#503] support jgroup in headless services for istio

johgoe:
I added the JGroup Port 7800 to the headless service. It's used in a Cluster to communicate with other nodes. With a service mesh like istio, all used ports must be defined or else the port is blocked.

### DIFF
--- a/pkg/resources/serviceports/service_port.go
+++ b/pkg/resources/serviceports/service_port.go
@@ -7,10 +7,18 @@ import (
 )
 
 var appProtocolHTTP = "http"
+var appProtocolTCP = "tcp"
 
 func GetDefaultPorts() *[]corev1.ServicePort {
 
 	ports := []corev1.ServicePort{
+		{
+			Name:        "jgroups",
+			Protocol:    "TCP",
+			Port:        7800,
+			AppProtocol: &appProtocolTCP,
+			TargetPort:  intstr.FromInt(int(7800)),
+		},
 		{
 			Name:        "console-jolokia",
 			Protocol:    "TCP",


gtully:
thanks for the contribution :-) 

I wonder if this use case can be handled with the new resourceTemplate _patch_ feature, it is designed to allow customisation of the resources that the operator creates. 
see: https://github.com/artemiscloud/activemq-artemis-operator/blob/main/docs/help/operator.md#custom-mods-via-a-strategic-merge-patch

Here is an example CR that will have the same effect as the code change: 

```
kind: ActiveMQArtemis
metadata:
  name: artemis-broker
spec:
  resourceTemplates:
  - selector:
     kind: Service
    patch:
     kind: Service
     spec:
      ports:
      - name: tcp-jgroup
        protocol: TCP
        port: 7800
        targetPort: 7800
```

gtully:
I guess I am also wondering why the jgroups discovery module would want to be exposed through istio, is it to have mtls everywhere?

johgoe:
Yes, we enabled istio everywhere for mtls and to limit access between services e.g. our wildfly service is allowed to access the jms ports, but other services are not.

You can find the istio requirement of exposed ports in headless services here: https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services

pkg/resources/serviceports/service_port.go
```diff
@@ -11,6 +11,12 @@ var appProtocolHTTP = "http"
 func GetDefaultPorts() *[]corev1.ServicePort {
 
 	ports := []corev1.ServicePort{
+		{
+			Name:       "tcp-jgroup",
```
brusdev, line 15: Could you change the port name to `jgroups-tcp` or just `jgroups`?

brusdev (commented):
The Jgroups 5 documentation also includes an examples of an headless service with the port 7800 for [DNS_PING](http://jgroups.org/manual5/index.html#_dns_ping)

pkg/resources/serviceports/service_port.go
johgoe: I can change it to jgroups and set the appProtocol to tcp. I used this name because of this documentation https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

brusdev: > I can change it to jgroups and set the appProtocol to tcp. I used this name because of this documentation https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

Your proposal LGTM +1

johgoe: I change it. It's ready to review again.

brusdev:
@johgoe your PR LGTM. Could you squash your commits?

johgoe:
@brusdev squash done.

brusdev:
@johgoe thanks for your contribution

Issue #557, opened 2023-03-30
Add configuration options for services and ingresses


Zodor:
I would like to have configuration options in the helmchart
Today we have no way of changing the typ of the services and setting the annotation which AWS EKS clusters are depending on. ClusterIP is not allowed when autocreating NLB's for example. We need to set NodePort och LoadBalancer.

**Example chart**

- New stuff below are the type and annotations under the acceptors, console and connectors

```
kind: ActiveMQArtemis
metadata:
  name: activemq-artemis-prod
spec:
  ingressDomain: example.io
  adminUser: admin
  adminPassword: blablabla
  deploymentPlan:
    size: 2
    image: quay.io/artemiscloud/activemq-artemis-broker-kubernetes:1.0.14
    messageMigration: true
    resources:
      limits:
        cpu: "1000m"
        memory: "2024Mi"
      requests:
        cpu: "500m"
        memory: "1024Mi"
  acceptors:
    - name: activemq
      port: 61616
      type: NodePort
      annotations:
          external-dns.alpha.kubernetes.io/hostname: amq-prod-0.example.io
          service.beta.kubernetes.io/aws-load-balancer-internal: "true"
          service.beta.kubernetes.io/aws-load-balancer-type: nlb
  console:
    expose: true
      type: NodePort
      annotations:
          service.beta.kubernetes.io/key: data
  addressSettings:
    addressSetting:
      - autoCreateJmsQueues: true
        autoCreateQueues: true
  connectors:
    - name: connectors
      host: amq-prod.example.io
      expose: true
      port: 61616
      annotations:
          alb.ingress.kubernetes.io/backend-protocol: HTTP

```

If the above could be implemented you have the options to install this in AWS EKS cluster and let EKS create Route53 dns records, setting up ALB/NLB's and you can connect to the brokers from outside the cluster in a easy way

gtully:
I wonder if this is getting to the point of probes, where we allowed an entire probe to be configured for liveness and readyness, reusing the corev1.Probe kind directly inline.

In this case, we would support a service or ingress or route definition in full, rather than providing some subset via an alias.

maybe worth a thought at this point. The simplest thing may be to just add some extra fields, but why not reference the existing kinds and  have the operator manage their lifecycle.

jseiser:
Would like to add onto this.

To deploy an ingress, we would need to be able to specify additioanl annotations, labels as well as specifying a specific Ingress Class.

[...]

PR #798, opened 2024-02-16
[#557] Add expose mode

brusdev:
The expose mode affects how the internal services for acceptors, connectors and console are exposed. Currently the supported modes are `route` and `ingress`. Default is `route` on OpenShift and `ingress` on Kubernetes.
* `route` mode uses OpenShift Routes to expose the internal service.
* `ingress` mode uses Kubernetes Nginx Ingress to expose the internal service with TLS passthrough.


gtully (approved):
Do we want to mention the default behavior when not set?
This allows consistent use of ingress. Nice!

brusdev:
Full test-suite executed.

brusdev:
Default behavior added to the field description:
```
Default is `route` on OpenShift and `ingress` on Kubernetes.
```

PR #800, opened 2024-02-19
[#557] Fix secure ingress on openshift

brusdev:


brusdev:
Full test-suite executed.

Issue #799, opened 2024-02-16
Restrict security context of pods by default


brusdev:
**Describe the feature**
The operator pod and the broker pods can run with a safer and restricted security context without issues. Restricting the security context of the pods by default allows to deploy a cluster of brokers also in Kubernetes namespaces with the [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) policy.

The minimal changes to run the pods in Kubernetes namespaces with the [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) policy:

- spec.containers[*].securityContext.allowPrivilegeEscalation: false
- spec.initContainers[*].securityContext.allowPrivilegeEscalation: false
- spec.ephemeralContainers[*].securityContext.allowPrivilegeEscalation: false

- spec.securityContext.runAsNonRoot: true
- spec.containers[*].securityContext.runAsNonRoot: true
- spec.initContainers[*].securityContext.runAsNonRoot: true
- spec.ephemeralContainers[*].securityContext.runAsNonRoot: true

- spec.securityContext.seccompProfile.type: RuntimeDefault
- spec.containers[*].securityContext.seccompProfile.type: RuntimeDefault
- spec.initContainers[*].securityContext.seccompProfile.type: RuntimeDefault
- spec.ephemeralContainers[*].securityContext.seccompProfile.type: RuntimeDefault

- spec.containers[*].securityContext.capabilities.drop: ["ALL"]
- spec.initContainers[*].securityContext.capabilities.drop: ["ALL"]
- spec.ephemeralContainers[*].securityContext.capabilities.drop: ["ALL"]

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
The security context of pods will be restricted by default in 1.1.0

PR #757, opened 2023-12-15
Add container level security context

Eric84626:
Fixes [683](https://github.com/artemiscloud/activemq-artemis-operator/issues/683)

### Type of change
Enhancement

### Description
- Add container level securityContext config
- Add unit test
- Documentation updated

PR #801, opened 2024-02-19
[#799] Restrict security context of pods by default

brusdev:
The operator pod and the broker pods can run with a safer and restricted security context without issues. Restricting the security context of the pods by default allows to deploy a cluster of brokers also in Kubernetes namespaces with the restricted policy.


brusdev:
I incorporated the commits from #757 in this PR to support an easy way for users to relax the security contexts.

jbertram:
My experience with the operator is pretty limited so I won't be able to provide any meaningful feedback.

docs/help/operator.md
```diff
@@ -1040,6 +1040,22 @@ spec:
 
 When deploying the above custom resource the operator will spread matching pods among the given topology
 
+### Container SecurityContext
+
+The ActiveMQArtemis custom resource offers a container level SecurityContext option for the broker holds security configuration that will be applied to the containers.
```
gtully, line 1045: the broker _that_ holds

Think that is missing

gtully (approved):
Looks great.

brusdev:
> My experience with the operator is pretty limited so I won't be able to provide any meaningful feedback.

@jbertram sorry it was my mistake

brusdev:
@howardgao I incorporated the commits from #743 in this PR test the operator deployment in a restricted namespace. I removed the changes related to the security contexts of the broker pods because they were causing a conflict with the commits from #757 and I split the test to be consistent.

gaohoward (approved):

kevinfrommelt:
@brusdev I just upgraded to 1.1.0 and it seems like the default security context that gets added is not compatible with the `activemq-artemis-broker-init` image for tags older than `1.0.25` b/c the default user in the image is `root`.
```
Warning  Failed     39m (x12 over 41m)   kubelet            Error: container has runAsNonRoot and image has non-numeric user (root), cannot verify user is non-root (pod: "dev-1-artemis-ss-2_dev-1(012c6399-3953-4291-b969-ed3999177ba4)", container: dev-1-artemis-container-init)
```
```
» docker inspect quay.io/artemiscloud/activemq-artemis-broker-init:1.0.23 | jq '.[0].Config.User'
"root"
» docker inspect quay.io/artemiscloud/activemq-artemis-broker-init:1.0.24 | jq '.[0].Config.User'
"root"
» docker inspect quay.io/artemiscloud/activemq-artemis-broker-init:1.0.25 | jq '.[0].Config.User'
"185"
```

Aside from updating to the latest, I can work around this by adding the following:
```
kind: ActiveMQArtemis
metadata:
  name: broker
spec:
  deploymentPlan:
    containerSecurityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsNonRoot: true
      runAsUser: 185
      seccompProfile:
        type: RuntimeDefault
```

brusdev:
@kevinfrommelt I confirm that the activemq-artemis-broker-init image for tags older than 1.0.25 have the default user `root` and they need your workaround to be used because for security reasons it is better using container as non root.

Issue #832, opened 2024-03-19
Support secure cluster connections


brusdev:
**Describe the feature**
The internal cluster connections use plain TCP connections by default. Add the support for secure cluster connections.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
The doc and test for secure cluster connections will be added in 1.1.1, see https://github.com/artemiscloud/activemq-artemis-operator/commit/e90a04418071be31b59d88adf06f608826c6213a

PR #833, opened 2024-03-19
[#832] Support secure cluster connections

brusdev:


controllers/activemqartemis_controller_test.go
```diff
@@ -9324,4 +9325,135 @@ var _ = Describe("artemis controller", func() {
...
+						Name:       "artemis",
+						Port:       61616,
+						SSLEnabled: true,
+						SSLSecret:  tlsSecretName,
+					},
+				}
+
+				crd.Spec.BrokerProperties = []string{
+					"connectorConfigurations.artemis.params.sslEnabled=true",
+					"connectorConfigurations.artemis.params.trustStorePath=/etc/" + tlsSecretName + "-volume/broker.ks",
```
gaohoward, line 9397: Should the trustStorePath point to `client.ts`?

brusdev: I don't think so because the client.ts should contain the client certificates the the broker server should trust. In this case the connector is used to connect to another broker server instances so the certificates to trust are included in the broker.ks.

docs/tutorials/ssl_broker_setup.md
```diff
@@ -201,6 +201,59 @@ Consumer ActiveMQQueue[TEST], thread=0 Consumer thread finished
 ```
 Now you get an idea how an SSL acceptor is configured and processed by the operator and see it in action!
 
+### Secure cluster connections
+The internal cluster connections rely on the internal acceptor listening on the port `61616` and the internal connector with the name `artemis`. They can be secured with the following steps, create a secret with the secure stores, enable ssl in the internal acceptor by using the acceptor fields `sslEnabled` and `sslSecret`, and enable ssl in the internal connector by using system properties
```
gtully, line 205: by using broker properties rather than system properties (I think)

gtully (approved):

Issue #821, opened 2024-03-08
Support per broker configuration for brokerProperties in -bp suffix secrets


gaohoward:
**Describe the feature**
broker-N currently not working if specified in the secret with -bp secrets

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

gaohoward:
broker-N currently not working if specified in the secret with -bp secrets

brusdev:
This enhancement will be included in 1.2.0

PR #834, opened 2024-03-21
[#821] Support broker-N in -bp suffix secrets

gaohoward:


controllers/activemqartemis_reconciler.go
```diff
@@ -2070,8 +2073,13 @@ func brokerPropertiesConfigSystemPropValue(customResource *brokerv1beta1.ActiveM
...
 		if strings.HasSuffix(extraSecretName, brokerPropsSuffix) {
-			// formated append to result with comma
-			result = fmt.Sprintf("%s,%s%s/", result, secretPathBase, extraSecretName)
+			if len(secretBrokerPropertiesData[extraSecretName]) == 1 {
+				// formated append to result with comma
+				result = fmt.Sprintf("%s,%s%s/", result, secretPathBase, extraSecretName)
+			} else {
+				// append to ordinal path
+				result = fmt.Sprintf("%s,%s%s/,%s%s/%s${STATEFUL_SET_ORDINAL}/", result, secretPathBase, extraSecretName, secretPathBase, extraSecretName, OrdinalPrefix)
+			}
```
brusdev, line 2082: Secrets created by a customer could also include no broker.properties files. The check is not required, the ordinal path could be always appended because the broker skips not existing dirs, see https://github.com/apache/activemq-artemis/blob/2.33.0/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java#L551
```suggestion
			// append to ordinal path
			result = fmt.Sprintf("%s,%s%s/,%s%s/%s${STATEFUL_SET_ORDINAL}/", result, secretPathBase, extraSecretName, secretPathBase, extraSecretName, OrdinalPrefix)
```

controllers/activemqartemis_reconciler.go
```diff
@@ -2355,6 +2368,34 @@ func brokerPropertiesData(props []string) map[string]string {
...
+			if err := resources.Retrieve(bpSecretKey, client, bpSecret); err != nil {
+				return nil, err
+			}
+
+			secretData := make(map[string]string)
+
+			//broker-N.<prop-file-key> contains per broker configurations
+			for key, data := range bpSecret.Data {
+				secretData[key] = string(data)
```
brusdev, line 2391: The data size could be very huge and it is not used, only the keys are required to mount indexed broker.properties files

controllers/activemqartemis_reconciler.go
```diff
@@ -2495,6 +2536,21 @@ func (r *ActiveMQArtemisReconcilerImpl) createExtraConfigmapsAndSecretsVolumeMou
 					}
 				}
 			}
+			if strings.HasSuffix(secret, brokerPropsSuffix) {
+				propsData, ok := bpBrokerPropsData[secret]
```
brusdev, line 2540: The reading of the keys of the secrets with brokerPropsSuffix could be moved here to avoid to keep in memory a map with all secrets.

brusdev (commented):
Nice feature! In this way users can include in secretes sensitive broker properties for a specific broker instance.

controllers/activemqartemis_reconciler.go
gaohoward: +1

controllers/activemqartemis_reconciler.go
gaohoward: +1

controllers/activemqartemis_reconciler.go
gaohoward: +1

controllers/activemqartemis_reconciler.go
```diff
@@ -2495,13 +2501,40 @@ func (r *ActiveMQArtemisReconcilerImpl) createExtraConfigmapsAndSecretsVolumeMou
...
+				if err := resources.Retrieve(bpSecretKey, client, bpSecret); err != nil {
+					return nil, nil, err
+				}
+
+				if len(bpSecret.Data) > 0 {
+					bpKeysMap := make(map[string]string)
+					for bpk := range bpSecret.Data {
+						bpKeysMap[bpk] = ""
+					}
+					for _, key := range sortedKeys(bpKeysMap) {
```
brusdev, line 2520: bpKeysMap could be avoided by using `sortedKeysStringKeyByteValue`, i.e.
```suggestion
					for _, key := range sortedKeysStringKeyByteValue(bpSecret.Data) {
```

controllers/activemqartemis_controller_test.go
```diff
@@ -9244,6 +9244,54 @@ var _ = Describe("artemis controller", func() {
...
+					"journal1.properties":       `journalFileSize=12345`,
+					"broker-0.first.properties": `globalMaxSize=512m`,
+				}
+			})
+
+			_, bpSecret2 := DeploySecret(defaultNamespace, func(candidate *corev1.Secret) {
+				candidate.Name = "config-2-bp"
+				candidate.StringData = map[string]string{
+					"journal2.properties":        `journalFileSize=12345`,
+					"broker-0.second.properties": `globalMaxSize=512m`,
```
gtully, line 9263: it may make sense to use different values for first and second, so they can be differenciated later

controllers/activemqartemis_controller_test.go
```diff
@@ -9244,6 +9244,54 @@ var _ = Describe("artemis controller", func() {
...
+				By("verify status ok")
+				Eventually(func(g Gomega) {
+					g.Expect(k8sClient.Get(ctx, brokerKey, createdCrd)).Should(Succeed())
+
+					condition := meta.FindStatusCondition(createdCrd.Status.Conditions, brokerv1beta1.ConfigAppliedConditionType)
+					g.Expect(condition).NotTo(BeNil())
+					g.Expect(condition.Status).To(Equal(metav1.ConditionTrue))
+
+				}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
```
gtully, line 9287: maybe also make a jolokia call to get the status or to check the expected value for globalMaxSize on each instance

controllers/activemqartemis_reconciler.go
gaohoward: +1

controllers/activemqartemis_controller_test.go
gaohoward: +1

brusdev (approved):

controllers/activemqartemis_controller_test.go
```diff
@@ -9322,6 +9322,66 @@ var _ = Describe("artemis controller", func() {
...
+				candidate.Name = "config-2-bp"
+				candidate.StringData = map[string]string{
+					"journal2.properties":        `journalMinFiles=3`,
+					"broker-0.second.properties": `journalFileSize=12M`,
+				}
+			})
+
+			By("Deploying the CRD")
+			_, crd := DeployCustomBroker(defaultNamespace, func(candidate *brokerv1beta1.ActiveMQArtemis) {
+				candidate.Spec.DeploymentPlan.Size = common.Int32ToPtr(1)
```
gtully, line 9347: I think this test needs to go further. deploy two instanced of the ss. and have in config-1-bp
					"broker-0.first.properties": `globalMaxSize=512m`,

and config-2-bp:
					"broker-1.second.properties": `globalMaxSize=12m`,

Then verify via jolokia that, ordinal 0 has globalMax 512m and ordinal 1 has 12M 

just validating the Status presence of the property names may not be sufficient

docs/help/operator.md
```diff
@@ -844,9 +844,41 @@ spec:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: config-1-bp
+stringData:
+  journal1.properties: |
+    journalFileSize=12345
+  broker-0.first.properties: |
+    globalMaxSize=512m
+  broker-1.first.properties: |
```
gtully, line 864: in the test there is first and second. I think they can both be called the same name, but it may be worth being consistent...

also, maybe call it: globalMemory.properties - to indicate that the config maps can help to categorise the config.

controllers/activemqartemis_controller_test.go
gtully: whether there is a need for first and second .properties, or whether the could both use broker-N.globalMem.properties: globalMaxSixe=X 
it may also help for the doc. Same comment in the doc section.

gtully (changes requested):

controllers/activemqartemis_controller_test.go
gaohoward: +1

docs/help/operator.md
gaohoward: +1

gtully (approved):

Issue #837, opened 2024-03-22
Allow configuring priorityClassName for broker deployment


KevinFrommeltAtFanatics:
**Describe the feature**
Add the ability to configure the `priorityClassName` for the broker pods. We would like to be able to set the `priorityClassName` to something other than the default.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
@KevinFrommeltAtFanatics you can customize the priorityClassName and other podSpec fields by using a strategic merge patch, for further details see [custom mods via a strategic merge patch](https://github.com/artemiscloud/activemq-artemis-operator/blob/1.1.0/docs/help/operator.md#custom-mods-via-a-strategic-merge-patch), i.e.
```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata: 
  name: artemis-broker
spec:
  resourceTemplates:
    - selector:
        kind: StatefulSet
      patch:
        kind: StatefulSet
        spec:
          template:
            spec:
              priorityClassName: high-priority
```

Issue #758, opened 2023-12-15
add a patch capability to the resource template, to allow arbitrary customisation of managed resources via the CR


gtully:
the resourceTemplate #588 allow annotations and labels to be specified on managed kube resources.
Often it is necessary to provide further tweaks to the resources that the operator creates on behalf of a CR. #683 #557 #630
One option is to have a user use `kubectl patch` but such mods need to be reapplied.
Having the ability to include a patch as part of a resource template, would make the patch mod part of the CR and the operator could apply the mods before applying to the apiserver. Strategic merge is now the default with merge behaviour controlled by annotations on the kube api resources themselves, the behaviour is well  understood.

PR #838, opened 2024-03-23
[#837] Documentation and test for setting PriorityClassName

kevinfrommelt:
^

### DIFF
--- a/controllers/activemqartemis_reconciler_test.go
+++ b/controllers/activemqartemis_reconciler_test.go
@@ -971,6 +971,60 @@ func TestProcess_TemplateCustomAttributeContainerSecurityContext(t *testing.T) {
 	assert.True(t, runAsRootOk)
 }
 
+func TestProcess_TemplateCustomAttributePriorityClassName(t *testing.T) {
+
+	var kindMatchSs string = "StatefulSet"
+
+	cr := &brokerv1beta1.ActiveMQArtemis{
+		ObjectMeta: metav1.ObjectMeta{Name: "cr"},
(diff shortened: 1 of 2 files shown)


kevinfrommelt:
Looks like the tests don't currently run b/c of this issue https://github.com/kubernetes-sigs/controller-runtime/issues/2720

brusdev (commented):
@kevinfrommelt thanks for your PR but you could already customize the priorityClassName and other podSpec fields by using a strategic merge patch, for further details see https://github.com/artemiscloud/activemq-artemis-operator/issues/837#issuecomment-2016538868

After #758 new fields to customize the resource managed by the operator are added only for use cases in which a [strategic merge patch](https://github.com/artemiscloud/activemq-artemis-operator/blob/1.1.0/docs/help/operator.md#custom-mods-via-a-strategic-merge-patch) doesn't work or would be too much complex.

You could change your PR to add a doc and a test to set the `priorityClassName` by using a [strategic merge patch](https://github.com/artemiscloud/activemq-artemis-operator/blob/1.1.0/docs/help/operator.md#custom-mods-via-a-strategic-merge-patch) if that works for your use case.

kevinfrommelt:
@brusdev Thanks for the info. I'll update the PR to just include docs and tests like you suggested.

kevinfrommelt:
@brusdev This is ready for review

brusdev (approved):
@kevinfrommelt great work!!!

brusdev:
@kevinfrommelt thanks for your contribution and for investigating the CI error, I fixed it in #839

Issue #845, opened 2024-04-03
Make: building fails with go 1.21.8


lavocatt:
Building with the golang version 1.21.8 leads to errors:

```
make build
test -s /home/tlavocat/dev/activemq-artemis-operator/bin/controller-gen && /home/tlavocat/dev/activemq-artemis-operator/bin/controller-gen --version | grep -q v0.11.2 || \
GOBIN=/home/tlavocat/dev/activemq-artemis-operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.11.2
/home/tlavocat/dev/activemq-artemis-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet -composites=false ./...
go: updates to go.mod needed; to update it:
	go mod tidy
make: *** [Makefile:137: vet] Error 1
```

After discussing the issue with @brusdev, we propose to add some checks at the beginning of the build process based on the content of the `go.mod` file to check that the used go version is in a range of supported versions.

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/fe5b44439d054eb620c1c74b204b12490703a735

PR #848, opened 2024-04-03
[#845] Make: check for go version compatibility

lavocatt:
Building with golang 1.21 requires changes to the go.mod file. This is most likely due to a change in behavior in golang's gomod (see https://go.dev/doc/modules/gomod-ref).

Since the migration to the newer version of go hasn't yet been made, and to help any user who might have a > 1.20 version of go on their system, the Makefile now brings in a new check to verify that the go version matches the one of the go.mod.


brusdev (approved):
Nice work! This check will help new users.

Issue #849, opened 2024-04-04
The cert-manager certificates from annotations don't work


brusdev:
**Describe the bug**
The cert-manager certificates from annotations are created after the deployment but the deploy fails because the manager check if the ssl secret exists:

```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: aa-con-cer-man185a
spec:
  acceptors:
    - expose: true
      exposeMode: ingress
      ingressHost: aa-con-cer-man185a.tests.artemiscloud.io
      name: tls
      port: 61617
      sslEnabled: true
      sslSecret: aa-con-cer-man185a-tls-0-svc-ing-test-secret
  resourceTemplates:
    - selector:
        kind: Ingress
        name: aa-con-cer-man185a-tls-0-svc-ing
      annotations:
        cert-manager.io/issuer: aa-con-cer-man185a-issuer
      patch:
        kind: Ingress
        spec:
          tls:
            - hosts:
                - aa-con-cer-man185a.tests.artemiscloud.io
              secretName: aa-con-cer-man185a-tls-0-svc-ing-test-secret
```

```
status:
  conditions:
    - lastTransitionTime: '2024-04-04T08:50:58Z'
      message: required secret not exist aa-con-cer-man185a-tls-0-svc-ing-test-secret
      reason: ResourceError
      status: 'False'
      type: Deployed
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #850, opened 2024-04-04
[#849] Support certificates from annotations

brusdev:


controllers/activemqartemis_controller_test.go
```diff
@@ -9598,14 +9598,62 @@ var _ = Describe("artemis controller", func() {
 			}
 		})
 
-		It("secure connections with multiple wildcard DNS names", func() {
+		It("secure connections with openshift serving certificate", func() {
+			if os.Getenv("USE_EXISTING_CLUSTER") == "true" && isOpenshift {
+				crd := generateArtemisSpec(defaultNamespace)
+
+				tlsSecretName := crd.Name + "-cert-provided-secret"
```
gtully, line 9605: nice, provided seems the perfect name and naming convention over config is a wise approach

pkg/utils/certutil/certutil.go
```diff
@@ -27,6 +27,10 @@ const (
 	Cert_annotation_key   = "cert-manager.io/issuer-name"
 	Bundle_annotation_key = "trust.cert-manager.io/hash"
 	Console_web_prefix    = "webconfig.bindings.artemis."
+
+	// Suffix of provided secrets that contain tls.crt and tls.key items in PEM format.
+	// Secrets with this suffix are not validated. It is useful for secrets created from annotations.
```
gtully, line 32: because they typically won't exist till the annotations are processed and the secret is 'provided' by some other controller.

maybe add in the above text or some variant

gtully (approved):
looks great, and nice to be able to build on the cert manager bits.

controllers/activemqartemis_controller_test.go
brusdev: I had to reduce the length of the suffix to avoid issues for names longer than 63 characters. The new suffix is `cp-secret`

pkg/utils/certutil/certutil.go
brusdev: I added that exact text, thanks for the suggestion

Issue #858, opened 2024-04-09
Validate brokerProperties and -bp content does not have duplicate keys to avoid OutOfSync condition


gtully:
If duplicate key exist in a java properties file, they are ignored by the reader and the last one wins. The broker does a checksum on the properties that it applies, but this excludes duplicates.
The operator does a checksum on the content  of a -bp secret and on the contents of the .Spec.BrokerProperties, however it currently does not check for duplicates. This means that if by accident there is a duplicate key on the content, the operator will report the OutOfSync condition in error.
It would be good to validate the CR and -bp secrets to trap duplicate keys as a validation error and avoid this potential problem that can be tricky to diagnose.

PR #859, opened 2024-04-09
[#858] validate no duplicate keys in brokerProperties or -bp secret c…

gtully:
…ontent - avoid OutOfSync in error


brusdev (approved):
Nice improvement. I'm sure users will be happy of it!!!

PR #860, opened 2024-04-09
[#849] Rename suffix of certificate provided secrets

brusdev:


gtully (commented):
I think we considered -cp for classpath, that is what comes to my mind.

is the motivation to keep the name small to not reach some limit? 
to keep it small, I would use -p and -provided otherwise.

brusdev:
> I think we considered -cp for classpath, that is what comes to my mind.
> 
> is the motivation to keep the name small to not reach some limit? to keep it small, I would use -p and -provided otherwise.

@gtully yes the main purpose is to keep the name small to not reach some limit. I would like to add something related the certificates because that suffix is used for `provided secrets that contain tls.crt and tls.key items in PEM format`. What about `-pt` tah should mean provided tls that is the type of [tls secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets)?

gtully:
I was putting the emphasis on the `provided` bit because the context will be in the attribute - tlsSecretName or sslSecretName etc.
maybe -pp for provided-pem (but I think -p for provided may be something we can reuse).

For sure the -secret is duplicate information, so I agree it needs to change.

brusdev:
I changed the suffix to `-ptls` as discussed offline.

gtully (approved):

Issue #266, opened 2022-07-06
Remove OPERATOR_WATCH_NAMESPACE - duplicate of SDK env var WATCH_NAMESPACE


gtully:
With the watch namesace support in the sdk, and with no evidence of this env var being used, lets remove as it now duplicates of WATCH_NAMESPACE 
Worst case scenario a user can add a kube env var that references WATCH_NAMESPACE

The NAMESPACE env var, that is not a duplicate, maybe we leave that one in place, but again, who uses it and why?

brusdev:
This project is now part of the ArkMQ initiative. This repository has been archived as all activities are now happening in the [corresponding ArkMQ repository](https://github.com/arkmq-org/activemq-artemis-operator). See [here](https://artemiscloud.io/community/transition/) for the ArkMQ transition details.

Feel free to create a new issue at https://github.com/arkmq-org/activemq-artemis-operator/issues

PR #862, opened 2024-04-11
[#266] use WATCH_NAMESPACE instead of OPERATOR_WATCH_NAMESPACE

lavocatt:
Replace the custom env variable OPERATOR_WATCH_NAMESPACE with WATCH_NAMESPACE.

@gtully I'm wondering though if that's what you had in mind in the issue or if you meant to remove all the code related to `OPERATOR_WATCH_NAMESPACE`. If it's the later, then we'll have to also rethink the `isLocalOnly` function. Let me know your thoughts on this.

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -546,7 +546,7 @@ func (r *ActiveMQArtemisReconcilerImpl) syncMessageMigration(customResource *bro
 
 func isLocalOnly() bool {
 	oprNamespace := os.Getenv("OPERATOR_NAMESPACE")
-	watchNamespace := os.Getenv("OPERATOR_WATCH_NAMESPACE")
+	watchNamespace := os.Getenv("WATCH_NAMESPACE")
 	return oprNamespace == watchNamespace
 }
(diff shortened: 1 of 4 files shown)


gtully:
I opened this from the observation that there seemed to be some duplication here, and though to remove all uses of OPERATOR_WATCH_NAMESPACE assuming that the same capability would be available through WATCH_NAMESPACE and assuming that both would be set.

But it is typically never that simple, and removing an env var is usually non trivial from a backward comparability point of view.
If it is a simple swap, and WATCH_NAMESPACE is always present when OPERATOR_WATCH_NAMESPACE is set, the I think we can do it.

In short, this issue has been open for a while because it is not clear what the value or the potential impact is.


Part of the context was in moving to an operator that is only cluster wide, but even then, it may be good to be able to limit the cluster wide operation, but if that was the case, it should be an operator sdk deployment thing, rather than something that was unique to our operator.

In short, I don't know for sure and have not investigated in any detail.

lavocatt:
I'm closing this PR for now, since it's stale and I'm working on other topics atm.

Issue #865, opened 2024-04-11
Add tests to verify ArtemisAddressCR use cases with broker properties


gtully:
The support for addresses and queues and address settings is mature in brokerProperties. We need more tests to validate the artemis address cr usecases from our tests with properties versions

PR #866, opened 2024-04-11
[#865] version of address cr tests with broker properties

gtully:


controllers/activemqartemisaddress_controller_test.go
```diff
@@ -975,12 +975,13 @@ func QueueStatAssertionInPod(brokerCrName string, podName string, namespace stri
...
+		Name:      podName,
+		Namespace: namespace,
+	}
+	pod := &corev1.Pod{}
 	Eventually(func(g Gomega) {
-		key := types.NamespacedName{Name: namer.CrToSS(brokerCrName), Namespace: namespace}
-		sfsFound := &appsv1.StatefulSet{}
-
-		g.Expect(k8sClient.Get(ctx, key, sfsFound)).Should(Succeed())
-		g.Expect(sfsFound.Status.ReadyReplicas).Should(BeEquivalentTo(1))
```
brusdev, line 983: Is the following check removed on purpose?
```
g.Expect(sfsFound.Status.ReadyReplicas).Should(BeEquivalentTo(1))
```

examples/address/artemis_broker_properties_address.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  brokerProperties:
+   - "# anycast, consumers compete"
+   - addressConfigurations."APP.JOBS".routingTypes=ANYCAST
+   - addressConfigurations."APP.JOBS".queueConfigs."APP.JOBS".routingType=ANYCAST
+   - "# mulitcast, consumers see all messages"
+   - addressConfigurations."APP.COMMANDS".routingTypes=MULTICAST
+   - "# allow delete (this is the source of truth)"
+   - addressSettings."APP.#".configDeleteAddresses=FORCE
+   - addressSettings."APP.#".configDeleteAddresses=FORCE
```
brusdev, line 14: Is the following duplicate line required?
```suggestion
   - addressSettings."APP.#".configDeleteAddresses=FORCE
```

brusdev (approved):
Great work!!! Those tests are another confirmation of the power of the broker properties.

controllers/activemqartemisaddress_controller_test.go
gtully: yes, so that it can be used on a CR with size N != 1

examples/address/artemis_broker_properties_address.yaml
gtully: that is an error, good catch, and it should be invalid if I had the latest operator running.
it should be configDeleteQueues

will fix.

PR #875, opened 2024-04-17
Fix Amq Broker Operator Scorecard Test Warnings

Msarawan:
Fixed scorecard test suggestions/warnings.


brusdev (commented):
@Msarawan your PR LGTM could you regenerate the manifests to fix the CI failure?
```
make generate-deploy && make bundle
```

Msarawan:
> @Msarawan your PR LGTM could you regenerate the manifests to fix the CI failure?
> 
> ```
> make generate-deploy && make bundle
> ```

Sure Domenico, i will update the branch by regenerating manifests. Thanks!

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
```diff
@@ -37,6 +37,9 @@ metadata:
                 "name": "activemq"
               }
             }
+          },
+          "status": {
+            "conditions": null
```
brusdev, line 42: is this null value required?

gaohoward (approved):

config/manifests/bases/activemq-artemis-operator.clusterserviceversion.yaml
```diff
@@ -2457,6 +2513,7 @@ spec:
   - email: rkieley@redhat.com
     name: Roderick Kieley
   maturity: beta
+  minKubeVersion: 1.25.0
```
brusdev, line 2516: Can you use a lower version like `1.20.0` unless there is a technical requirement to avoid backward compatibility issues for users running the operator on an outdated Kubernetes cluster?

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
Msarawan: @brusdev  The null values are auto generated. They don't represent any real conditions. but the `status.conditions` has to be there in order to pass the test.

config/manifests/bases/activemq-artemis-operator.clusterserviceversion.yaml
Msarawan: Sure, i will check with adding `1.20.0` and re-run the tests once again.

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
brusdev: I guess you need an non-empty status to fix the scorecard test warning. If this is the case you could use an empty array for conditions to avoid the null value in the examples, i.e.
```suggestion
            "conditions": []
```

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
```diff
@@ -211,6 +247,9 @@ metadata:
               "image": "placeholder",
               "size": 2
             }
+          },
+          "status": {
+            "podStatus": null
```
brusdev, line 252: Could you use a condition with an empty array also in this example?
```suggestion
            "conditions": []
```

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
```diff
@@ -224,6 +263,9 @@ metadata:
...
```
brusdev, line 268: Could you use a condition with an empty array also in this example?
```suggestion
            "conditions": []
```

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
Msarawan: sure, i will modify with empty array here.

bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml
Msarawan: Yes, if it's okay to keep `[]` to the `podStatus` filed then i will add there. but, we can replace the `podStatus` filed with `conditions` field. because the `podStatus` key  is added by referring the status field of `activemqartemis_types.go` file.

brusdev (approved):

Issue #878, opened 2024-04-17
Add tests that exercise broker VIEW and EDIT role security settings with the operator 


gtully:
Add some tests to validate the artemis security cr use case with brokerProperties and make use of the artemis VIEW/EDIT roles for jolokia access.

PR #879, opened 2024-04-17
[#878] version of security cr tests with broker properties and view/e…

gtully:
…dit roles


brusdev (approved):
This is very promising!

Issue #880, opened 2024-04-17
Add Deprecation to Security CR in favour of Artemis.Spec.brokerProperties


gtully:
The --jaas-config secret extra mount provides a means to configure jaas. brokerProperties allow roles and rbac to be configured. the VIEW/EDIT roles allow rbac for jmx/jolokia. This combined features make the security cr obsolete.

Issue #877, opened 2024-04-17
Add Deprecation to Address CR in favour of Artemis.Spec.brokerProperties


gtully:
brokerProperties an -bp secrets provide a consistent, reliable and extensible way to specify addresses and queues that makes use of artemis config reload from disk.
see example: https://github.com/artemiscloud/activemq-artemis-operator/blob/8d36322afb9befe5187c3bc61eba6491917ae6e3/examples/address/artemis_broker_properties_address.yaml#L7

brusdev:
The field `spec.brokerProperties` of the ActiveMQArtemis CR can replace ActiveMQArtemisAddress CR allowing a single source of truth for a deployment of ActiveMQ Artemis on Kubernetes.

The brokerProperties secrets (with the suffic -bp) are only an alternative of the the field `spec.brokerProperties` to split the brokerProperties accross multiple files if that is required.

PR #881, opened 2024-04-17
[#877 #880] deprecate address,security and addressSettings in favour …

gtully:
…of brokerProperties

### DIFF
--- a/api/v1alpha1/activemqartemissecurity_types.go
+++ b/api/v1alpha1/activemqartemissecurity_types.go
@@ -404,6 +404,7 @@ type ActiveMQArtemisSecurityStatus struct {
 //+kubebuilder:resource:path=activemqartemissecurities,shortName=aas
 //+operator-sdk:csv:customresourcedefinitions:resources={{"Secret", "v1"}}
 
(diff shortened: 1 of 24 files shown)


brusdev (commented):
@gtully what about enforcing the deprecation with the [version deprecation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation)?

gtully:
That assumes a change in the version. We have kept v1beta1 for some time. I think it would only make sense if we were changing the version

gtully:
but it does give better feedback. I will investigate a little. thanks for the idea :-)

gtully:
added the annotation, i did not yet get the api server to return the warning, that is with mk and the updated crds

gtully:
success!
```
[activemq-artemis-operator]$ kubectl apply -f examples/address/address_queue.yaml 
Warning: The ActiveMQArtemisAddress CRD is deprecated. Use the spec.brokerProperties attribute in the ActiveMQArtemis CR to create addresses and queues instead
activemqartemisaddress.broker.amq.io/artemis-address-queue created
```

Issue #885, opened 2024-04-19
Add ready printer column to ActiveMQArtemis CRD


brusdev:
**Describe the feature**
The kubectl tool relies on server-side output formatting. Your cluster's API server decides which columns are shown by the kubectl get command. You can customize these columns for a CustomResourceDefinition, see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/426af6b7110e8f8375badb0d75540bb4c5676567

PR #886, opened 2024-04-22
[#885] Add ready printer column to ActiveMQArtemis CRD

brusdev:
The kubectl tool relies on server-side output formatting. Your cluster's API server decides which columns are shown by the kubectl get command. You can customize these columns for a CustomResourceDefinition, see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

```
$ kubectl apply -f examples/artemis/artemis_single.yaml 
activemqartemis.broker.amq.io/artemis-broker created
```

```
$ kubectl get aa --watch
NAME             READY   AGE
artemis-broker   False   3s
artemis-broker   True    11s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,13 +123,13 @@ ifeq ($(ENABLE_WEBHOOKS),true)
 ## v2alpha3, v2alpha4 and v2alpha3 requires allowDangerousTypes=true because they use float32 type
 	cd config/manager && $(KUSTOMIZE) edit add resource webhook_secret.yaml 
 	$(CONTROLLER_GEN) rbac:roleName=$(OPERATOR_CLUSTER_ROLE_NAME) crd:allowDangerousTypes=true webhook paths="./..." output:crd:artifacts:config=config/crd/bases
-	find config -type f -exec sed -i.bak -e '/creationTimestamp/d' {} \; -exec rm {}.bak \;
+	find config -type f -exec sed -i.bak -e '/creationTimestamp:/d' {} \; -exec rm {}.bak \;
 else
 ## Generate ClusterRole and CustomResourceDefinition objects.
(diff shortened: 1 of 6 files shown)


lavocatt:
I'm wondering it would be good to have the same status than the one in the OpenShift plugin? Over there it says `5 ok / 5` when all the conditions are met. Wdyt?

gtully (approved):
nice, it is great to expose our Ready status.

gtully:
it begs the question, when false, what is the next command to run, to see the false conditions. 
afaik, currently we don't populate or evaluate all the conditions once there is a false so we don't have the x/N information of ok statuses.

gtully:
> I'm wondering it would be good to have the same status than the one in the OpenShift plugin? Over there it says `5 ok / 5` when all the conditions are met. Wdyt?

I think we short cut in the status generation on the first false. We would need to improve on that to have an X ok/N detail

brusdev:
> > I'm wondering it would be good to have the same status than the one in the OpenShift plugin? Over there it says `5 ok / 5` when all the conditions are met. Wdyt?
> 
> I think we short cut in the status generation on the first false. We would need to improve on that to have an X ok/N detail

I see your point but I also like the current status that only show the first fatal error because in some way drives the users to resolve all error in an ordered sequence: validation, deployment, running

brusdev:
> I'm wondering it would be good to have the same status than the one in the OpenShift plugin? Over there it says `5 ok / 5` when all the conditions are met. Wdyt?

Sorry I missed your suggestion, it LGTM but that requires a new field  in the status. Can you create a new enhancement issue?

lavocatt:
Here it is: https://github.com/artemiscloud/activemq-artemis-operator/issues/903

Issue #867, opened 2024-04-12
add a tutorial for the send receive usecase


lavocatt:
**Describe the feature**

Write a new tutorial for the send and receive usecase over minikube.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/8d4fa779d2277455bd0d584e6c4f59ebb77bc3c3

PR #890, opened 2024-04-22
[#867] move up the send receive tutorials

lavocatt:
The two tutorials were created in a subfolder, that was a mistake.


brusdev (commented):
Without the `send-receive` parent folder the file names are less intuitive. Could you add a prefix like `send_receive_with_` ?

brusdev (approved):

PR #895, opened 2024-04-23
[#897] simplify api detection logic and report errors

gtully:


gtully:
I toyed with putting the detection in the reconcile loop but now realise we will need it for controller creation, with the selective watches. it is a little cleaner now, checked upfront and stored as an attribute by the controller that cares.

brusdev (approved):
This is a great improvement! The code is cleaner and more readable.

Issue #898, opened 2024-04-24
Make the operator periodically reconcile


gaohoward:
**Describe the feature**
The reconcile should be performed periodically to recover any dependent resources accidentally deleted.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #899, opened 2024-04-24
[#898] Make the operator periodically reconcile

gaohoward:


gtully:
I am not convinced on this one, maybe we need a different variable for our regular retrys, and keep the this as a real periodic resync option when > 0

We currently retry on failure and make use of this period, it would make sense to split that out, b/c it would also make sense to not do periodic re sync most of the time unless there are external resources, but we already have that logic, just that it uses this variable.

gaohoward:
> I am not convinced on this one, maybe we need a different variable for our regular retrys, and keep the this as a real periodic resync option when > 0
> 
> We currently retry on failure and make use of this period, it would make sense to split that out, b/c it would also make sense to not do periodic re sync most of the time unless there are external resources, but we already have that logic, just that it uses this variable.

It makes sense to use a different var from the one we already uses. Doing regular resync is not for external resources specifically if `external resources` means extra mounts or such that. It is for keeping the resources associated with a CR deployment intact (that includes services/secrets/configmaps/ingresses/routes). So that when they are accidentally destroyed the next re-sync can recreate them.

gtully:
for anythinng we create, we should watch with our owner reference. We currently do that only for Pods and SS, we had a difficulty with external resources b/c we don't own them, but I think we should just add routes and services to our watch list and the existing periodic reconcile when we have extra mounts will work for the other cases.
see: https://github.com/artemiscloud/activemq-artemis-operator/blob/main/controllers/activemqartemis_controller.go#L744

gaohoward:
@gtully I'll close this (also the assocated issue) and raise a new PR. Thanks!

Issue #900, opened 2024-04-24
The operator fails to detect OpenShift fails on ROSA


brusdev:
**Describe the bug**
The operator sometimes fails to detect OpenShift on ROSA with the followign error:
```
<*url.Error | 0xc001099950>: 
Get "https://api.j5voq-aztbf-vv4.hqo4.p3.openshiftapps.com:443/apis/operator.openshift.io/v1?timeout=32s": proxyconnect tcp: EOF
{
    Op: "Get",
    URL: "https://api.j5voq-aztbf-vv4.hqo4.p3.openshiftapps.com:443/apis/operator.openshift.io/v1?timeout=32s",
    Err: <*net.OpError | 0xc000a843c0>{
        Op: "proxyconnect",
        Net: "tcp",
        Source: nil,
        Addr: nil,
        Err: <*errors.errorString | 0xc0000a4120>{s: "EOF"},
    },
}
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/a44c87c3f5a0ea5a561f62cc7afaeac65dfc20b8

PR #901, opened 2024-04-24
[#900] Retry to detect OpenShift on failures

brusdev:


gtully (approved):
30seconds is an age, but we do want to know and the sleep is long enough to not spam the server. We may as well be trying to find out.
this seems sensible.

Issue #905, opened 2024-04-25
Adding watch resources owned by the operator


gaohoward:
**Describe the bug**

Currently the operator only watches it's statefulset and pod resources. Other resources like services, etc created by the operator should be added to the watch list too. 

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #906, opened 2024-04-25
[#905] Adding watch resources owned by the operator

gaohoward:


gtully (approved):
that looks correct and sensible to me, we track what we own.

gaohoward:
> that looks correct and sensible to me, we track what we own.

Thanks!

PR #914, opened 2024-04-29
ROSA suite test

brusdev:


gtully (approved):
everything is async and everything can fail and some things can be slow :-)

Issue #920, opened 2024-04-30
Deleting ActiveMQArtemisAddress CRs causes infinite reconcile loops


brusdev:
**Describe the bug**
Creating an ActiveMQArtemisAddress CR and deleting it causes infinite reconcile loops.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/57cdb4ebaae5e4c39d53b34fd190872aac4d8960

PR #921, opened 2024-04-30
[#920] Fix address controller infinite reconcile loops

brusdev:
Creating an ActiveMQArtemisAddress CR and deleting it causes infinite reconcile loops.


controllers/common_util_test.go
```diff
@@ -1108,3 +1108,123 @@ func (m *NillCluster) GetAPIReader() client.Reader {
 func (m *NillCluster) Start(ctx context.Context) error {
 	return nil
 }
+
+type dummyClient struct {
```
gtully, line 1112: the fakeClient may be an alternative, but possibly not as functional as your dummy.

gtully (approved):

controllers/common_util_test.go
brusdev: @gtully thanks for the suggestion, I didn't know that fakeClient can intercept calls. I removed the dummyClient in favor of the fakeClient.

PR #923, opened 2024-05-02
Federation pub sub

gtully:


gtully:
This will benefit from and really needs the support in https://issues.apache.org/jira/browse/ARTEMIS-4754, will leave as draft till 2.34.0 is released and referenced.

gtully:
no real need to wait for 2.34 to exercise what is currently there.

PR #933, opened 2024-05-22
[NO-ISSUE] test of jdbc ha

gtully:


controllers/activemqartemis_jdbc_ha_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+				// a clone of peerA
+				peerA.DeepCopyInto(peerB)
+				peerB.Name = peerPrefix + "-peer-b"
+
+				By("provisioning the broker peer-a")
+				Expect(k8sClient.Create(ctx, &peerA)).Should(Succeed())
+
+				By("provisioning the broker peer-b")
+				Expect(k8sClient.Create(ctx, peerB)).Should(Succeed())
```
brusdev, line 204: One of the peers crashes because of the readiness probe:
```
Readiness probe failed:
Back-off restarting failed container aa-jdbc-ha51a-peer-a-container in pod aa-jdbc-ha51a-peer-a-ss-0_test(2630511f-9729-4b6a-b477-0b636ab5a194)
```

controllers/activemqartemis_jdbc_ha_test.go
```diff
@@ -0,0 +1,283 @@
...
+				svc := &corev1.Service{
+					TypeMeta: metav1.TypeMeta{
+						APIVersion: "v1",
+						Kind:       "Service",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      brokerService,
+						Namespace: defaultNamespace,
+					},
+					Spec: corev1.ServiceSpec{
```
brusdev, line 243: publishNotReadyAddresses: true could speed up an eventual client failover

brusdev (commented):
It is a great idea of having 2 primary brokers contending for the same db lock like peers!

controllers/activemqartemis_jdbc_ha_test.go
gtully: exactly, we need to avoid that. we can only stay in the starting phase for max 300s so a startup probe wont cut it.
I think we need a liveness probe that knows about the waiting to lock sate and is happy to report live in that state and a readiness probe that gates the network access in the normal way.

I think we can build these over the broker logging of specific events, and can look at broker status via a jolokia agent at a later date.

controllers/activemqartemis_jdbc_ha_test.go
gtully: possibly, it may hinder normal operation too, we need to check the impact on the service if there are "not ready" pods in the mix.

controllers/activemqartemis_jdbc_ha_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+				By("Configuring probes to keep Pod alive while waiting to obtain shared jdbc lock")
+
+				peerA.Spec.DeploymentPlan.LivenessProbe = &corev1.Probe{
+					ProbeHandler: corev1.ProbeHandler{
+						Exec: &corev1.ExecAction{
+							Command: []string{
+								"/bin/bash", "-x", "-c",
+								// look for starting && !stopped
+								`LOGFILE=/home/jboss/amq-broker/log/startup_probe.log; grep "AMQ221000" $LOGFILE && ! grep "AMQ221002" $LOGFILE`,
```
gtully, line 236: this probe could be improved to also run the acceptor check once activated.

controllers/activemqartemis_jdbc_ha_test.go
```diff
@@ -0,0 +1,386 @@
...
+								"template": map[string]interface{}{
+									"spec": map[string]interface{}{
+										"initContainers": []interface{}{
+											map[string]interface{}{
+												"name":  "download-jdbc-driver",
+												"image": version.LatestInitImage,
+												"command": []interface{}{
+													"sh",
+													"-c",
+													"mkdir -p /amq/init/config/extra-libs && curl -o /amq/init/config/extra-libs/postgresql-42.7.1.jar https://jdbc.postgresql.org/download/postgresql-42.7.1.jar",
```
gtully, line 177: in production this would need to come from a container or from pre-populated volume mount. this needs some more investigation. we are not alone!

Issue #934, opened 2024-05-28
suppress deprecation warnings from the api server


gtully:
The operator must support deprecated CRs but on each call to the api server the expected deprecation warning is raised. There is a default de-dup log manager option but there is no point in tracking them at all as they are expected.
With the informer cache resync they can appear as spam!

gtully:
`W0515 13:15:29.638667 1 warnings.go:70] The ActiveMQArtemisSecurity CRD is deprecated. Use the spec.brokerProperties attribute in the ActiveMQArtemis CR and -jaas-config extraMount to configure security instead`

PR #935, opened 2024-05-28
[#934] suppress deprecated warnings from the api server as we know we…

gtully:
… need to work with deprecated crs

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,6 +30,7 @@ import (
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


gaohoward (approved):

gaohoward:
The downside of this is that there are no warnings at all....

gtully:
> The downside of this is that there are no warnings at all....

only in the operator logs, for the client that the operator uses to chat to the api server.

end users of kubectl will have their own client will see the warning in the normal way.

Also for the kube client in the test suite. That will still process the warnings as per the defaults.

main.go
```diff
@@ -184,6 +185,9 @@ func main() {
 		Logger:                 setupLog,
 	}
 
+	mgrOptions.Client.WarningHandler.SuppressWarnings = true
+	rest.SetDefaultWarningHandler(&rest.NoWarnings{})
```
brusdev, line 189: Could those warnings be logged with the debug level by creating a custom WarningHandler?

brusdev (commented):
The WarningHandler is a great solution!

main.go
gtully: They could, and maybe in the future there are other warnings that may be all we get. let me change that to debug logging.

gtully:
- --zap-log-level=2

`2024-05-29T09:52:42.665Z LEVEL(-2) The ActiveMQArtemisAddress CRD is deprecated. Use the spec.brokerProperties attribute in the ActiveMQArtemis CR to create addresses and queues instead {"code": 299, "agent": "-"}
`

gtully:
> > The downside of this is that there are no warnings at all....
> 
> only in the operator logs, for the client that the operator uses to chat to the api server.
> 
> end users of kubectl will have their own client will see the warning in the normal way.
> 
> Also for the kube client in the test suite. That will still process the warnings as per the defaults.

it makes sense to log at level 2, thanks for the review

brusdev (approved):

PR #946, opened 2024-06-11
Fix trust secret

brusdev:


gtully (approved):
seems sensible to require a .pem format as the other supported keys are already locked down.

gtully:
btw, nice use of the reason string to capture the error in the status

pkg/utils/certutil/certutil.go
```diff
@@ -160,14 +160,17 @@ func isSecretFromBundle(secret *corev1.Secret) bool {
...
-	//extract the key of the secret's only entry
-	bundleName := ""
+func getBundleNameFromSecret(secret *corev1.Secret) (string, error) {
+	//extract the bundle target secret key that ends with .pem
+	//the bundle target secret could include keys for additional formats jks/pkcs12
 	for key := range secret.Data {
-		bundleName = key
-		break
+		//the bundle target secret key must ends with .pem
+		if strings.HasSuffix(key, ".pem") {
```
gaohoward, line 168: Maybe add a note in the doc?

gaohoward (approved):

pkg/utils/certutil/certutil.go
brusdev: I added a note.

Issue #945, opened 2024-06-11
Trust secret doesn't support bundle with additional format certificates


brusdev:
**Describe the bug**
When the trust secret has additional format certificates, the broker pods are restarted frequently, i.e.

```
apiVersion: trust.cert-manager.io/v1alpha1
kind: Bundle
metadata:
  name: my-bundle
spec:
  sources:
  - secret:
      name: "my-foo-cert-secret"
      key: "tls.crt"
  - secret:
      name: "my-bar-cert-secret"
      key: "tls.crt"
  target:
    secret:
      key: "root-certs.pem"
    additionalFormats:
      jks:
        key: "bundle.jks"
      pkcs12:
        key: "bundle.p12"
```

```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata: 
  name: broker
spec:
  acceptors:
    - name: tls-acceptor
      port: 61617
      sslEnabled: true
      needClientAuth: true
      sslSecret: my-broker-cert-secret
      trustSecret: my-bundle
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/416f9dbda97a4d9951c0f956b8186c9cfa1638a3

PR #952, opened 2024-06-13
[#945] fix up typo in test check for bundle truststore PEMCA

gtully:


brusdev:
@gtully good catch!!!

Issue #956, opened 2024-06-14
Broker properties from a secret key with the prefix broker- are not loaded


brusdev:
**Describe the bug**
The broker properties from a secret key with the prefix `broker-` are not loaded, i.e.
```
kind: Secret
apiVersion: v1
metadata:
  name: my-secret-bp
data:
  broker-my-address.properties: >-  
    addressesSettings.my-address.managementMessageAttributeSizeLimit=-1
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/dd7e466d0ec985831f3a976d1d1055a172cc4cb0

PR #957, opened 2024-06-14
[#956] Fix broker property with ordinal parser

brusdev:


controllers/activemqartemis_reconciler.go
```diff
@@ -2468,40 +2468,32 @@ func alder32Of(props []string) []byte {
 }
 
 func brokerPropertiesData(props []string) map[string]string {
```
gtully, line 2470: this is great, a nice simplification and a good fix.
only comment is I would like a test of this function too, the user of the newly minted and simpler code.

gtully (approved):

controllers/activemqartemis_reconciler.go
brusdev: Sure, I added 3 unit tests for the function BrokerPropertiesData

gtully (approved):
:-)

PR #963, opened 2024-06-19
[#962] Operator cannot access broker status when console is secured with pem cert

### LINKED ISSUE #962: Operator cannot access broker status when console is secured with pem cert
**Describe the bug**
If the console is sslEnabled with a certificate from cert-manager (i.e. pem format cert) the operator cannot update the status and the Conditions shows UnableToRetrieveStatus reason. for example
```
  - lastTransitionTime: "2024-06-19T04:02:30Z"
    message: 'Get "http://HMwwfpiz:***@artemis-broker-ss-0.artemis-broker-hdls-svc.default.svc.cluster.local:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22amq-broker%22/Status":
      net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x03\x00\x02\x02P"'
    reason: UnableToRetrieveStatus
    status: Unknown
    type: BrokerPropertiesApplied
  - lastTransitionTime: "2024-06-19T04:02:41Z"
    message: 'Get "http://HMwwfpiz:***@artemis-broker-ss-0.artemis-broker-hdls-svc.default.svc.cluster.local:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22amq-broker%22/Status":
      net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x03\x00\x02\x02P"'
    reason: UnableToRetrieveStatus
    status: Unknown
    type: BrokerVersionAligned
```
The reason is that when using pem cert the jolokia couldn't get correct protocols and uses 'http' protocol. It should use 'https' protocol.

> [!TIP]

gaohoward:


brusdev (commented):
@howardgao your fix LGTM but depending on the env vars that users now can customize seems risky. It is fine only if we consider it a temporary fix waiting for the new control-plane endpoint on which @gtully  is working.

pkg/utils/jolokia_client/jolokia_client.go
```diff
@@ -139,12 +142,14 @@ func resolveJolokiaRequestParams(namespace string,
...
-			if !userDefined && oneVar.Name == "AMQ_PASSWORD" {
+			} else if !userDefined && oneVar.Name == "AMQ_PASSWORD" {
 				jolokiaPassword = getEnvVarValue(&oneVar, &podNamespacedName, client, nil)
-			}
-			if oneVar.Name == "AMQ_CONSOLE_ARGS" {
+			} else if oneVar.Name == "AMQ_CONSOLE_ARGS" {
 				jolokiaProtocol = getEnvVarValue(&oneVar, &podNamespacedName, client, nil)
+			} else if oneVar.Name == "JAVA_ARGS_APPEND" {
+				jolokiaProtocol, port = getProtocolFromJavaArgs(oneVar.Value)
+			} else if oneVar.Name == "AMQ_NAME" {
```
gtully, line 151: same question, name from an env var, why?

gtully (commented):
I may be missing something, but it the root of the problem that the jolokia client does not use https, so the just the protocol determination needs to be better.
Do we really want the choice of port to be an option, and I guess by extension the name.
The less options the better, and in a container there is already isolation so less reason to use custom ports.

controllers/activemqartemis_controller_cert_manager_test.go
```diff
@@ -1119,3 +1124,42 @@ func testConfiguredWithCertAndBundle(certSecret string, caSecret string) {
...
+	brokerCr, createdBrokerCr := DeployCustomBroker(defaultNamespace, func(candidate *brokerv1beta1.ActiveMQArtemis) {
+
+		candidate.Name = brokerCrName
+
+		candidate.Spec.DeploymentPlan.Size = common.Int32ToPtr(1)
+		candidate.Spec.DeploymentPlan.ReadinessProbe = &corev1.Probe{
+			InitialDelaySeconds: 1,
+			PeriodSeconds:       1,
+			TimeoutSeconds:      5,
+		}
```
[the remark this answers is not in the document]
gaohoward: I copied from another test. I think it can speed up the pod status.

pkg/utils/artemis/artemis.go
```diff
@@ -66,6 +66,14 @@ func GetArtemis(_ip string, _jolokiaPort string, _name string, _user string, _pa
 	return &artemis
 }
 
+func (artemis *Artemis) GetJolokia() jolokia.IJolokia {
+	return artemis.jolokia
+}
+
+func (artemis *Artemis) GetName() string {
+	return artemis.name
```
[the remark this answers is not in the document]
gaohoward: I added this method purely for test purposes. The usage of name is not changed.

pkg/utils/jolokia_client/jolokia_client.go
gaohoward: This is where the name come from. So I think make the name not hardcoded just in case people changes it.

gaohoward: The port and name added here don't add any new behaviors it's just for this particular case, neither it'll break existing ones. I'll remove them because they are not part of fixing the main issue.

gaohoward:
> really want the choice of port to be an option



> @howardgao your fix LGTM but depending on the env vars that users now can customize seems risky. It is fine only if we consider it a temporary fix waiting for the new control-plane endpoint on which @gtully is working.

yeah I'm aware of that. The whole thing seems broken but until the new solution is out we don't want to patch things that are doomed to be replaced. :)

gtully (approved):
looks good and it is not something we can patch.

We will lock this down to https soon, at least for a secure deployment

PR #966, opened 2024-06-21
[NO-ISSUE] Fix Expect without assertion in suite_test.go

gaohoward:
@brusdev Those are typos right?

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -217,7 +217,7 @@ func setUpNamespace() {
 	}
 
 	err := k8sClient.Create(ctx, &testNamespace)
-	Expect(err == nil || errors.IsConflict(err))
+	Expect(err == nil || errors.IsConflict(err)).To(BeTrue())
 
 	if isOpenshift {
 		Eventually(func(g Gomega) {
@@ -317,7 +317,7 @@ func setUpTestProxy() {
 	}
 
 	err = k8sClient.Create(ctx, &testProxyDeployment)
-	Expect(err != nil || errors.IsConflict(err))
+	Expect(err == nil || errors.IsConflict(err)).To(BeTrue())
 
 	testProxyService := corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
@@ -336,7 +336,7 @@ func setUpTestProxy() {
 	}
 
 	err = k8sClient.Create(ctx, &testProxyService)
-	Expect(err != nil || errors.IsConflict(err))
+	Expect(err == nil || errors.IsConflict(err)).To(BeTrue())
 
 	testProxyIngress := ingresses.NewIngressForCRWithSSL(
 		nil, types.NamespacedName{Name: testProxyName, Namespace: testProxyNamespace},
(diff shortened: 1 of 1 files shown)


controllers/suite_test.go
```diff
@@ -217,7 +217,7 @@ func setUpNamespace() {
 	}
 
 	err := k8sClient.Create(ctx, &testNamespace)
-	Expect(err == nil || errors.IsConflict(err))
+	Expect(err).NotTo(HaveOccurred())
```
brusdev, line 220: This change doesn't LGTM because the test-suite should be able to be executed on an existing test namespace.

controllers/suite_test.go
```diff
@@ -317,7 +317,7 @@ func setUpTestProxy() {
 	}
 
 	err = k8sClient.Create(ctx, &testProxyDeployment)
-	Expect(err != nil || errors.IsConflict(err))
+	Expect(err).NotTo(HaveOccurred())
```
brusdev, line 320: You are right, this seems wrong, but the conflict errors should be ignored:
```suggestion
	Expect(err == nil || errors.IsConflict(err))
```

brusdev (commented):
Good catch! Those are typos but in same cases we need to ignore errors and `Expect(err).NotTo(HaveOccurred())` is not sufficient to ignore errors.

controllers/suite_test.go
gaohoward: +1

controllers/suite_test.go
gaohoward: +1 (I shall add `.To(BeTrue())` assertion)

controllers/suite_test.go
```diff
@@ -336,15 +336,15 @@ func setUpTestProxy() {
 	}
 
 	err = k8sClient.Create(ctx, &testProxyService)
-	Expect(err != nil || errors.IsConflict(err))
+	Expect(err).NotTo(HaveOccurred())
```
brusdev, line 339: The test-suite can skip error if the test proxy service already exist:
```suggestion
	Expect(err == nil || errors.IsConflict(err))
```

controllers/suite_test.go
```diff
@@ -336,15 +336,15 @@ func setUpTestProxy() {
...
 
 	testProxyIngress := ingresses.NewIngressForCRWithSSL(
 		nil, types.NamespacedName{Name: testProxyName, Namespace: testProxyNamespace},
 		map[string]string{}, testProxyName+"-dep-svc", strconv.FormatInt(int64(testProxyPort), 10),
 		true, "", testProxyHost, isOpenshift)
 
 	err = k8sClient.Create(ctx, testProxyIngress)
-	Expect(err != nil || errors.IsConflict(err))
+	Expect(err).NotTo(HaveOccurred())
```
brusdev, line 347: The test-suite can skip error if the test proxy ingress already exist:
```suggestion
	Expect(err == nil || errors.IsConflict(err))
```

controllers/suite_test.go
brusdev: Yes, all those occurrences are missing the assertion.

brusdev (approved):
@howardgao nice work! Thanks for fixing it.

Issue #968, opened 2024-07-01
Replace HTTP test proxy with SSH tunnels


brusdev:
**Describe the feature**
SSH tunnels support HTTPS and other tcp connections.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/bcd2351ad2260f22a843cadea89a66e1825bf5e6

PR #969, opened 2024-07-01
[#968] Replace HTTP test proxy with SSH tunnels

brusdev:
SSH tunnels support HTTPS and other tcp connections.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -31,9 +31,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/crypto/ssh"
 
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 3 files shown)


gaohoward:
@brusdev  I just found that when 'test' namespace already exists, this line panicks:

https://github.com/artemiscloud/activemq-artemis-operator/blob/0babfdc623fc29ce8b721331756ff21b6baf1543/controllers/suite_test.go#L223

I think we should use 
`Expect(err == nil || errors.IsAlreadyExists(err)).To(BeTrue())`

brusdev:
> @brusdev I just found that when 'test' namespace already exists, this line fails:
> 
> https://github.com/artemiscloud/activemq-artemis-operator/blob/0babfdc623fc29ce8b721331756ff21b6baf1543/controllers/suite_test.go#L223
> 
> I think we should use `Expect(err == nil || errors.IsAlreadyExists(err)).To(BeTrue())`

@gaohoward I have just fixed it

gtully (approved):
great to have the https path sorted, looks great.

gaohoward (approved):
I give my approval as it works. I'm still having issues but I think that's specific to my special environment. I'll investigate afterwards.

gaohoward (approved):

Issue #976, opened 2024-07-09
Add an example to monitor broker metrics by using Prometheus


brusdev:
**Describe the feature**
The operator documentations already include a section that describes how to monitor broker metrics by using Prometheus. Can you add a yaml file in the examples folder?

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resloved by https://github.com/artemiscloud/activemq-artemis-operator/commit/f4f5decea4cdb3a8a6f50212bb2b6e5568f4879c

PR #977, opened 2024-07-09
[#976] Add an example to monitor broker metrics by using Prometheus

brusdev:


brusdev:
I updated the example to use the port `console-jolokia` that exists even if the console is not exposed.

gtully:
> I updated the example to use the port `console-jolokia` that exists even if the console is not exposed.

good, there is also just plain 'jolokia' for the agent use case, so glad that is still available.

gtully (approved):

PR #980, opened 2024-07-10
[#979] Configure drainer pod to expose it's IP address

### LINKED ISSUE #979: Configure drainer pod to expose it's IP address
**Describe the bug**
When creating the drainer pod for scaledown, the drainer.sh uses its hostname in its connector which is not resolvable.
(https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/issues/105)
Using pod's IP can be a simple solution in this case. The operator should configure the pod template so that the drainer script
in side the pod container could obtain it's own IP address (e.g. use an env var)

> [!TIP]

gaohoward:


gaohoward:
Set it as draft as it depends on https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/pull/106

gaohoward:
The test is supposed to fail until [this PR](https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/pull/106) is merged.

gtully:
I find it difficult to reason about this change in the context that the pod wants to be a client of the cluster but is also not part of the cluster. The IP will probably work, but that is transient. I think if the first attempt to scaledown fails it will cause another similar problem as there will be two different IPs in play. 
I wonder if we should take the first step to a different approach and keep the scaling down pod alive? All of the other logic can remain similar, with the jmx call etc, but at least the cluster logic will be preserved. What we may need to do in addition is stop/pause the acceptors (or add in a connection router that rejects all)

This may be a step to far, but it may also be a good step in the right direction.

gaohoward:
> I find it difficult to reason about this change in the context that the pod wants to be a client of the cluster but is also not part of the cluster. The IP will probably work, but that is transient. I think if the first attempt to scaledown fails it will cause another similar problem as there will be two different IPs in play. I wonder if we should take the first step to a different approach and keep the scaling down pod alive? All of the other logic can remain similar, with the jmx call etc, but at least the cluster logic will be preserved. What we may need to do in addition is stop/pause the acceptors (or add in a connection router that rejects all)
> 
> This may be a step to far, but it may also be a good step in the right direction.

As we discussed that we can improve this in the next release as part of the effort to refactor the whole scale down mechanism. For the the IP solution should work as expected and is a valid fix for the current implementation.

gaohoward:
Until next release to bring in fix on the broker container image, this PR should not be merged as the test will fail.

brusdev:
@gaohoward I updated the related images, see https://github.com/artemiscloud/activemq-artemis-operator/commit/58bcb0b3c714a750500a13a1360b7d77a1b267a5

Can you rebase your branch?

gaohoward:
> @gaohoward I updated the related images, see [58bcb0b](https://github.com/artemiscloud/activemq-artemis-operator/commit/58bcb0b3c714a750500a13a1360b7d77a1b267a5)
> 
> Can you rebase your branch?

will do.

brusdev:
@howardgao the test `Scale down controller scaledown to the right pod` failed 3 times in a row on my laptop, could you double-check?
```
  [FAILED] Timed out after 180.001s.
  The function passed to Eventually failed at /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemisscaledown_controller_test.go:388 with:
  Expected
      <string>: {"request":{"mbean":"org.apache.activemq.artemis:address=\"DLQ\",broker=\"amq-broker\",component=addresses","attribute":"MessageCount","type":"read"},"value":0,"timestamp":1721729906,"status":200}
  to contain substring
      <string>: "value":2
  In [It] at: /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemisscaledown_controller_test.go:390
```

gaohoward:
> @howardgao the test `Scale down controller scaledown to the right pod` failed 3 times in a row on my laptop, could you double-check?
> 
> ```
>   [FAILED] Timed out after 180.001s.
>   The function passed to Eventually failed at /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemisscaledown_controller_test.go:388 with:
>   Expected
>       <string>: {"request":{"mbean":"org.apache.activemq.artemis:address=\"DLQ\",broker=\"amq-broker\",component=addresses","attribute":"MessageCount","type":"read"},"value":0,"timestamp":1721729906,"status":200}
>   to contain substring
>       <string>: "value":2
>   In [It] at: /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/activemqartemisscaledown_controller_test.go:390
> ```

will take a look. Thanks

gaohoward:
Hi Dom, regarding the test failure I think we should revert the test  to align with this revert 
https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/commit/4b3d68c8631bbf5340c9614a416303e8e350a9a3
Without this commit the test would fail for sure.  
I'll revert the test in another PR if you agree.

brusdev:
> Hi Dom, regarding the test failure I think we should revert the test to align with this revert [artemiscloud/activemq-artemis-broker-kubernetes-image@4b3d68c](https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/commit/4b3d68c8631bbf5340c9614a416303e8e350a9a3) Without this commit the test would fail for sure. I'll revert the test in another PR if you agree.

@howardgao alternatively, you could push a separate commit to revert that test in this same branch you are using for this PR. In this way we can validate and merge both changes at the same time.

gaohoward:
> > Hi Dom, regarding the test failure I think we should revert the test to align with this revert [artemiscloud/activemq-artemis-broker-kubernetes-image@4b3d68c](https://github.com/artemiscloud/activemq-artemis-broker-kubernetes-image/commit/4b3d68c8631bbf5340c9614a416303e8e350a9a3) Without this commit the test would fail for sure. I'll revert the test in another PR if you agree.
> 
> @howardgao alternatively, you could push a separate commit to revert that test in this same branch you are using for this PR. In this way we can validate and merge both changes at the same time.

@brusdev Test reverted.

brusdev (approved):
Full test-suite passed.

gaohoward:
> Full test-suite passed.

Thank you Dom

PR #982, opened 2024-07-12
[#981] Avoid using JAVA_ARGS_APPEND for internal configuration

### LINKED ISSUE #981: CRD env JAVA_ARGS_APPEND fromSource needs to be flagged as invalid
**Describe the bug**
The operator currently uses this env var to configure a broker (such as defining custom logging file url)
If use adds the JAVA_ARGS_APPEND in to the spec.env using `ValueFrom` (like from a secret) it will get overridden by the operator. 

> [!TIP]

gaohoward:
This PR add tests on how to properly append to JAVA_ARGS_APPEND env var

### DIFF
--- a/controllers/activemqartemis_controller.go
+++ b/controllers/activemqartemis_controller.go
@@ -263,6 +263,13 @@ func (r *ActiveMQArtemisReconcilerImpl) validate(customResource *brokerv1beta1.A
 		}
 	}
 
+	if validationCondition.Status != metav1.ConditionFalse {
+		condition, retry = r.validateEnvVars(customResource)
+		if condition != nil {
+			validationCondition = *condition
+		}
+	}
+
 	validationCondition.ObservedGeneration = customResource.Generation
 	meta.SetStatusCondition(&customResource.Status.Conditions, validationCondition)
 
@@ -402,6 +409,35 @@ func (r *ActiveMQArtemisReconcilerImpl) validateExposeModes(customResource *brok
 		}, false
 	}
 
+	return nil, false
+}
+
+func (r *ActiveMQArtemisReconcilerImpl) validateEnvVars(customResource *brokerv1beta1.ActiveMQArtemis) (*metav1.Condition, bool) {
+
+	internalVarNames := map[string]string{
+		debugArgsEnvVarName:      debugArgsEnvVarName,
+		javaOptsEnvVarName:       javaOptsEnvVarName,
+		javaArgsAppendEnvVarName: javaArgsAppendEnvVarName,
+	}
+
+	invalidVars := []string{}
+
+	for _, envVar := range customResource.Spec.Env {
+		if _, ok := internalVarNames[envVar.Name]; ok {
+			if envVar.ValueFrom != nil {
+				invalidVars = append(invalidVars, envVar.Name)
+			}
+		}
+	}
+
+	if len(invalidVars) > 0 {
+		return &metav1.Condition{
+			Type:    brokerv1beta1.ValidConditionType,
+			Status:  metav1.ConditionFalse,
+			Reason:  brokerv1beta1.ValidConditionInvalidInternalVarUsage,
+			Message: fmt.Sprintf("Don't use valueFrom on env vars that the operator can mutate: %v. Instead use a different var and refernece it in its value field.", invalidVars),
+		}, false
+	}
 	return nil, false
 }
 
--- a/docs/help/operator.md
+++ b/docs/help/operator.md
@@ -818,8 +818,34 @@ spec:
       value: -XshowSettings:system
 
 ```
+---
+**NOTE**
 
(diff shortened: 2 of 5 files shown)


gaohoward:
(Test is supposed to fail at the moment until the dependent PR is merged)

gtully:
In some cases we want to split, but in general the intent is that we use JDK_JAVA_ARGS and we append to any existing value, and JAVA_ARGS_APPEND is a way to override as a duplicate command line argument. They are env vars that are at either end of the java command line.

The user should always have control to override *if* necessary, it is like a build in escape hatch.

But the append depends on env values. However we can combine, b/c env vars can be dependent.

If the user wants an env var from source, that can be separate and included in the env vars we care about.

I tried a little test of this:

```
crd.Spec.Env = []corev1.EnvVar{
				{
					Name: "ENV_FROM_X",
					ValueFrom: &corev1.EnvVarSource{
						SecretKeyRef: &corev1.SecretKeySelector{
							LocalObjectReference: corev1.LocalObjectReference{Name: secret.Name},
							Key:                  "envKey",
						},
					}},
				{Name: "JAVA_ARGS_APPEND", Value: "-Dhawtio.realm=console $(ENV_FROM_X)"},
			}
```

and it works as expected, $(ENV_FROM_X) gets expanded into the value of the envKey from the secret just fine.

gaohoward:
> In some cases we want to split, but in general the intent is that we use JDK_JAVA_ARGS and we append to any existing value, and JAVA_ARGS_APPEND is a way to override as a duplicate command line argument. They are env vars that are at either end of the java command line.
> 
> The user should always have control to override _if_ necessary, it is like a build in escape hatch.
> 
> But the append depends on env values. However we can combine, b/c env vars can be dependent.
> 
> If the user wants an env var from source, that can be separate and included in the env vars we care about.
> 
> I tried a little test of this:
> 
> ```
> crd.Spec.Env = []corev1.EnvVar{
> 				{
> 					Name: "ENV_FROM_X",
> 					ValueFrom: &corev1.EnvVarSource{
> 						SecretKeyRef: &corev1.SecretKeySelector{
> 							LocalObjectReference: corev1.LocalObjectReference{Name: secret.Name},
> 							Key:                  "envKey",
> 						},
> 					}},
> 				{Name: "JAVA_ARGS_APPEND", Value: "-Dhawtio.realm=console $(ENV_FROM_X)"},
> 			}
> ```
> 
> and it works as expected, $(ENV_FROM_X) gets expanded into the value of the envKey from the secret just fine.

That makes sense. Thanks!

gaohoward:
@gtully test updated. please review. Thanks!

gtully (commented):
I think this is the right approach.

Did you consider doing this check at the validation phase?
I imagine if valueFrom and value are defined that creating the stateful set will fail/error which is very similar. ie: a deployment failure.
If we identify the env vars we modify, in some internal collection - we already use createorappend for those in the code, then we can trap a validation error if any env var matches that set and has valueFrom set.
if that set ever gets out of sync, the stateful set error will be a fallback.
I think it makes sense for us to document or callout the env vars we modify or want to  modify. The set should be small.
in the error message we should point to using dependent variables as the desired approach. It can be very specific, and can be very specific, in a validation failure condition/error.

gaohoward:
@gtully PR updated. Now
- check valueFrom at validation phase
- document internal used env vars

controllers/activemqartemis_controller.go
```diff
@@ -405,6 +412,36 @@ func (r *ActiveMQArtemisReconcilerImpl) validateExposeModes(customResource *brok
...
+			}
+		}
+	}
+
+	if len(invalidVars) > 0 {
+		return &metav1.Condition{
+			Type:    brokerv1beta1.ValidConditionType,
+			Status:  metav1.ConditionFalse,
+			Reason:  brokerv1beta1.ValidConditionInvalidInternalVarUsage,
+			Message: fmt.Sprintf("Using valueFrom with internal env vars %v will get ignored. Instead use a different var and refernece it in its value field.", invalidVars),
```
gtully, line 438: Maybe just. Don't use valueFrom on env vars that the operator can mutate.

... and reference as a dependent var....

gtully: The ignored but seems wrong and the kube docs call out dependent vars

docs/help/operator.md
```diff
@@ -818,8 +818,34 @@ spec:
...
+
+There are a few well-known environment variables that are used by the operator internally to configure brokers, as shown below
+
+* JAVA_ARGS_APPEND
+* JAVA_OPTS
+* DEBUG_ARGS
+
+If you want to add some values to any of them, make sure you define it in **spec.env** using `value` field. If you use `valueFrom` for the env var you will get validation error condition in your CR's status.
+
+If you really need `valueFrom` to define the values for the above env vars you can use a different var and then reference it in the internal var's `value` field.
```
gtully, line 834: this is great. Possible add a reference to the interdependent env vars section of the kube to add further clarity

https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/

controllers/activemqartemis_controller.go
gaohoward: Updated.

gtully (approved):

brusdev (approved):

PR #985, opened 2024-07-18
[#984] Do not treat tls secrets as PEM type if it is intended to be used as old sslSecret

### LINKED ISSUE #984: Do not treat tls secrets as PEM type if it is intended to be used as old sslSecret
**Describe the bug**
In cases that the user configured a tls secret (come from cert-manager) as an old style ssl Secret, the operator should treat it like so (current it just treated it as a pem secret and configure ssl using PEM cert store).

> [!TIP]

gaohoward:


gtully (approved):
looks good, we  give precedence to annotated secrets that have password credentials.

brusdev (approved):

PR #991, opened 2024-07-26
[NO-ISSUE] Test whitespaced security roles

brusdev:


gaohoward (approved):

gaohoward:
Test looking good. However I wouldn't recommend people to use those special chars in their applications. It's just confusing themselves. :)

brusdev:
@gaohoward thanks for your review. I also wouldn't recommend people to use those special chars in their applications but some users are already using them and the purpose of this test is only to validate that they work and avoid regressions.

Issue #1008, opened 2024-09-13
Log ignored secrets


brusdev:
**Describe the feature**
The secrets with credentials, acceptors/connectors and console args are ignored if their owner references don't include one for the the ActiveMQArtemis CR. This behavior is expected because it allows users to edit them but it is hard to troubleshoot without a line in the operator log.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/f0f2a8c3b6961b65aec0a901b351a0243dff8110

PR #1009, opened 2024-09-13
[#1008] Log ignored secrets

brusdev:


gtully (approved):
that should help diagnose in the future

Issue #1007, opened 2024-09-13
persistence=true PVC claim reconcile loop due to  PersistentVolumeClaim.Spec.VolumeMode implicit default


gtully:
There is a repeated SS reconcile for CR.persistence=true as the PVC Spec template does not respect the current deployed spec, and blindly overwrites.

PR #1010, opened 2024-09-13
[#1007] respect deployed resource on PVC reconcile and fill in volume…

gtully:
…Mode default


gtully:
Worth some thought is the ObjectMeta in the PVC template, not sure it is  good idea.

brusdev (approved):
Using one function for both the default PV and the ExtraVolumeClaimTemplates is a great idea!

brusdev:
Full test-suite passed

gtully:
> Worth some thought is the ObjectMeta in the PVC template, not sure it is good idea.

I am going to leave it, we don't add the same default labels as we do for the persistenceEnabled PVC, so there possibly will be the need to add labels. A resourceTemplate will work in this case also. Maybe it is a good convenience.

Issue #1012, opened 2024-09-17
Exposing console on microshift doesn't work


brusdev:
**Describe the bug**
When trying deploying a broker with expose console on microshift default installtion the the broker instance is not created. 

```
apiVersion: broker.amq.io/v1beta1
kind: ActiveMQArtemis
metadata:
  name: ex-aao
spec:
  console:
    expose: true
```

Looking in the activemqartemis object, the status say:
```
  message: .Spec.Console has invalid ingress settings, IngressHost unspecified and no Spec.IngressDomain default domain provided
  observedGeneration: 1
  reason: InvalidIngressSettings
  status: "False"
  type: Valid
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/4013e34e601c82cb94ea534b5b1454578292cc6f

PR #1015, opened 2024-09-18
[#1012] Fix MicroShift detection

brusdev:


gtully (commented):
Looks good otherwise. Makes sense to just check for route

brusdev:
> Typo in the commit message!

Good catch! I fixed the commit message.

Issue #1031, opened 2024-10-14
List out verbs explicitly in RBAC definitions


brusdev:
**Describe the feature**
Instead of using the wildcard character in RBAC definitions, the best practice is to explicitly list out each verb or resource.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/0a135d5ca1cd6b7d6c8ed2e110bce26d2b6b8cdb

PR #1032, opened 2024-10-14
[#1031] List out verbs explicitly in RBAC definitions

brusdev:


brusdev:
Full test-suite passed

gtully (commented):
that is indeed better.

Issue #1036, opened 2024-10-18
Enable useClientAuth on console breaks status checks from the operator


doxsch:
We would like to protect the console using client certificates. To do this, we have enabled useClientAuth in the console configuration. 

Configuration:
````
...
console:                                                                                  
  expose: true                                                                            
  exposeMode: ingress                                                                     
  name: console                                                                           
  sslEnabled: true                                                                        
  sslSecret: my-mapping-ssl-cert                                         
  trustSecret: my-mapping-ssl-cert                                       
  useClientAuth: true 
...
````

This has worked so far. The JaaS TextFileCertificateLoginModule maps the certificate provided by the client to a user and group.

However, once this is activated, the following errors appear in the status field of the ActiveMQArtemis resource:

```
...
- lastTransitionTime: "2024-10-17T10:48:03Z"
  message: 'Get "https://aQ0IvDAb:***@my-broker-ss-0.my-broker-hdls-svc.my-namespace.svc.cluster.local:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22amq-broker%22/Status": remote error: tls: bad certificate'
  reason: UnableToRetrieveStatus
  status: Unknown
  type: BrokerPropertiesApplied
- lastTransitionTime: "2024-10-17T10:48:32Z"
  message: 'Get "https://aQ0IvDAb:***@my-broker-ss-0.my-broker-hdls-svc.my-namespace.svc.cluster.local:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22amq-broker%22/Status": remote error: tls: bad certificate'
  reason: UnableToRetrieveStatus
  status: Unknown
  type: BrokerVersionAligned
- lastTransitionTime: "2024-10-16T06:30:50Z"
  message: 'Get "https://aQ0IvDAb:***@my-broker-ss-0.my-broker-hdls-svc.my-namespace.svc.cluster.local:8161/console/jolokia/read/org.apache.activemq.artemis:broker=%22amq-broker%22/Status": remote error: tls: bad certificate'
  reason: UnableToRetrieveStatus
  status: Unknown
  type: JaasPropertiesApplied
...
```

I think the Jolokia client itself should also provide a client certificate. At the moment, there seems to be no way to do this?

gtully:
[...]

Issue #1017, opened 2024-09-18
support for read only root file-system


brusdev:
### Discussed in https://github.com/artemiscloud/activemq-artemis-operator/discussions/1014

<div type='discussions-op-text'>

<sup>Originally posted by **robert-mcnamara** September 17, 2024</sup>
Hi Artemis Cloud maintainers,

We use the artemis-operator to deploy Artemis on EKS.

Our EKS platform team is enforcing a strict security requirements where workloads must have a read-only root file-system. This applies to init containers and containers.

We have tested out the following configuration when creating Artemis instances against our desired use-cases and so far testing has been positive.

```
extraVolumeMounts:
- mountPath: /home/jboss/amq-broker
  name: amq-jboss-home
- mountPath: /opt/jboss/container/jolokia/etc
  name: jolokia-configuration
- mountPath: /tmp
  name: temp-dir
extraVolumes:
- emptyDir: {}
  name: amq-jboss-home
- emptyDir: {}
  name: jolokia-configuration
- emptyDir: {}
  name: temp-dir
```

These changes have enabled the init container and main broker container to start and function with `readOnlyRootFilesystem` set to `true`, by using volume of type `emptyDir` mounted on the paths the Artemis init-container is writing to.

We would like to discuss how this could be supported 'natively' without a user have to apply a change like above.

Perhaps something like;
If readOnlyRootFilesystem is enabled, the operator would add the emptyDir volumes and mount for a tmp and 'run' directory, and the Artemis images could be updated to install into or use the tmp and run directory as needed.
Your input on this would be much appreciated.

Note: we recently experienced similar while exploring IBM MQ on EKS and found a newer version of IBM MQ approached this in a similar way
Ref: https://github.com/ibm-messaging/mq-container/blob/master/docs/usage.md#running-with-a-read-only-root-filesystem</div>

gtully:
with the current release, this is a good use of extraVolumes.
 
I am working on a restricted deployment flag, where the broker is very much locked down, readOnlyRootFilesystem would be a default in that scenario. 
I think 'native support' would make sense in that context. There is a POC in progress at https://github.com/gtully/activemq-artemis-operator/tree/control_plane that I will flesh out a bit more by the end of next week.

gtully:
Am going to resolve this in the context of the restricted flag.

PR #1037, opened 2024-10-18
support restricted option, jolokia agent, rbac, read only root, minimal

gtully:


gtully:
this pr includes support for:
 - the jolokia client using mtls #1036 
 - read only root file system #1017

gtully:
I have remove the use of JDK_JAVA_ARGS in restricted mode, possibly, to avoid command line limitations, we may have to push these into a cmd file in the future, but that could also be a secret, as an easy way to supply different values.

brusdev (commented):
Great work!!! It provides a predictable and secure deployment.

controllers/activemqartemis_reconciler.go
```diff
@@ -301,6 +303,10 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessStatefulSet(customResour
 }
 
 func isClustered(customResource *brokerv1beta1.ActiveMQArtemis) bool {
+	if isRestricted(customResource) {
```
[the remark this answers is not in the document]
gtully: I think that will have to be documented or will flow from things like existing RBAC roles and the ability to work without passwords. I think it will be easier to demonstrate/document the subset that works through examples and tests. In essence what is left will be configured through brokerProperties, as a way to open up the broker.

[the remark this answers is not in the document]
gtully: I added some doc to mark this feature experimental and provide some high level detail on what it does. By experimenting we will be able to understand the ways in which this mode of operation needs to be extended.

doxsch:
> * the jolokia client using mtls

If I understand the code correctly the mtls jolokia client will only be available in restricted mode. But I think if `spec.console.useClientAuth` is true, the controller should also use the mtls jolokia client otherwise the issue https://github.com/artemiscloud/activemq-artemis-operator/issues/1036 won't be fixed.

pkg/utils/jolokia/jolokia.go
```diff
@@ -94,6 +113,16 @@ func (j *Jolokia) getClient() *http.Client {
...
 		}
+		if j.restricted {
+			httpClientTransport.TLSClientConfig.InsecureSkipVerify = false
+			httpClientTransport.TLSClientConfig.GetClientCertificate =
+				func(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+					return common.GetOperatorClientCertificate(j.client, cri)
+				}
+		}
+		if rootCas, err := common.GetRootCAs(j.client); err == nil {
+			httpClientTransport.TLSClientConfig.RootCAs = rootCas
```
brusdev, line 124: Could the jolokia client only trust the operand certificate?

gtully:
> > * the jolokia client using mtls
> 
> If I understand the code correctly the mtls jolokia client will only be available in restricted mode. But I think if `spec.console.useClientAuth` is true, the controller should also use the mtls jolokia client otherwise the issue #1036 won't be fixed.

yes. I think it will be easy to fix #1036 with these changes. But #1036 is not resolved by this change

pkg/utils/jolokia/jolokia.go
gtully: I think that depends on the content of the trust bundle, if there is no bundle then I think the OS defaults are in play. I am thinking that the operator has a domain of trust that is gets from the bundle, that domain can be as narrow or as wide as necessary.

gtully: the implication with self signed certs would be that the bundle needs to be updated for each CR deployed. But using a shared self signed issuer can avoid that.

brusdev (approved):

Issue #1039, opened 2024-10-22
Image reference for 2.37 in CLusterServiceVersion points to the wrong SHA256


solipsistadventurer:
The ClusterServiceVersion manifest contains references to the init and kubernetes images on quay.io. These links are based on their SHA256 digest.
However, the ClusterServiceVersion for Operator version 1.2.5 appears to reference the wrong SHA256 for the corresponding version 2.37.0 application images on quay.io

image: **activemq-artemis-broker-init:artemis.2.37.0**
CSV: [e2a761725d5b](https://github.com/artemiscloud/activemq-artemis-operator/blob/09e6ae8e88e5b1b12087586398a2b8f51314be51/bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml#L3303)
Quay.io: [b6f61188379d](https://quay.io/repository/artemiscloud/activemq-artemis-broker-init/manifest/sha256:b6f61188379def8c796c738549c099bd40bcd9bb44f5c30682c5edd136347aca)

image: **activemq-artemis-broker-kubernetes:artemis.2.37.0**
CSV: [cd9a01cfe791](https://github.com/artemiscloud/activemq-artemis-operator/blob/09e6ae8e88e5b1b12087586398a2b8f51314be51/bundle/manifests/activemq-artemis-operator.clusterserviceversion.yaml#L3305)
Quay.io: [5e768aefade6](https://quay.io/repository/artemiscloud/activemq-artemis-broker-kubernetes/manifest/sha256:5e768aefade6d18cd65cb1eb5ee233844f9f7e7b52631f93a0982b67d382239d)

brusdev:
[...]

PR #1040, opened 2024-10-22
Update activemq-artemis-operator.clusterserviceversion.yaml

solipsistadventurer:
Correct the quay.io image SHA256 references to the correct application version
This is a solution to issue #1039


brusdev:
@solipsistadventurer thanks for your contribution. Those SHA256s will be automatically updated by the [update action](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/.github/workflows/update.yml) before the next release. For further details see https://github.com/artemiscloud/activemq-artemis-operator/issues/1039#issuecomment-2430081341

PR #1041, opened 2024-10-25
[#1036] when operator  cert and trust bundle are present, use them on…

gtully:
… calls to jolokia, fix console.useClientAuth PEM support


brusdev (approved):
This is a nice security improvement!

PR #1043, opened 2024-10-29
[#1036] allow operator default cert and ca secret names to be overridden

gtully:


brusdev (approved):
The new more specific secret names reduce conflict risks and simplify the identification. The new environment variables allow users to adapt the operator setup to different environments.

Issue #1044, opened 2024-10-29
Update docker image base


kevinfrommelt:
**Describe the feature**

Update the base image being used to generate the docker image for the operator. The current image is over 2 years old and contains some CVEs with fixes.

brusdev:
@kevinfrommelt thanks for the heads-up, I have just added a step to the update the base image to the workflow that is executed before every release, see https://github.com/artemiscloud/activemq-artemis-operator/commit/5107c4b968d1e450f8bc6aa75ea8a04c9f165bc2

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/753728d67dd9fdbd699d98e380a05590994de904

kevinfrommelt:
@brusdev Do you have an estimate of when the next release will be created?

brusdev:
@kevinfrommelt I have just pushed the release [1.2.7](https://github.com/artemiscloud/activemq-artemis-operator/releases/tag/1.2.7)

PR #1045, opened 2024-10-29
[#1044] Update docker image base version latest minimal image

kevinfrommelt:
Update docker image base version to latest minimal image (`ubi8-minimal:8.10-1086`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,7 +39,7 @@ RUN cp -r $REMOTE_SOURCE_DIR/app/* .
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
 RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -ldflags="-X '${GO_MODULE}/version.BuildTimestamp=`date '+%Y-%m-%dT%H:%M:%S'`'" -o manager main.go
 
-FROM registry.access.redhat.com/ubi8:8.6-855 as base-env
+FROM registry.access.redhat.com/ubi8-minimal:8.10-1086 as base-env
 
 ENV BROKER_NAME=activemq-artemis
 ENV USER_UID=1000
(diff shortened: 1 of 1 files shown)


kevinfrommelt:
Is it possible to switch to the ubi8/ubi-minimal image? https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=overview

Dockerfile
```diff
@@ -61,9 +61,6 @@ COPY --from=builder /opt/app-root/src/entrypoint/entrypoint ${USER_HOME}/bin/ent
 # Set operator bin owner and permissions
 RUN chown -R `id -u`:0 ${USER_HOME}/bin && chmod -R 755 ${USER_HOME}/bin
 
-# Upgrade packages
-RUN dnf update -y --setopt=install_weak_deps=0 && rm -rf /var/cache/yum
```
brusdev, line 65: I guess you are removing this line because dnf is not available on ubi8-minimal. If this is the only reason you can replace dnf with microdnf, see https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/assembly_types-of-container-images_building-running-and-managing-containers#con_understanding-the-ubi-init-images_assembly_types-of-container-images

brusdev (commented):
Using ubi8-minimal should reduce the size of the container image and the surface for security issues.

Dockerfile
kevinfrommelt: @brusdev Good call. Updated.

kevinfrommelt:
@brusdev Makes sense, I can wait.

Our security scanner flagged several python packages (ex. setuptools, urllib, etc) with CVEs in the ubi8 image and many of those are still present in the latest ubi8.

brusdev:
@kevinfrommelt your change LGTM, can you squash your commits into one commit?

kevinfrommelt:
@brusdev Yep. All set.

Issue #1053, opened 2024-11-20
Test migration from ActiveMQArtemisAddress CRs to broker properties


brusdev:
**Describe the bug**
The deletion of an ActiveMQArtemisAddress CR causes the deletion of the related address/queue only if the value of the field RemoveFromBrokerOnDelete is true.

The procedure to migrate from ActiveMQArtemisAddress CRs to broker properties is:

1. Update the value of the field RemoveFromBrokerOnDelete to false in the ActiveMQArtemisAddress CR
2. Add broker properties for the address/queue defined in the ActiveMQArtemisAddress CR
3. Delete the ActiveMQArtemisAddress CR

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/artemiscloud/activemq-artemis-operator/commit/1b267cbbb00cb29af4afd7c58f3d863e67fa5209

PR #1054, opened 2024-11-20
[#1053] Test migration from ActiveMQArtemisAddress CRs to broker properties

brusdev:


gtully (approved):
great to capture the migration use case. nice

Issue #1082, opened 2025-01-21
ArkMQ Operator 1.2.9 fails to reconcile the ActiveMQArtemisAddress CRs on Apache ActiveMQ Artemis version is 2.39.0


brusdev:
**Describe the bug**
The manager of the ArkMQ Operator 1.2.9 fails to reconcile the ActiveMQArtemisAddress CRs  on Apache ActiveMQ Artemis version is 2.39.0 and logs the following error:
```
ERROR ActiveMQArtemisAddressReconciler Error creating ActiveMQArtemisAddress {"address": "myAddress4", "error": "invalid character '<' looking for beginning of value"}
github.com/go-logr/logr.Logger.Error
 /opt/app-root/src/go/pkg/mod/github.com/go-logr/logr@v1.2.4/logr.go:299
github.com/arkmq-org/activemq-artemis-operator/controllers.createAddressResource
 /opt/app-root/src/controllers/activemqartemisaddress_controller.go:253
github.com/arkmq-org/activemq-artemis-operator/controllers.(*ActiveMQArtemisAddressReconciler).createQueue
 /opt/app-root/src/controllers/activemqartemisaddress_controller.go:213
github.com/arkmq-org/activemq-artemis-operator/controllers.(*ActiveMQArtemisAddressReconciler).Reconcile
 /opt/app-root/src/controllers/activemqartemisaddress_controller.go:134
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
 /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
 /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
 /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
 /opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227
```

> [!TIP]
[...]

PR #1083, opened 2025-01-21
[#1082] Fix jolokia illegal paths

brusdev:
Since Apache ActiveMQ Artemis 2.39.0 the char " in the path of the jolokia requests cause: Error 400 Suspicious Path Character, Illegal Path Character


gaohoward (approved):

gtully (approved):
makes sense, great.

Issue #1084, opened 2025-01-22
UnableToRetrieveStatus when restricted is enabled


brusdev:
**Describe the bug**
Sometimes the status of the BrokerPropertiesApplied and the BrokerVersionAligned conditions is Unknown with the reason UnableToRetrieveStatus when restricted is enabled and the test `minimal restricted rbac [It] operator role access` fails:
```
  [FAILED] in [It] - /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/controll_plane_test.go:167 @ 01/22/25 14:49:43.016
  << Timeline

  [FAILED] Timed out after 180.000s.
  The function passed to Eventually failed at /home/dbruscin/Workspace/brusdev/artemiscloud/activemq-artemis-operator/controllers/controll_plane_test.go:165 with:
  Expected
      <bool>: false
  to be true
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/83f68cc0905cc40d0b0b87b4bc4fa93c85ab9c3b

PR #1085, opened 2025-01-22
[#1084] Retry transient artemis status errors

brusdev:


gtully (approved):
great, transient is a good name

Issue #1086, opened 2025-01-22
use a broker name based on the cr name in place of the default and hard coded name


gtully:
The original Template model supported an AMQ_HOST env var that would specify the broker name. However once the operator started interacting with the broker jolokia endpoint, it used a hard coded broker name that did not respect the AMQ_HOST env var.  Today setting the env var, breaks the operator/broker interactions. The hard coded name is of little value, but it does appear as a property in  mbean names.

If the default name was based on the CR name, there would be no good reason to change it as it would be sensible and meaningful.
under the restrictedMode flag, we can fix this by using a more sensible default.

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/fc30ea09714585afdd9e979f5ebf8b1aaf73071a

PR #1087, opened 2025-01-22
[#1086] use cr.name as broker name in restricted mode

gtully:


brusdev:
Another limitation of the non-restricted deployment is that all broker instances related to a single ActiveMQArtemis CR have the same broker name. Is it a limitation also in restricted mode?

gtully:
in restricted mode, the plan size will be restricted to 1. I have raised #1089 1089 to enforce this. good catch

gtully:
I have updated this PR with the fix to #1089 thanks!

gtully:
> Another limitation of the non-restricted deployment is that all broker instances related to a single ActiveMQArtemis CR have the same broker name. Is it a limitation also in restricted mode?

yes. but even more restricted in that the deployment size should always be the default of 1. It should be ignored.

brusdev (approved):

PR #1092, opened 2025-02-11
[#1086] remove probe use of hardcoded name to reflect cr name

gtully:


brusdev (approved):
Giod catch! Is there a simple way to test the liveness probe?

Issue #1093, opened 2025-02-12
setting AMQ_NAME env var breaks the operator to broker status checks


gtully:
The operator uses a hard coded broker name attribute amq-broker and this is used for all Cr instances. The name ends up in jmx object names and metrics. There is  template AMQ_NAME env var that can change the broker config, but the operator still uses the hardoded amq-broker for calls to jolokia, breaking the status checks and the address controller operation.
The name attribute could be configured through brokerProperties and the result is the same. The operation needs to respect the AMQ_NAME env var and adjust calls to jolokia as appropriate.
This will allow changing the name attribute of the operand. In restricted mode the name attribute uses the cr.Name, which is more sensible, #1086 but it may want to be modified.

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/13891f77a1b11b9f88a964d00dadbb760d2b3d3b

PR #1094, opened 2025-02-12
[#1093] respect AMQ_NAME env var as a means to change the broker name…

gtully:
… attribute


brusdev (approved):

brusdev:
Full test suite passed

Issue #1081, opened 2025-01-21
cluster_wide_install_opr.sh on MacOS is broken


toniharju:
**Describe the bug**
The cluster_wide_install_opr.sh script gives the following error on MacOS:

```
sed: 2: "/WATCH_NAMESPACE/,/- na ...": unexpected EOF (pending }'s)
```

Apparently the problem is that MacOS has BSD sed by default which doesn't allow the semi-colons to be used in the way they are used.

I managed to fix it by changing the sed to the following POSIX standard way of doing it:

```
  sed -e "/WATCH_NAMESPACE/,/- name/ { 
    /WATCH_NAMESPACE/b
    /valueFrom:/bx
    /- name/b
    d
    :x
    s/valueFrom:/value: '${WATCH_NAMESPACE}'/
  }" $DEPLOY_PATH/operator.yaml | $KUBE_CLI apply -f -
```

I haven't tested this with GNU sed to see if it works with that.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
@toniharju could you try if replacing that line with the following line fixes the error?
```
sed -z "s/WATCH_NAMESPACE.*metadata.namespace/WATCH_NAMESPACE\n          value: ''/" deploy/operator.yaml > o.yaml 
```

toniharju:
> [@toniharju](https://github.com/toniharju) could you try if replacing that line with the following line fixes the error?
> 
> ```
> sed -z "s/WATCH_NAMESPACE.*metadata.namespace/WATCH_NAMESPACE\n          value: ''/" deploy/operator.yaml > o.yaml 
> ```

Sorry it took this long to get back to this.

Unfortunately BSD sed, or at least the version that comes pre-installed on MacOS does not support the -z flag.

brusdev:
@toniharju I created the PR #1095 to fix this issue, could you review it?

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/453e794b58c0717af397add3f209974f38891d7b

PR #1095, opened 2025-02-19
[#1081] Fix cluster_wide_install_opr.sh on MacOS

brusdev:


toniharju (changes requested):
Shouldn't the first `sed` not be the following:
```
sed "/WATCH_NAMESPACE/,/metadata.namespace/ s/valueFrom:/value: '$WATCH_NAMESPACE'/"
```

It is currently missing the `$WATCH_NAMESPACE` variable, always setting WATCH_NAMESPACE to empty instead of the value that was typed in.

brusdev:
@toniharju good catch!

Issue #685, opened 2023-09-19
Add block reconcile annotation


zamsong123:
**Is your feature request related to a problem? Please describe.**
In our business process, we need to block the resources from reconciling, if we add an annotation as "block-reconcile": "true", the reconciler will ignore this CR.

**Describe the solution you'd like**
Add a if logic in reconciler, if the CR is marked as block-reconcile, ignore it.

**Describe alternatives you've considered**
Optional. A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
Optional. Add any other context or screenshots about the feature request here.

brusdev:
@zamsong123 your enhancement is interesting, are you aware of other operators with a similar feature?

zamsong123:
@brusdev  yes, we are user of the Confluent operator and the operator is using the feature, that is very helpful in our daily operation of the clusters (e.g. when we upgrade the clusters, we will block reconcile during the process)
and the setup is very simple: add an annotation in CR (in Confluent, it's "block-reconcile": "true")

zamsong123:
@gtully thanks for your info, this also works,
however, if the value is already set as a semantic version, the block reconcile process will be more complex, and after unblock, we need to set the value back? 
I would suggest we add a new comprehensive feature for this, I tried to make some changes and it's easy and simple, tested without issue.

gtully:
great, I think we can close this issue. thanks for the feedback.

lishanley:
Hi, hope you are well! - I would like to provide a potential solution to this issue and present an annotation based solution on the CR for blocking reconcile.

I have this solution tested and would like to show this in a feature branch?

brusdev:
Hi @lishanley, sure contributions are always welcome, feel free to share your feature branch.

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/de0f020f271e0912f1b687316bf98cff7abf7aee

PR #1109, opened 2025-03-21
[#685] add support and condition for arkmq.org/block-reconcile annota…

gtully:
…tion


brusdev (approved):
This is a great addition that will help the users to have much control on the upgrade procedure.

Issue #1110, opened 2025-03-24
Typo in the secret name


john9x:
https://github.com/arkmq-org/activemq-artemis-operator/blob/3976414ecd736fc8757bd23fc934238a669d2b69/docs/help/operator.md?plain=1#L1230

must be

[broker cr name]-credential**s**-secret

brusdev:
@john9x good catch! Are planning to contribute with a PR to fix this typo?

john9x:
PR #1111

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/9d1346b2456a010596a8a785cfae3b29739ad64d

PR #1111, opened 2025-03-24
Typo in the secret name #1110

john9x:


john9x:
>add the .gitattributes file

yeah. my bad. I was a bit hasty

PR #1112, opened 2025-03-24
Add .gitattributes to ensure correct EOLs in a scripts

john9x:


.gitattributes
```diff
@@ -0,0 +1,15 @@
+# Any files should be left untouched by default
+* text=false
+* -text
```
brusdev, line 3: According to the git doc the `text=false` is equivalent to `-text` https://git-scm.com/docs/gitattributes
```suggestion
# Disable auto-converting line endings by default
* text=false
```

.gitattributes
```diff
@@ -0,0 +1,15 @@
...
+# Unix Scripts should use lf
+*.bash     text eol=lf
+*.fish     text eol=lf
+*.sh       text eol=lf
+
+
+# These are explicitly windows files and should use crlf
+*.bat      text eol=crlf
+*.cmd      text eol=crlf
+*.ps1      text eol=crlf
```
brusdev, line 15: I found a more complete list of patterns from https://github.com/gitattributes/gitattributes/blob/master/Common.gitattributes:
```suggestion
# Scripts
*.bash     text eol=lf
*.fish     text eol=lf
*.ksh      text eol=lf
*.sh       text eol=lf
*.zsh      text eol=lf
# These are explicitly windows files and should use crlf
*.bat      text eol=crlf
*.cmd      text eol=crlf
*.ps1      text eol=crlf
```

brusdev (commented):
@john9x this is a nice improvement to collaborate effectively with people who use different operating systems!

john9x:
> collaborate effectively with people who use different operating systems!

definitely. I'm on Windows + WSL

Issue #1117, opened 2025-03-25
[FAIL] security without controller brokerProperties rbac [It] security with management role access


brusdev:
**Describe the bug**
The test security without controller brokerProperties rbac [It] security with management role access is failing:
```
  [FAILED] Timed out after 199.845s.
  The function passed to Eventually failed at /home/jenkins/workspace/activemq-artemis/activemq-artemis-operator/activemq-artemis-operator/controllers/activemqartemissecurity_broker_properties_test.go:193 with:
  Expected
      <string>: 
  to contain substring
      <string>: 403
  In [It] at: /home/jenkins/workspace/activemq-artemis/activemq-artemis-operator/activemq-artemis-operator/controllers/activemqartemissecurity_broker_properties_test.go:196
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

gtully:
for reference, the support for the hawtio.role attribute was dropped in https://github.com/hawtio/hawtio/commit/34dd500b5370e52578024f690f3cee3b4b4fa2b4  for 4.3

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/865a8eb410a6142a05f6698a1c21be3b235e000d

PR #1118, opened 2025-03-25
[#1117] Fix security without controller after updating related images to 2.40.0

brusdev:


gtully (approved):
That warrants a release note in Artemis, it is a subtle change. How did you track it down?
Respect :⁠-⁠)

Issue #1116, opened 2025-03-24
ActiveMQArtemis CR OutOfSync, mismatched checksum on property file


john9x:
**Describe the bug**
The cluster deployed with config-bp secret 
```
apiVersion: v1
kind: Secret
metadata:
  name: config-bp
stringData:
  config.json: |
    {
      "addressSettings": { ... }
    ....
    }
```
Status of CR always OutOfSync
```
     Message:               out of sync on pod artemis-broker-ss-0, mismatched checksum on property file config.json, expected: 2594123130, current: 46476296. A delay can occur before a volume mount projection is refreshed.                                                                                                                                                                                                           
│     Reason:                OutOfSync                                                                                                                                                                                 │
│     Status:                False                                                                                                                                                                                     │
│     Type:                  BrokerPropertiesApplied
```
> [!TIP]
> Vote this issue reacting with :+1: or :-1:

john9x:
Maybe related to #633 - I have addresses and queues with dots

gtully:
can you paste in an example of the -bp secret content that causes the issue?

john9x:
@gtully 
```
apiVersion: v1
kind: Secret
metadata:
  name: config-bp
stringData:
  config.json: |
    {
      "addressSettings": {
        "#": {         
          "maxDeliveryAttempts": 2,
          "redeliveryDelay": 10000,
          "maxRedeliveryDelay": 100000,
          "addressFullPolicy": "PAGE"
        },
        "server.push": {
          "maxDeliveryAttempts": 2
        },
        "server.email": {
          "maxDeliveryAttempts": 5
        }
      },
      "addressConfigurations": {        
        "server.push": {
          "routingTypes": "ANYCAST",
          "queueConfigs": {
            "navixy.server.push": {
              "routingType": "ANYCAST"
            }
          }
        },
        "navixy.server.email": {
          "routingTypes": "ANYCAST",
          "queueConfigs": {
            "navixy.server.email": {
              "routingType": "ANYCAST"
            }
          }
        }
      },
      "securityRoles": {
        "#": {
          "admin": {
            "createNonDurableQueue": true,
            "deleteNonDurableQueue": true,
            "createDurableQueue": true,
            "deleteDurableQueue": true,
            "createAddress": true,
            "deleteAddress": true,
            "consume": true,
            "browse": true,
            "send": true,
            "manage": true
          }
        },
        "emd.#": {
          "admin": {
[...]

PR #1122, opened 2025-03-27
#1116 - fix check sum on json properties with dots being out of order

gtully:


gtully:
the case where a key contains a double quote is still problematic. There is a meta key.surround property in brokerProperties for this case, allowing an alternative surround string, we need to support something like that in json too.

gtully:
leaving this as a draft pending the release of 2.41.0 that supports the fileAlder32 status entry for a json properties file.

brusdev (approved):
The check sum on the file bytes is a great fix!

Issue #1128, opened 2025-04-07
add jmx exporter prometheus metrics agent for restricted mode


gtully:
restricted mode has no jetty endpoint, just a jolokia agent behind rbac. the prometheus jmx exporter now supports mtls and jaas authentication, such that it can interact with the mbean server rbac filter. 
As a result, access to metrics can now be restricted using the security settings rbac as prometheus scrapes mbeans for metrics as an authenticated user.
This gives a unified secure management and operational view of the broker.

PR #1129, opened 2025-04-07
[#1128] restricted mode metrics via prometheus jmx exporter agent

gtully:


controllers/activemqartemis_reconciler.go
```diff
@@ -2066,18 +2078,69 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) PodTemplateSpecForCR(customReso
 		fmt.Fprintln(jolokia_config, "useSslClientAuthentication=true")
 		fmt.Fprintln(jolokia_config, "disabledServices=org.jolokia.service.history.HistoryMBeanRequestInterceptor")
 		fmt.Fprintln(jolokia_config, "disableDetectors=true")
-		fmt.Fprintln(jolokia_config, "debug=false")
+		fmt.Fprintln(jolokia_config, "debug=true")
```
brusdev, line 2081: Is this change on purpose?

controllers/controll_plane_test.go
```diff
@@ -167,6 +173,57 @@ var _ = Describe("minimal", func() {
...
+			}
+
+			By("Checking metrics with mtls are visible")
+			Eventually(func(g Gomega) {
+
+				resp, err := httpClient.Get("https://" + serverName + ":8888/metrics")
+				if verbose {
+					fmt.Printf("Resp form metrics get resp: %v, error: %v\n", resp, err)
+				}
+				g.Expect(err).Should(Succeed())
```
brusdev, line 206: Should an HTTP request with another certificate fail?

gtully: the next step could be to allow the the rbac to be extended, possibly by providing keys to config in some control plane override secret

Issue #1130, opened 2025-04-08
Remove global environment defaults


brusdev:
**Describe the feature**
Remove global environment defaults

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/d3482fab6d0060794226c9e5a6fa67d209abc35a

PR #1132, opened 2025-04-09
[#1130] Remove global environment defaults

brusdev:


gtully (commented):
is all of the reconcile based on the content of the secret? afaik part of why the env reconcile worked is because of the static hack.

gtully:
> is all of the reconcile based on the content of the secret? afaik part of why the env reconcile worked is because of the static hack.

indeed it is, the env var from secret has the autoGen flag to track the creation event

brusdev:
I added a check to the `default user credential secret` test to verify that envs with autoGen flag don't change after a reconcile.

PR #1136, opened 2025-04-10
[NO-ISSUE] JSON configuration example

john9x:


gtully:
I think this depends on the fix for #1116

PR #1148, opened 2025-04-24
Move operator version update logic from update.yml to Makefile

### LINKED ISSUE #1147: Update GitHub Action to dynamically set version in config/metadata/activemq-artemis-operator.annotations.yaml
Currently, the `config/metadata/activemq-artemis-operator.annotations.yaml`  and `bundle.Dockerfile` contains a hardcoded `version` value. As part of automation improvements, update the GitHub workflow to dynamically set the version value during the release process.

Msarawan:
- Moved the operator version update logic from the update.yml GitHub Action to the Makefile
under the generate and bundle targets.
- This ensures consistency and allows PR checks to validate the operator version
across all relevant files automatically.

fixes:  [#1147](https://github.com/arkmq-org/activemq-artemis-operator/issues/1147)

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -68,10 +68,8 @@ jobs:
           VERSION_MINOR=${VERSION_MINOR/\*/${CURRENT_VERSION_MINOR}} && VERSION_MINOR=${VERSION_MINOR/+/$((CURRENT_VERSION_MINOR+1))}
           VERSION_PATCH=${VERSION_PATCH/\*/${CURRENT_VERSION_PATCH}} && VERSION_PATCH=${VERSION_PATCH/+/$((CURRENT_VERSION_PATCH+1))}
           VERSION="${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH}"
-          sed -i "s~Version = \"${CURRENT_VERSION}\"~Version = \"${VERSION}\"~" version/version.go
           sed -i "s~^VERSION ?=.*~VERSION ?= ${VERSION}~" Makefile
           sed -i "s~^OPERATOR_VERSION :=.*~OPERATOR_VERSION := ${VERSION}~" Makefile
-          sed -i "s~^LABEL version=.*~LABEL version=\"${VERSION}\"~g" Dockerfile
           make build && make generate-deploy && make bundle && git status
           git commit --all --message "Update version to ${VERSION}" || echo "nothing to commit"
 
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,9 @@ endif
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	sed -i "s~Version = \"${CURRENT_VERSION}\"~Version = \"${VERSION}\"~" version/version.go
+	sed -i "s~^LABEL version=.*~LABEL version=\"${VERSION}\"~g" Dockerfile
+
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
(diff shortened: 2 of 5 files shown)


.github/workflows/update.yml
```diff
@@ -72,6 +72,7 @@ jobs:
           sed -i "s~^VERSION ?=.*~VERSION ?= ${VERSION}~" Makefile
           sed -i "s~^OPERATOR_VERSION :=.*~OPERATOR_VERSION := ${VERSION}~" Makefile
           sed -i "s~^LABEL version=.*~LABEL version=\"${VERSION}\"~g" Dockerfile
+          sed -i "s~^  version: \".*\"~  version: \"${VERSION}\"~" config/metadata/activemq-artemis-operator.annotations.yaml
```
gtully, line 75: is this the source of the bundle.Dockerfile ?

bundle.Dockerfile
```diff
@@ -23,4 +23,4 @@ COPY bundle/tests/scorecard /tests/scorecard/
 LABEL name="arkmq-org/activemq-artemis-operator-bundle"
 LABEL description="ActiveMQ Artemis Broker Operator Bundle"
 LABEL maintainer="Roddie Kieley <rkieley@redhat.com>"
```
gtully, line 25: do we want to replace roddie's email with a reference to https://arkmq.org/

.github/workflows/update.yml
Msarawan: > is this the source of the bundle.Dockerfile ?

Yes, the `bundle.Dockerfile` is generated from the contents of `config/metadata/activemq-artemis-operator.annotations.yaml`.

Specifically, the make bundle command includes this line in the Makefile:

```
sed -e 's/annotations://' -e 's/  /LABEL /g' -e 's/: /=/g'  config/metadata/$(BUNDLE_PACKAGE).annotations.yaml >> bundle.Dockerfile
```

bundle.Dockerfile
RoddieKieley: Replacing it is a good idea, perhaps info@arkmq.org?

.github/workflows/update.yml
brusdev: @Msarawan your change is consistent with the current implementation but the current implementation is wrong because this project has a Makefile and when a user executes `make bundle` the version in the `config/metadata/activemq-artemis-operator.annotations.yaml` file should be updated.

I propose to move all lines that update the version except the one that update the `Makefile` to the related targets in the `Makefile`, i.e.:

- `sed -i "s~Version = \"${CURRENT_VERSION}\"~Version = \"${VERSION}\"~" version/version.go` to the `generate` target
- `sed -i "s~^LABEL version=.*~LABEL version=\"${VERSION}\"~g" Dockerfile` tp the `generate` target
- `sed -i "s~^  version: \".*\"~  version: \"${VERSION}\"~" config/metadata/activemq-artemis-operator.annotations.yaml`

A benefit of this proposal is that PR checks will validate the operator version across all those files.

bundle.Dockerfile
gtully: or the back to the discussions?

brusdev: The email address info@arkmq.org for the maintainer makes sense to me because it allows to send confidential emails, we could add another annotation for the support that points to the GitHub discussions however if required.

.github/workflows/update.yml
Msarawan: Thanks @brusdev , i will update my PR with your suggestions...Thank you!

brusdev (approved):

Makefile
```diff
@@ -289,6 +292,7 @@ bundle: manifests operator-sdk kustomize ## Generate bundle manifests and metada
 	sed -i.bak "s/createdAt:.*/`grep -oP 'createdAt:.*' config/manifests/bases/$(BUNDLE_PACKAGE).clusterserviceversion.yaml`/" ./bundle/manifests/*.yaml && rm ./bundle/manifests/*.bak
 	sed -i.bak 's|containerImage: .*|containerImage: '${IMG}'|' ./bundle/manifests/*.yaml && rm ./bundle/manifests/*.bak
 	sed 's/annotations://' config/metadata/$(BUNDLE_PACKAGE).annotations.yaml >> bundle/metadata/annotations.yaml
+	sed -i "s~^ version: \".*\"~ version: \"${VERSION}\"~" config/metadata/activemq-artemis-operator.annotations.yaml
```
brusdev, line 295: The file `config/metadata/$(BUNDLE_PACKAGE).annotations.yaml` is used in the line `sed 's/annotations://' config/metadata/$(BUNDLE_PACKAGE).annotations.yaml >> bundle/metadata/annotations.yaml`
```suggestion
	sed -i "s~^ version: \".*\"~ version: \"${VERSION}\"~" config/metadata/$(BUNDLE_PACKAGE).annotations.yaml
	sed 's/annotations://' config/metadata/$(BUNDLE_PACKAGE).annotations.yaml >> bundle/metadata/annotations.yaml
```

Issue #723, opened 2023-10-24
Operator test suite update needed if wanting to test pre-built container images


RoddieKieley:
**Is your feature request related to a problem? Please describe.**
When testing the operator code and associated broker container images, it is often important to not only test the exact code in question but instead run the test suite against specific pre-built container images.

**Describe the solution you'd like**
Upon discussion a number of places to start were identified.
- The version.go file lines [21 and 22](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/version/version.go#L21-L22)
- The Makefile line [9](https://github.com/artemiscloud/activemq-artemis-operator/blob/main/Makefile#L9)

There may be other places that need to be updated and other structural things done, but in theory this could be a place to start looking and updating.

**Additional context**
Looking to make the existing operator test suite more widely applicable for use.
This [documentation](https://artemiscloud.io/docs/help/building/#building-the-operator-image) may be relevant.

brusdev:
This project is now part of the ArkMQ initiative. This repository has been archived as all activities are now happening in the [corresponding ArkMQ repository](https://github.com/arkmq-org/activemq-artemis-operator). See [here](https://artemiscloud.io/community/transition/) for the ArkMQ transition details.

Feel free to create a new issue at https://github.com/arkmq-org/activemq-artemis-operator/issues

PR #1159, opened 2025-05-15
[#723] Support tests with custom container images

brusdev:


brusdev:
Full test-suite passed

controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -169,6 +171,59 @@ var _ = Describe("artemis controller", Label("do"), func() {
 		})
 	})
 
+	Context("operator deployment in default namespace", Label("do-operator-restricted"), func() {
+		It("default broker versions", func() {
+			if os.Getenv("DEPLOY_OPERATOR") == "true" {
+				By("Uninstall existing operator")
+				uninstallOperator(false, defaultNamespace)
+
+				By("install the operator again with custom relaated images")
```
gtully, line 180: little typo in relaated

controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -169,6 +171,59 @@ var _ = Describe("artemis controller", Label("do"), func() {
 		})
 	})
 
+	Context("operator deployment in default namespace", Label("do-operator-restricted"), func() {
```
gtully, line 174: I don't think that label is accurate. maybe delete or change. or change both to deploy-operator?

gtully (commented):
How does this impact the work flow where we publish a custom image as say am in our local registry and then run:
make OPERATOR_IMAGE_REPO=localhost/am OPERATOR_VERSION=v1.0 test-mk-do

that was building an IMG var in the makefile, but it looks like that is no longer in play?

Just for reference. we reference that workflow in https://github.com/arkmq-org/activemq-artemis-operator/blob/main/docs/help/building.md

controllers/activemqartemis_controller_deploy_operator_test.go
brusdev: fixed

controllers/activemqartemis_controller_deploy_operator_test.go
brusdev: I changed that label to `do-operator-with-custom-related-images`

Issue #1169, opened 2025-05-30
Restricted mode fails on OCP with FIPS enabled


brusdev:
**Describe the bug**
The broker pods deployed with the restricted mode on OCP with FIPS enabled fails to start:
```
jmx_prometheus_javaagent.io.prometheus.jmx.common.http.ConfigurationException: Exception loading SSL configuration, PEMCFG not found
```

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/28a974593b0988d929680becda57dfeec135681d

PR #1170, opened 2025-05-30
[#1169] Fix restricted mode security providers for FIPS

brusdev:


gtully (approved):
good to have both entries at last + 1, nice and consistent.

Issue #1177, opened 2025-06-25
cache broker status for a reconcile loop iteration


gtully:
Each reconcile loop makes three calls to the broker status via jmx over jolokia. this is two too many. We should reuse an instance per loop to avoid successive calls in a short time period.

gtully:
In a use case with 50+ extra mount secrets, this is more of a bug!

PR #1178, opened 2025-06-26
[#1177] cache jolokia array and broker status for the reconcile loop

gtully:


brusdev (approved):
Nice unit tests!
Full test suite passed.

PR #1179, opened 2025-06-27
first cut of service and app cr's that split broker and application

gtully:
provisioning roles. leaverage restricted mode artemis cr.


gtully:
it is time for some more eyes on this, to mould it around some real concrete use cases. Getting the spilt of features or functions between the two cr's is crucial and needs debate, but we need to be able to discuss something concrete. For example, and acceptor, part of the broker or the app?  It is currently part of the broker, but I think moving it is the next step :-)

controllers/activemqartemis_reconciler.go
```diff
@@ -2166,7 +2171,7 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) PodTemplateSpecForCR(customReso
 						},
 					},
 				},
-				InitialDelaySeconds:           1,
+				InitialDelaySeconds:           5,
```
brusdev, line 2174: Is this change on purpose?

main.go
```diff
@@ -293,6 +293,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	serviceReconciler := controllers.NewActiveMQArtemisServiceReconciler(
```
brusdev, line 296: Is the one for ActiveMQArtemisApp missing?

brusdev (commented):
This is a great feature that makes provisioning messaging apps on Kubernetes easy!

controllers/activemqartemisapp_controller.go
```diff
@@ -0,0 +1,555 @@
...
+//+kubebuilder:rbac:groups=broker.amq.io,namespace=activemq-artemis-operator,resources=activemqartemisapps/finalizers,verbs=update
+
+func (reconciler *ActiveMQArtemisAppReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	reqLogger := reconciler.log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name, "Reconciling", "ActiveMQArtemisApp")
+
+	instance := &broker.ActiveMQArtemisApp{}
+	var err = reconciler.Client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
gtully: for that one, I think it is normal operation. I don't think anything should be logged.

controllers/activemqartemisapp_controller.go
```diff
@@ -0,0 +1,555 @@
...
+
+		for name, addr := range addressTracker.multiCast {
+			fmt.Fprintf(buf, "addressConfigurations.\"%s\".routingTypes=MULTICAST\n", name)
+
+			for _, role := range addr.senderRoles {
+				fmt.Fprintf(buf, "securityRoles.\"%s\".\"%s\".send=true\n", name, producerRole(role))
+			}
+			for _, role := range addr.consumerRoles {
+				fmt.Fprintf(buf, "securityRoles.\"%s\".\"%s\".consume=true\n", name, consumerRole(role))
+			}
```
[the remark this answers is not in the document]
gtully: yes, the address collection is split on routing type, I think we won't support the multiple routing types scenario on an address. But we can revisit.

main.go
gtully: indeed it is, for the deploy operator scenario to work!

controllers/activemqartemis_reconciler.go
gtully: yes but probably unnecessary, it related to my test env.  I will revert as it will limit availability on faster machines. good catch

lavocatt:
I'm just starting to read the code but I already thought about a nice addition to the PR. It would be great to have a tutorial using the new tutorial_tester capabilities, would you like to add one?

controllers/activemqartemisservice_test.go
```diff
@@ -0,0 +1,484 @@
...
+							"forWorkQueue": "true",
+						}},
+
+					Auth: []brokerv1beta1.AppAuthType{
+						brokerv1beta1.MTLS,
+					},
+
+					Capabilities: []brokerv1beta1.AppCapabilityType{
+						{
+							// Role: appName, TODO respect role for mtls
```
lavocatt, line 269: I have a question regarding this TODO.
Indeed working on the tutorial, I noticed that I was setting commonNames in the certificates but these names aren't propagated to other fields anywhere.
So my question is, how does that work?

controllers/activemqartemisservice_test.go
```diff
@@ -0,0 +1,484 @@
...
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      appName,
+					Namespace: defaultNamespace,
+				},
+				Spec: brokerv1beta1.ActiveMQArtemisAppSpec{
+
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"forWorkQueue": "true",
```
lavocatt, line 260: I think it would be better if `forWorkQueue` would be in small caps `forworkqueue` in order to match the style of the casing in the yaml:
I''m also wondering the same for `matchLabels` that is using camel case.

```yaml
kubectl apply -f - <<EOF
kind: ActiveMQArtemisApp
metadata:
  name: first-app
  namespace: service-app-project
spec:
  selector:
    matchLabels:
      forWorkQueue: "true"
  auth:
    - mtls
  capabilities:
    - producerandconsumerof:
      - queuename: "APP.JOBS"
EOF
```

controllers/activemqartemisservice_test.go
gtully: the app operator looks for a secret and extracts the common name from the cert and uses that as the authenticated id for the cert login module, that id gets the associated the relevant producer and consumer roles. But there is currently only support for a single identity, the one from the single secret. It is just a limitation of the code at the moment that it does not support a different role.

lavocatt:
I've been experimenting with diagrams to illustrate the interaction with the kube api, the interaction between the new CRDs and the chain of trust

# interaction with the kubernetes api (could be part of some internal documentation)

```mermaid
graph LR
    subgraph "User"
        UserClient["kubectl"]
    end

    subgraph "Kubernetes API"
        APIServer["API Server"]
    end

    subgraph "Operator Pod"
        OperatorManager["Controller Manager"]
        ServiceReconciler["ActiveMQArtemisService<br/>Reconciler"]
        AppReconciler["ActiveMQArtemisApp<br/>Reconciler"]
        ArtemisReconciler["ActiveMQArtemis<br/>Reconciler"]
        OperatorManager -- manages --> ServiceReconciler
        OperatorManager -- manages --> AppReconciler
        OperatorManager -- manages --> ArtemisReconciler
    end

    subgraph "Custom Resources"
        ServiceCR["ActiveMQArtemisService CR<br/>(e.g. messaging-service)"]
        AppCR["ActiveMQArtemisApp CR<br/>(e.g. first-app)"]
    end

    subgraph "Managed Resources"
        BrokerSecrets["Secrets (JAAS, properties, etc.)<br/>(e.g. messaging-service-jaas-config)"]
        BrokerSvc["Service<br/>(e.g. messaging-service)"]
        BrokerStatefulSet["StatefulSet<br/>(e.g. messaging-service-ss)"]
        BrokerPod["Pod(s)<br/>(e.g. messaging-service-ss-0)"]
    end
    
    subgraph "Client Applications"
        ProducerJob["Job (producer)"]
        ConsumerJob["Job (consumer)"]
    end

    UserClient -- "kubectl apply" --> ServiceCR
    UserClient -- "kubectl apply" --> AppCR
    UserClient -- "kubectl apply" --> ProducerJob
    UserClient -- "kubectl apply" --> ConsumerJob

    APIServer -- notifies --> ServiceReconciler
    ServiceReconciler -- watches --> ServiceCR
    ServiceReconciler -- "creates/manages" --> BrokerSvc
    ServiceReconciler -- "creates/manages" --> BrokerSecrets
    ServiceReconciler -- "creates" --> ArtemisCR

    APIServer -- notifies --> ArtemisReconciler
    ArtemisReconciler -- watches --> ArtemisCR
    ArtemisReconciler -- "creates/manages" --> BrokerStatefulSet
    
    BrokerStatefulSet -- "manages" --> BrokerPod
    BrokerPod -- "mounts" --> BrokerSecrets

    APIServer -- notifies --> AppReconciler
    AppReconciler -- watches --> AppCR
    AppReconciler -- "updates" --> BrokerSecrets

    ProducerJob -- "connects via" --> BrokerSvc
    ConsumerJob -- "connects via" --> BrokerSvc
    BrokerSvc -- "routes to" --> BrokerPod

    linkStyle 3,4,5,6 stroke-width:2px,stroke:blue
    linkStyle 7,12,17 stroke-dasharray: 5 5
    linkStyle 8,13,18 stroke-dasharray: 2 3
    linkStyle 9,10,11,14,19 stroke:green

    classDef user fill:#c9d8f7,stroke:#333,stroke-width:2px;
    classDef k8s_api fill:#f2f2f2,stroke:#333,stroke-width:2px;
    classDef operator fill:#ffebcc,stroke:#333,stroke-width:2px;
    classDef crd fill:#e6ccff,stroke:#333,stroke-width:2px;
    classDef managed fill:#d9ead3,stroke:#333,stroke-width:2px;
    classDef client_app fill:#d0e0e3,stroke:#333,stroke-width:2px;

    class UserClient user;
    class APIServer k8s_api;
    class OperatorManager,ServiceReconciler,AppReconciler,ArtemisReconciler operator;
    class ServiceCR,AppCR,ArtemisCR crd;
    class BrokerSecrets,BrokerSvc,BrokerStatefulSet,BrokerPod managed;
    class ProducerJob,ConsumerJob client_app;
```

# sequence diagram

This is showing the flow of operation to get something up and running

```mermaid
sequenceDiagram
    participant User
    participant Operator
    participant Service as ActiveMQArtemisService
    participant App as ActiveMQArtemisApp
    participant BrokerPod as Broker Pod
    participant Producer
    participant Consumer

    User->>Operator: Apply ActiveMQArtemisService CR
    Operator->>BrokerPod: Create/Configure Broker StatefulSet
    BrokerPod->>BrokerPod: Start Broker
    Operator-->>User: Service Ready

    User->>Operator: Apply ActiveMQArtemisApp CR
    Operator->>Operator: Find matching Service via selector
    Operator->>BrokerPod: Update Broker Config (queues, roles)
    BrokerPod->>BrokerPod: Reload Config
    Operator-->>User: App Ready

    User->>Producer: Create Producer Job
    Producer->>BrokerPod: Connect (mTLS)
    Producer->>BrokerPod: Send Message to APP.JOBS

    User->>Consumer: Create Consumer Job
    Consumer->>BrokerPod: Connect (mTLS)
    Consumer->>BrokerPod: Receive Message from APP.JOBS
```

# documentation of the chain of trust

in order to visualize how the certs are being used in between the various peers

```mermaid
graph TD
    subgraph "System Components"
        Broker["Broker Pod"]
        Clients["Client Pods"]
        Operator["Operator"]
    end

    subgraph "Certificates"
        ServiceCert["Certificate<br/>messaging-service-broker-cert"]
        AppCert["Certificate<br/>first-app-app-cert"]
        ManagerCert["Certificate<br/>activemq-artemis-manager-cert"]
    end

    subgraph "Trust Anchor"
        RootCA["Root CA"]
    end

    Broker -- "Presents to Clients" --> ServiceCert
    Clients -- "Present to Broker" --> AppCert
    Operator -- "Presents to Broker" --> ManagerCert

    Clients -- "Trusts" --> RootCA
    Broker -- "Trusts" --> RootCA

    ServiceCert -- "Issued by" --> RootCA
    AppCert -- "Issued by" --> RootCA
    ManagerCert -- "Issued by" --> RootCA
```

gtully:
> The ActiveMQArtemisAppReconciler should add the `createAddress` and `createNonDurableQueue` roles to multiCast addresses to support simple MQTT apps, see https://github.com/brusdev/activemq-artemis-operator/tree/_poc_app_service_mqtt

for the first cut, to avoid auto creation, I would be leaning towards...

							// my-client.mytopic on address mytopic
							ProducerOf: []brokerv1beta1.AppAddressType{{Name: "mytopic"}},
							ConsumerOf: []brokerv1beta1.AppAddressType{{QueueName: "mytopic::my-client.mytopic"}},

but it may assume too much about the internal formatting of queues by artemis by the mtqq protocol adapter.
but the general idea is that the FQQN is used to specify the client capability and that means multicast,  and in that way we don't need auto create.
The qos is also encoded into those FQQN in some cases, which makes this trickier.
But we cannot size if we let clients dictate the resource usage.

AppAddressType may need to expose the qos in some way to capture the durability requirement.
the presence of a backlog or size on the fqqn (to indicate that there is  a persistence requirement) may be sufficient.

brusdev:
> for the first cut, to avoid auto creation, I would be leaning towards...
> 
> ```
> 						// my-client.mytopic on address mytopic
> 						ProducerOf: []brokerv1beta1.AppAddressType{{Name: "mytopic"}},
> 						ConsumerOf: []brokerv1beta1.AppAddressType{{QueueName: "mytopic::my-client.mytopic"}},
> ```

ConsumerOf  with QueueName creates a anycast queue while MQTT subscribers need a multicast queue. What about adding `SubscriberOf` capability?
```
// my-client.mytopic on address mytopic
ProducerOf: []brokerv1beta1.AppAddressType{{Name: "mytopic"}},
SubcriberOf: []brokerv1beta1.AppAddressType{{Name: "mytopic", Id: "my-client"}},
```

gtully:
how about, CompetingConsumerOf and ConsumerOf, competing means that it is anycast, consumerOf implies you get all messages, so multicast.
Whether the generic FQQN can be sufficient for the id, or whether we need to be closer to an mqtt api, I don't yet know.

gtully:
>  I'm torn about exposing fields to hide internal naming convection of the Apache ActiveMQ Artemis protocol managers or just accepting FQQN and leave the burden to the users. 

I think the FQQN option is the most useful and powerful, we will have to make it clear when and how to use it for sure.
I think we can augment that with alternatives if the need arises.

> 
> We should also figure out a way to expose other queue features: durability, lifetime, expiration, redelivery, filter.
For sure, there is lots of work to do here. Some of those fit into a "broker service type", some like durability need to be converted into some sort of resource constraint on memory or disk. Filters possibly could be in the capability or address as a first class entity. 

lots of scenario tests to do, to figure out the correct split of concerns and the use cases that we can deterministically accommodate.

controllers/activemqartemisservice_test.go
```diff
@@ -0,0 +1,518 @@
...
+			fmt.Fprintf(buf, "?transport.trustStoreType=PEMCA\\&transport.trustStoreLocation=/app/tls/ca/ca.pem")
+			fmt.Fprintf(buf, "\\&transport.keyStoreType=PEMCFG\\&transport.keyStoreLocation=/app/tls/pem/tls.pemcfg")
+
+			serviceUrl := buf.String()
+
+			By("deploying single producer to send one message")
+			producer := jobTemplate(
+				"producer",
+				1,
+				[]string{"/bin/sh", "-c", "exec java -classpath /opt/amq/lib/*:/opt/amq/lib/extra/* org.apache.activemq.artemis.cli.Artemis producer --protocol=AMQP --user p --password passwd --url " + serviceUrl + " --message-count 1 --destination queue://APP.JOBS;"},
```
lavocatt, line 425: `--user p --password passwd` do we need to keep these?

PROJECT
```diff
@@ -135,4 +135,20 @@ resources:
...
+  path: github.com/arkmq-org/activemq-artemis-operator/api/v1beta1
+  version: v1beta1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: amq.io
+  group: broker
+  kind: ActiveMQArtemisApp
+  path: github.com/arkmq-org/activemq-artemis-operator/api/v1beta1
+  version: v1beta1
```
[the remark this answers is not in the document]
brusdev: We can change the name of the group as part of the [Project Refactoring and Roadmap to v3.0.0](https://github.com/arkmq-org/activemq-artemis-operator/discussions/1319)

controllers/activemqartemisservice_test.go
```diff
@@ -0,0 +1,595 @@
...
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      serviceName,
+					Namespace: defaultNamespace,
+					Labels:    map[string]string{"env": "production"},
+				},
+				Spec: brokerv1beta1.ActiveMQArtemisBrokerServiceSpec{
+
+					// to get jaas config in properties
+					Image: StringToPtr(brokerImage),
```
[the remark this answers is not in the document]
gtully: removed it now that 2.50.0 is released

controllers/activemqartemis_controller_unit_test.go
```diff
@@ -125,7 +129,7 @@ func TestValidateBrokerPropsDuplicateOnFirstEquals(t *testing.T) {
 	assert.True(t, strings.Contains(condition.Message, "nameWith"))
 }
 
-func TestValidateBrokerPropsDuplicateOnFirstEqualsIncorrectButUnrealisticForOurBrokerConfigUsecase(t *testing.T) {
+func TestValidateBrokerPropsDuplicateOnFirstEqualsCorrect(t *testing.T) {
```
lavocatt: I don't understand why we need to change the behavior on duplicated properties.
This could cause regression for existing deployments isn't it?

controllers/activemqartemis_reconciler.go
lavocatt: I'd consider to refactor this function. The problem I'm facing with it is that it does lots, and splitting it would mean an easier time building the mental picture of what's actually going on in the secrets generation.

My current mental picture is that this is a flow chart, entry points, bunch of transformations and secrets as a result. We could name each transformation as a sub function, with some docs and tests. Even spiting that into multiple files for every main topic. There's at least, authentication, rbac, scrapping and pki as big topics.

lavocatt (commented):
That's great work!

I've caught a few things that could get improved:

1. Documentation: especially on the business critical sections such as scheduling and user facing bits
2. Code extraction: getting the common broker properties/secrets generation and handling in a common package to reduce the scattering of fmt.println functions accross the codebase
3. Bin-packing: consolidate the scheduling of the apps onto the services

Let me know if you want to discuss these items further.

PR #1193, opened 2025-07-28
[NO-ISSUE] tutorials: use the markdown-runner

lavocatt:
At the core of this contribution is to the usage of the newly created
    https://github.com/arkmq-org/markdown-runner tool to test the tutorials.
    
    Now on every PR the tutorials are going to get executed.


.github/workflows/tutorial.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Install markdown-runner
+        run: go install github.com/arkmq-org/markdown-runner@latest
+
+      - name: Check for changed tutorial files
+        id: changed-tutorial-files
+        if: github.event_name == 'pull_request'
+        uses: tj-actions/changed-files@v44
+        with:
+          files: docs/tutorials/**
```
brusdev, line 57: What is the purpose of this step?

.github/workflows/tutorial.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: |
+          LATEST_TAG=$(gh release list --limit 1 --json tagName --jq .[0].tagName)
+          echo "tag=${LATEST_TAG}" >> $GITHUB_OUTPUT
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run the tutorials for the latest release build
+        if: ${{ (github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && !inputs.skipTutorials)) }}
+        run: |
+          echo "Checking out latest release tag: ${{ steps.get_tag.outputs.tag }}"
```
brusdev, line 71: Why checking out the latest release tag? Should the current branch/tag (the one selected before clicking the button `Run workflow` or the main for scheduled executions) be checked?

.github/workflows/tutorial.yml
```diff
@@ -0,0 +1,91 @@
+name: Tutorials
```
brusdev, line 1: I like the idea to have an independent workflow for tutorials but most of the steps duplicate the ci workflow steps. An alternative to split the those workflows could be to define a new GitHub action to setup/execute markdown-runner.

brusdev (commented):
Using markdown-runner will reduce the required direct deps and the possible related security issues!

.github/workflows/tutorial.yml
lavocatt: I agree with you that duplicating steps is not optimal. I think there's mechanisms we could use to compose different workflows together.

I haven't had time to study these, but there's https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action and https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows.

Maybe we could figure out something nice for that?
The question is, would this be a follow-up work to tidy-up the CI or would you prefer to have this in this PR?

Asking the question because it might take some time to converge to a solution, and Gary's work might depend on this PR for his own tutorial.

brusdev (approved):

PR #1206, opened 2025-09-11
Update operator package name with arkmq-org prefix

Msarawan:
Updated operator package name with `arkmq-org` prefix.


brusdev (commented):
Can you also update the catalog resources and steps in the ci.yaml?

brusdev (commented):
The scorecard tests are failing because the field metadata.name in config/manifests/bases/arkmq-org-activemq-artemis-operator.clusterserviceversion.yaml is wrong.
We need to change it to `arkmq-org-activemq-artemis-operator.v0.0.0` and execute `make bundle` to fix the scorecard test failures.

brusdev (approved):

PR #1207, opened 2025-09-11
fix the tutorials

lavocatt:
In this PR we make sure that every tutorial targets the broker version of the operator.


brusdev (commented):
Keeping the tutorials up to date and working is essential, nice improvement!

Issue #1224, opened 2025-09-28
Helm chart for broker installation


john9x:
My draft is [here](https://github.com/john9x/activemq-artemis-broker-chart)
We need to decide when it will be hosted (separate repo? this repo /helm-charts ?) and then I can provide a PR.

brusdev:
Maybe the name `arkmq-org-activemq-artemis-broker` is generic and the Apache ActiveMQ Artemis is highly configurable. I'm not sure we could create a single helm chart to cover all deployment cases. I would start with a more specific name, i.e. if the helm chart deploys a single broker the name could be arkmq-org-activemq-artemis-single-broker,  if the helm chart deploys a federation broker the name could be arkmq-org-activemq-artemis-broker-federation, ...

john9x:
Should we support the CRs ActiveMQArtemisAddress, ActiveMQArtemisScaledown, ActiveMQArtemisSecurity ? AFAIK it is deprecated CRs.

john9x:
> a section for extra mounts to handle the edge use cases

The ability to use manually added configs in extraMounts remains, sure. But I don't like that these secrets would be created externally, out of Helm scope.

What about a more complex approach?
```yaml
configurations:
  - type: properties
    name: props-config (? not sure that we need a name)
    content:
      - "a=b"
... additionally content can be expressed as string
    content: |
      a=b
      c=d
  - type: json (?)
    name: json-config
    content:
      a: b
... additionally content can be expressed as string
    content: |
      {"a":"b"}
```

john9x:
LGTM

PR #1234, opened 2025-10-09
[#1224] Helm chart for broker installation

john9x:
Draft for substantive discussion


gtully:
Hi @john9x these are great. 

I am thinking this is essentially a new API for a broker service. There is an opportunity here to fix some things. like splitting the broker config from the kube config.

it may make sense to consider a use case and have the values meet that use case. Then consider carefully what else is exposed. Every 'option' in the values needs to be  maintained etc.
Which use cases need a service on something other than 61616?  - and does the broker ever need to be on another port? configuring the port of the service may be sufficient

acceptors[] in particular, those result in ingress with some config options and broker tls config, but the same can be configured with broker properties and helm defined ingress etc. 

I guess my comment is to keep this very simple and targeted at first and then carefully extend the API when there are cases that the operator should clearly  be doing work that helm and brokerProperties combined cannot do.

PR #1242, opened 2025-10-14
[#1227] allow control plane config override

lavocatt:
In restricted mode, the operator generates default control plane authentication configuration including certificate-based user mappings (`_cert-users`), role mappings (`_cert-roles`), and JAAS login configuration (`login.config`). These defaults may not work for all scenarios.

You can override these defaults by creating an optional secret named `[cr-name]-control-plane-override` in the same namespace as your ActiveMQArtemis CR. This secret allows you to customize the authentication configuration for:

* **mTLS communications** between operator and broker (operand)
* **mTLS communications** between broker instances (operand to operand)
* **Metrics and monitoring** access control


lavocatt (commented):
The code sounds sensible to me. I just think that a few tests can be removed. Looking forward for your feedback as well on this one.

gtully (changes requested):

docs/help/operator.md
```diff
@@ -1058,6 +1058,98 @@ With the possiblity of configuring arbritary jaas login modules directly, the Ar
...
+
+The following keys in the override secret will replace the default generated values:
+
+* `_cert-users`: Certificate DN to username mapping file (Java properties format
+  with regex patterns)
+* `_cert-roles`: Username to role mapping file (Java properties format)
+* `login.config`: JAAS login configuration for the http_server_authenticator
+  realm
+
+Any other keys in the secret will be ignored.
```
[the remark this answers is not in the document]
lavocatt: Interesting, I would've thought that limiting the override to a set of secrets would make sense.

controllers/activemqartemis_reconciler.go
```diff
@@ -2756,6 +2761,47 @@ func getPropertiesResourceNsName(artemis *brokerv1beta1.ActiveMQArtemis) types.N
...
 }
 
+// applyControlPlaneOverrides applies control plane configuration overrides from secrets.
+// It first checks for CR-specific override secret ([cr-name]-control-plane-override),
+// then falls back to shared override secret (control-plane-override).
+// Each key in the override secret completely replaces the corresponding key in brokerPropertiesMapData.
+func applyControlPlaneOverrides(customResource *brokerv1beta1.ActiveMQArtemis, client rtclient.Client, brokerPropertiesMapData map[string]string) error {
+	ctx := context.Background()
+
+	// Try CR-specific override secret first
```
gtully: one small observation. we have logic that deals with cr specific and generic secrets in a few places, is there an opportunity to combine those into some function that we can make use of here. 
I note that in some cases we want to error if not found, unlike this case, but that can be in a method name, resolveSecret, resolveRequiredSecret etc

gtully (approved):
one small comment to consider.

brusdev (approved):

gtully (approved):

controllers/controll_plane_test.go
```diff
@@ -240,5 +263,254 @@ var _ = Describe("minimal", func() {
...
+							func(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+								return common.GetOperatorClientCertificate(k8sClient, cri)
+							}
+
+						if rootCas, err := common.GetRootCAs(k8sClient); err == nil {
+							httpClientTransport.TLSClientConfig.RootCAs = rootCas
+						}
+
+						resp, err := httpClient.Get("https://" + serverName + ":8888/metrics")
+						fmt.Printf("Test 1 - Operator cert: status=%d, err=%v\n", resp.StatusCode, err)
```
brusdev: resp could be nil if err is not nil and cause a panic error getting `resp.StatusCode`

PR #1248, opened 2025-10-20
[#1247] keep scaling down brokers alive till empty, reuse discovery g…

gtully:
…roup and track with condition

### DIFF
--- a/controllers/activemqartemis_reconciler.go
+++ b/controllers/activemqartemis_reconciler.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"hash/adler32"
 	"regexp"
+	"slices"
 	"sort"
 	"unicode"
 
@@ -96,6 +97,12 @@ const (
 	debugArgsEnvVarName      = "DEBUG_ARGS"
 	javaOptsEnvVarName       = "JAVA_OPTS"
 	jdkJavaOptionsEnvVarName = "JDK_JAVA_OPTIONS"
+
+	ScaleDownConfigTrigger        = "HAPolicyConfiguration.scaleDownConfiguration.enabled=false"
+	ScaleDownConfigTriggerOn      = "HAPolicyConfiguration.scaleDownConfiguration.enabled=true"
+	OrdinalPropertiesSuffix       = "for_ordinal_"
+	OrdinalPropertiesSuffixEnd    = "_only"
+	BrokerPropertiesOrdinalFormat = UncheckedPrefix + "%s" + OrdinalPrefixSep + OrdinalPropertiesSuffix + "%d" + OrdinalPropertiesSuffixEnd
 )
 
 var defaultMessageMigration bool = true
@@ -195,9 +202,6 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) Process(customResource *brokerv
 	if err != nil {
 		reconciler.log.Error(err, "error processing resources")
 	}
-
-	//empty the collected objects
-	reconciler.requestedResources = make(map[reflect.Type]map[string]rtclient.Object)
 
 	reconciler.log.V(1).Info("Reconciler Processing... complete", "CRD ver:", customResource.ObjectMeta.ResourceVersion, "CRD Gen:", customResource.ObjectMeta.Generation)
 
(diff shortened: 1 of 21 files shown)


controllers/activemqartemis_reconciler.go
```diff
@@ -372,23 +375,129 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessCredentials(customResour
...
+
+	if len(reconciler.jolokiaEndpoints) == 0 {
+		return NewJolokiaClientsNotFoundError(errors.New("waiting for jolokia endpoints to become available"))
+	}
+
+	if len(reconciler.jolokiaEndpoints) < int(ordinalToDrain) {
+		return NewJolokiaClientsNotFoundError(errors.New(fmt.Sprintf("waiting for jolokia endpoint for ordinal %d to become available", ordinalToDrain)))
+	}
+
+	jk := reconciler.jolokiaEndpoints[ordinalToDrain]
```
brusdev: Could you move those lines to a common function to reduce the code duplication?

controllers/managed_scaledown_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+				}, timeout, interval).Should(Succeed())
+
+				By("Scaling down to ss-0")
+				Eventually(func(g Gomega) {
+
+					getPersistedVersionedCrd(brokerCrd.ObjectMeta.Name, defaultNamespace, createdBrokerCrd)
+					createdBrokerCrd.Spec.DeploymentPlan.Size = common.Int32ToPtr(1)
+					g.Expect(k8sClient.Update(ctx, createdBrokerCrd)).Should(Succeed())
+					g.Expect(meta.IsStatusConditionTrue(createdBrokerCrd.Status.Conditions, brokerv1beta1.ScaleDownPendingConditionType)).Should(BeTrue())
```
brusdev: The update could be executed multiple times because ScaleDownPendingConditionType is not true or  the ScaleDownPendingConditionType could have already been removed.

pkg/utils/common/common.go
```diff
@@ -622,6 +618,10 @@ func getDeploymentCondition(cr *brokerv1beta1.ActiveMQArtemis, client rtclient.C
 				crReadyCondition.Message = fmt.Sprintf("%s %s", crReadyCondition.Message, PodStartingStatusDigestMessage(startingPodName, pod.Status))
 			}
 		}
+		// qualify if scaling down
+		if meta.IsStatusConditionTrue(cr.Status.Conditions, brokerv1beta1.ScaleDownPendingConditionType) {
+			crReadyCondition.Reason = brokerv1beta1.ScaleDownPendingConditionType
```
brusdev: Is this reason tested?

docs/getting-started/quick-start.md
```diff
@@ -241,6 +241,36 @@ Connection brokerURL = tcp://artemis-broker-ss-0.artemis-broker-hdls-svc.activem
...
+metadata:
+  name: artemis-broker
+spec:
+  deploymentPlan:
+    size: 2
+    persistenceEnabled: true
+    messageMigration: true
+  brokerProperties:
+		-	"HAPolicyConfiguration=PRIMARY_ONLY"
+		- "HAPolicyConfiguration.scaleDownConfiguration.discoveryGroup=my-discovery-group"
```
brusdev: A little indentation typo:
```suggestion
	- "HAPolicyConfiguration=PRIMARY_ONLY"
	- "HAPolicyConfiguration.scaleDownConfiguration.discoveryGroup=my-discovery-group"
```

docs/getting-started/quick-start.md
```diff
@@ -241,6 +241,36 @@ Connection brokerURL = tcp://artemis-broker-ss-0.artemis-broker-hdls-svc.activem
...
+    - "HAPolicyConfiguration.scaleDownConfiguration.enabled=true"
```
brusdev: The code suggests it should be false, is it?
```suggestion
    - "HAPolicyConfiguration.scaleDownConfiguration.enabled=false"
```

brusdev (commented):
This new procedure for draining messages on scale down is a great improvement, it reduces the risk of inconsistent states and exposes the required status to detect and resolve unexpected issues.

lavocatt (commented):
That's a great piece of work. If I may add some feedback:

1. Problem: If messages never drain (stuck messages, slow consumers, etc.), the scale-down will hang forever. There's no:
    - Maximum wait time
    - Force-drain option
    - Admin override mechanism
    - Alerting when stuck

2. Persistent Volumes Not Handled
  - From the docs: "Only when the broker is empty will it be removed from the stateful set and it's PV will be available for reclaim."
    - Question: What happens to the PV of the scaled-down pod?
    - Is it automatically deleted?
    - Does it remain bound?
    - Can it be reused if scaling back up?

3. Missing Test Coverage

**Tier1:**

- What's tested inmanaged_scaledown_test.go:
  - Tests happy path (3→1 scale down)
- However, edge cases are missing
  - Doesn't test scale down with failing jolokia endpoints
  - Doesn't test multiple consecutive scale-downs (3→2→1)
  - Doesn't test scale-up after scale-down
  - Doesn't test what happens if ForceFailover() fails
  - Doesn't test scale-down interruption (user deletes CR mid-scale)

**Tier2:**

Fault Injection (as follow-up)

Consider adding chaos engineering tests to validate resilience.

Scale-down is a long-running, stateful operation that crosses pod boundaries. These are exactly the scenarios where distributed systems fail in production. Even basic fault injection tests would significantly increase confidence.

controllers/activemqartemis_reconciler.go
```diff
@@ -96,6 +97,12 @@ const (
 	debugArgsEnvVarName      = "DEBUG_ARGS"
 	javaOptsEnvVarName       = "JAVA_OPTS"
 	jdkJavaOptionsEnvVarName = "JDK_JAVA_OPTIONS"
+
+	ScaleDownConfigTrigger        = "HAPolicyConfiguration.scaleDownConfiguration.enabled=false"
+	ScaleDownConfigTriggerOn      = "HAPolicyConfiguration.scaleDownConfiguration.enabled=true"
+	OrdinalPropertiesSuffix       = "for_ordinal_"
+	OrdinalPropertiesSuffixEnd    = "_only"
+	BrokerPropertiesOrdinalFormat = UncheckedPrefix + "%s" + OrdinalPrefixSep + OrdinalPropertiesSuffix + "%d" + OrdinalPropertiesSuffixEnd
```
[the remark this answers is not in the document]
gtully: the intent of the name is to not match the existing default filter, so not end in .json or .properties such that only a single broker can see it, as there is now no sub dir. For the moment this must coexist, and work while being present twice as a config location, one with a custom filter and one without.


maybe broker-0 is clearer, for_broker_0_only, would that help?

controllers/activemqartemis_reconciler.go
```diff
@@ -3051,6 +3219,16 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) createExtraConfigmapsAndSecrets
 	return extraVolumes, extraVolumeMounts, nil
 }
 
+func hasOrdinalPropertieKeyInData(brokerPropsData map[string][]byte) bool {
+	for _, key := range sortedKeysStringKeyByteValue(brokerPropsData) {
```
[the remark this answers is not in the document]
gtully: good catch, it's not needed

brusdev (approved):

brusdev:
Ful test suite passed.

Issue #1255, opened 2025-11-03
Create and verify catalog source: flaky CI


lavocatt:
error: resource mapping not found for name: "arkmq-org-broker-operator-catalog-source" namespace: "operators" from "STDIN": no matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1"
ensure CRDs are installed first

<img width="3136" height="377" alt="Image" src="https://github.com/user-attachments/assets/3e57497b-96df-4ca5-83c2-f8720011d68f" />

PR #1258, opened 2025-11-05
[#1255] Move tutorials at the end of the CI workflow

brusdev:
The execution of the tutorials is moved to the end of the CI workflow because it deletes the minikube instance required by other steps.


lavocatt (approved):
LGTM as it should fix the problem with the tutorials.
Thanks @brusdev

Issue #1259, opened 2025-11-06
Support variables in patches


brusdev:
**Describe the feature**
Patches allow you to customize the specifications of a managed resource. The names of these resources depend on the name of the custom resource. Supporting variables in patches would allow for reusable and easier-to-document patches.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/a35cc8f95d7cc2f9d2392c34d225490e4553db28

PR #1260, opened 2025-11-06
[#1259] Support variables in patches

brusdev:


gtully (approved):
it is a little difficult to follow the code, but it seems sensible to do the formatting on the patch values.
and they are still being copied to make them safe for mutation.

controllers/activemqartemis_reconciler.go
```diff
@@ -1040,6 +1038,25 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) applyTemplate(index int, templa
...
+		for objectArrayIndex, objectArrayValue := range objectArray {
+			targetArray[objectArrayIndex] = formatTemplatedObject(customResource, objectArrayValue, ordinal, itemName, resType)
+		}
+		return targetArray
+	} else if objectString, isObjectString := object.(string); isObjectString {
+		return formatTemplatedString(customResource, objectString, ordinal, itemName, resType)
+	}
+	return object
+}
+
```
lavocatt: The recursive formatTemplatedObject creates new objects at every level. For large patches with many resources, this could be memory-intensive. However this could be a very minor problem if we only expect small patches.

lavocatt (commented):
This is a great functionality, however I think more tests and documentation are needed.

As for E2E testing, there is an existing TODO that could be addressed now that this functionality is getting implemented:

https://github.com/arkmq-org/activemq-artemis-operator/blob/406b4f6b87e56275f50305e4e515cf34b62e0bf2/controllers/activemqartemis_controller_test.go#L7151

Also adding an example to the examples would be a great documentation addition.

controllers/activemqartemis_reconciler.go
brusdev: This should not be an issue because we expect only small patches and the formatted patch goes out of scope and becomes unreachable when applyTemplate returns. At this point, the related memory becomes eligible for collection by Go's garbage collector, which will automatically reclaim it during its next collection cycle.

controllers/activemqartemis_reconciler.go
```diff
@@ -1040,6 +1038,25 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) applyTemplate(index int, templa
 	return nil
 }
 
+func formatTemplatedObject(customResource *brokerv1beta1.ActiveMQArtemis, object interface{}, ordinal string, itemName string, resType string) interface{} {
```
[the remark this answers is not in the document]
brusdev: It makes sense, I'll add this feature to the doc.

controllers/activemqartemis_controller_test.go
```diff
@@ -7208,6 +7207,91 @@ var _ = Describe("artemis controller", func() {
...
+				g.Expect(deployedCondition.Status).To(Equal(metav1.ConditionFalse))
+				g.Expect(deployedCondition.Reason).To(Equal(brokerv1beta1.DeployedConditionCrudKindErrorReason))
+				g.Expect(deployedCondition.Message).To(ContainSubstring("error applying strategic merge patch"))
+			}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
+
+			// cleanup
+			CleanResource(&crd, crd.Name, defaultNamespace)
+		})
+
+		It("expect failed to create new StatefulSet", func() {
```
lavocatt: I would also rename this one to something more specific:

`It("should report validation error when patched resource contains invalid field values", func() {`

And probably a comment later on explaining why the INVALID_VALUE isn't compatible with the enum of the podManagementPolicy

lavocatt (commented):
That's great test and docs addition. I think a good next step would be to improve the way we bubble up the errors to the users.

controllers/activemqartemis_controller_test.go
```diff
@@ -7208,6 +7207,91 @@ var _ = Describe("artemis controller", func() {
...
+			brokerKey := types.NamespacedName{Name: crd.Name, Namespace: crd.Namespace}
+
+			By("verifying started via Status")
+			Eventually(func(g Gomega) {
+				g.Expect(k8sClient.Get(ctx, brokerKey, createdCrd)).Should(Succeed())
+				deployedCondition := meta.FindStatusCondition(createdCrd.Status.Conditions, brokerv1beta1.DeployedConditionType)
+				g.Expect(deployedCondition).NotTo(BeNil())
+				g.Expect(deployedCondition.Status).To(Equal(metav1.ConditionFalse))
+				g.Expect(deployedCondition.Reason).To(Equal(brokerv1beta1.DeployedConditionCrudKindErrorReason))
+				g.Expect(deployedCondition.Message).To(ContainSubstring("error applying strategic merge patch"))
```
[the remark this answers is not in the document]
brusdev: > Aftrer doing a bit of AI investigation, it seems that [at line 1034 in activemqartemis_reconciler.go](https://github.com/brusdev/activemq-artemis-operator/blob/8896bda8fd5b5eafe7750d426a745db0faaa89de/controllers/activemqartemis_reconciler.go#L1034) we're using `fmt.Errorf(..., %v)` which doesn't preserve the error chain - should probably be `%w` instead so we don't lose the detailed validation errors from `StrategicMergeMapPatch` or `FromUnstructuredWithValidation`.

From the user log point of view, `%v` and `%w` should produce the same error string. We should use `%w` when we need to inspect the original error object itself because it preserves the error chain however it is not the case.

> Also might be worth adding a log statement right there with the patch content (especially helpful when variables like `$(CR_NAME)` are involved) so operators can debug what actually got applied. Something like `reconciler.log.Error(err, "Failed to apply patch", "template_index", index, "target", target.GetName(), "patch", patch)` before the return?

That is a good point, the patch formatting could produce an unexpected output. I'll add a debug log so users can enable it only if they need to troubleshoot unexpected issues, avoiding leaking sensitive info by default.

controllers/activemqartemis_controller_test.go
brusdev: I'll use a more specific description.

lavocatt (approved):
LGTM, I like how the tests are explicit now. Thanks

PR #1280, opened 2025-12-02
[#1271] restricted mode: enable cross namespace deployment

lavocatt:
This contributions enables the cross namespace deployment of a broker. The operator is now looking for the broker-cert and the prometheus-cert in the namespace of the to-be deployed broker.

Fixes #1271


gtully (approved):
looks good, I would just remove the stack trace from the error that we report.

PR #1291, opened 2025-12-15
[#1285] ci: freeing disk space for the tests

lavocatt:
Our tests are using more storage than what's available on the machines by default. This commits add a routine cleanup that will free up to 31GB of disk in under 3 minutes (from the action's author documentation).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,11 @@ jobs:
         ports:
           - 5000:5000
     steps:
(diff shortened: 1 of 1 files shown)


gtully (approved):
looks good. I had tried some variants on my branch and at some point deleted some thing required for tar zxf to fail, that is used in the send_receive when the broker install is downloaded. so long as that tutorial still works this is great.

brusdev (approved):
Great, this improvement saved 26GiB in 1min!

Issue #1261, opened 2025-11-06
Set manager resources


brusdev:
**Describe the feature**
Missing resource requests can lead to resource contention, node instability, and unpredictable application performance.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #1293, opened 2025-12-16
[#1261] Set manager resources

brusdev:
Missing resource requests can lead to resource contention, node instability, and unpredictable application performance.

### DIFF
--- a/controllers/activemqartemis_controller_deploy_operator_test.go
+++ b/controllers/activemqartemis_controller_deploy_operator_test.go
@@ -20,7 +20,10 @@ package controllers
 
 import (
 	"bytes"
+	"context"
+	_ "embed"
 	"os"
+	"strconv"
 	"strings"
 
 	"bufio"
@@ -34,8 +37,15 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+//go:embed testdata/dummy_broker.py
+var dummyPythonScript string
 
 var _ = Describe("artemis controller", Label("do"), func() {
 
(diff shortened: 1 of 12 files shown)


controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -34,9 +36,27 @@ import (
...
+from http.server import SimpleHTTPRequestHandler, HTTPServer
+class DummyHandler(SimpleHTTPRequestHandler):
+	def do_GET(self):
+		response_content = '{\"request\":{\"mbean\":\"org.apache.activemq.artemis:broker=\\\"0.0.0.0\\\"\",\"attribute\":\"Status\",\"type\":\"read\"},\"value\":\"{\\\"configuration\\\":{\\\"properties\\\":{\\\"broker.properties\\\":{\\\"alder32\\\":\\\"1\\\",\\\"fileAlder32\\\":\\\"1285359230\\\"}}},\\\"server\\\":{\\\"jaas\\\":{\\\"properties\\\":{}},\\\"state\\\":\\\"STARTED\\\",\\\"version\\\":\\\"2.42.0\\\",\\\"nodeId\\\":\\\"27d63164-aa5b-11f0-a502-cabd4cd943ed\\\",\\\"identity\\\":null,\\\"uptime\\\":\\\"1 minute\\\"}}\",\"status\":200}'.encode('utf-8');
+		self.send_response(200);
+		self.send_header('Content-type', 'application/json');
+		self.send_header('Content-Length', str(len(response_content)));
+		self.end_headers();
+		self.wfile.write(response_content);
+HTTPServer(('', 8161), DummyHandler).serve_forever()
```
lavocatt: I'm wondering if this shouldn't live in its own file instead, if we want to make improvements to the dummy broker in the future, it'll be easier to find.

lavocatt (commented):
It seems a very wise idea to limit the amount of resources the manager can take. Would it make sense to add an entry to the documentation as well? It would help users understand when it would matter to change the default config.

controllers/activemqartemis_controller_deploy_operator_test.go
```diff
@@ -243,10 +263,140 @@ var _ = Describe("artemis controller", Label("do"), func() {
...
+
+					currentMemoryBytes, err := strconv.ParseInt(
+						strings.TrimSpace(*currentMemoryText), 10, 64)
+					g.Expect(err).To(BeNil())
+
+					g.Expect(currentMemoryBytes < 256*1024*1024).Should(BeTrue())
+				}, timeout, interval).Should(Succeed())
+
+				uninstallOperator(false, tempNs)
+				deleteNamespace(tempNs, false, Default)
```
lavocatt: Another question, why don't we wait for the deletion of the namespaces here? Is it because a namespace with a 1000 brokers is very slow to be deleted?
I assume that the fact that this namespace isn't deleted in time for the next test isn't impacting the next test, otherwise it would fail. Therefore, if the presence of this namespaces isn't colliding with other tests, maybe we should simply leave it there?

lavocatt (approved):

PR #1312, opened 2026-01-19
[#915] provide a status condition when a resource template is unmatched

makella19:
The changes are to flag with a status condition unknown when a resource template is unmatched with any of the operator generated resources and the tests for the new logic

### DIFF
--- a/api/v1beta1/activemqartemis_types.go
+++ b/api/v1beta1/activemqartemis_types.go
@@ -847,7 +847,4 @@ const (
 
 	ReconcileBlockedType   = "ReconcileBlocked"
 	ReconcileBlockedReason = "AnnotationPresent"
-
-	UnmatchedResourceTemplateConditionType = "UnmatchedResourceTemplate"
-	UnmatchedResourceTemplateReason        = "TemplateDidNotMatchAnyResources"
 )
--- a/api/v1beta1/activemqartemis_types.go
+++ b/api/v1beta1/activemqartemis_types.go
@@ -847,4 +847,7 @@ const (
 
 	ReconcileBlockedType   = "ReconcileBlocked"
 	ReconcileBlockedReason = "AnnotationPresent"
+
+	UnmatchedResourceTemplateConditionType = "UnmatchedResourceTemplate"
+	UnmatchedResourceTemplateReason        = "TemplateDidNotMatchAnyResources"
 )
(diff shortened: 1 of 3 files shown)


api/v1beta1/activemqartemis_types.go
```diff
@@ -847,4 +847,7 @@ const (
 
 	ReconcileBlockedType   = "ReconcileBlocked"
 	ReconcileBlockedReason = "AnnotationPresent"
+
+	UnmatchedResourceTemplateConditionType = "UnmatchedResourceTemplate"
+	UnmatchedResourceTemplateReason        = "TemplateDidNotMatchAnyResources"
```
brusdev: UnmatchedResourceTemplateConditionType/UnmatchedResourceTemplateReason could be a special of ValidConditionType/ValidConditionUnknownReason.

controllers/activemqartemis_reconciler.go
```diff
@@ -1641,6 +1648,28 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessResources(customResource
...
+	// Check for matched resource templates and status condition update
+	var unmatchedIndices []int
+	for i := range customResource.Spec.ResourceTemplates {
+		if !reconciler.matchedTemplates[i] {
+			unmatchedIndices = append(unmatchedIndices, i)
+		}
+	}
+
+	if len(unmatchedIndices) > 0 {
+		message := fmt.Sprintf("Resource Template(s) at index/indices %v did not match operator generated resources", unmatchedIndices)
```
brusdev: Having all the unmatched indices in the condition message is very helful for the users however having at least one of the unmatched index is sufficient because the users will see the next unmatched index after fixing the one in the condition message.

brusdev (commented):
This is a nice improvement of the status that will help the users to troubleshoot invalid templates.

controllers/activemqartemis_reconciler.go
```diff
@@ -1641,6 +1648,28 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) ProcessResources(customResource
...
+			Type:               brokerv1beta1.ValidConditionType,
+			Status:             metav1.ConditionUnknown,
+			Reason:             brokerv1beta1.ValidConditionUnknownReason,
+			Message:            message,
+			ObservedGeneration: customResource.Generation,
+		})
+	} else {
+		//to revert the status if the unmatched resource is corrected
+		meta.RemoveStatusCondition(&customResource.Status.Conditions, brokerv1beta1.ValidConditionType)
+	}
```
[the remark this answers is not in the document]
makella19: As we cannot move the check at the validation phase, updated the changes to only set to Unknown if there is no fatal validation error and removed the remove status condition as the reconcile loop should update it back to true

brusdev (approved):
LGTM, full test suite passed!

gtully (approved):

PR #1335, opened 2026-02-11
Add new `brokers.arkmq.org` CRD

### LINKED ISSUE #1329: Migrate CRD from ActiveMQArtemis.amq.io to Broker.arkmq.org
**Describe the feature**
Add the new Broker.arkmq.org Custom Resource Definition to the operator, apply deprecation notices/tags to the ActiveMQArtemis.amq.io CR and update the controller logic to support the new Broker resource while maintaining backward compatibility for the deprecated type as per the discussion in link #1319.

> [!TIP]

Msarawan:
Added new `brokers.arkmq.org` CRD under api/v1beta2 and refactored the reconciler to use v1beta2. Broker as the canonical type. The ActiveMQArtemis controller converts v1beta1 CRs to Broker via a static field-mapped converter before reconciliation, and converts the status back afterward. The Broker controller passes its CR directly with no conversion. This reuses the existing reconciliation logic without code duplication while supporting both CRDs.

fixes: [issue#1329](https://github.com/arkmq-org/activemq-artemis-operator/issues/1329)


gtully (commented):
I added a few comments, the code reuse approach looks great. nice work.

brusdev (commented):
While converting from a Broker struct to an ActiveMQArtemis struct is quick and flexible with a JSON round-trip, it can introduce performance overhead compared to alternatives like interfaces or generics. Additionally, we should reverse the direction of this conversion. Since we are deprecating ActiveMQArtemis in favor of Broker, the Broker struct may receive new fields in the future that will not be present in the ActiveMQArtemis struct.

brusdev (commented):
The suite test (controllers/suite_test.go) is not setting up BrokerReconciler and is not adding v1beta2 to scheme.

brusdev (commented):
Implementing a static converter to map ActiveMQArtemis to the Broker struct will enhance performance and provide compile-time safety. This mitigates the risk of silent failures resulting from future schema discrepancies.

PR #1341, opened 2026-02-27
[#1340] Update owner reference APIVersion when ActiveMQArtemis is upraded

brusdev:


gtully (approved):

controllers/activemqartemis_reconciler.go
```diff
@@ -1725,7 +1728,8 @@ func (reconciler *ActiveMQArtemisReconcilerImpl) CompareSecret(deployed, request
 
 func (reconciler *ActiveMQArtemisReconcilerImpl) CompareConfigMap(deployed, requested rtclient.Object) bool {
 	// our single configMap is immutable, the name indicates a change
-	return deployed.GetName() == requested.GetName()
+	return deployed.GetName() == requested.GetName() &&
+		reconciler.ensureOwnerReferenceAPIVersion(reconciler.customResource, deployed, requested)
```
gtully: does this work with an immutable object?

gtully: It looks like it does, and it makes sense for annotations etc, the immutability is all about the data.

PR #1342, opened 2026-02-27
[NO-ISSUE] ensure no pods created for operator memory check do test

gtully:


brusdev (approved):
Great improvement, now the test requires less than 180s

PR #1345, opened 2026-03-05
[#1330] changes for refactoring of activemq artemis

makella19:


brusdev (commented):
I completely understand how much effort you’ve put into this—it is a massive undertaking! While splitting changes by file is one way to go, it might be even more helpful for everyone if we group the generated files together with their original source and split commits by subject, like the manager, bundle, helm charts, workflows, go module, container image repos, ...

Since putting everything into one single PR might get a little overwhelming to manage and review, please don’t hesitate to break it up into several smaller ones. Perhaps a great place to start would be a first PR focused specifically on refactoring the Apache Artemis project references along with their related images.

The new related images prefixes could be:
- RELATED_IMAGE_BROKER_KUBERNETES
- RELATED_IMAGE_BROKER_INIT

PR #1353, opened 2026-03-19
[#1330] Update operator deployment metadata

makella19:


brusdev (commented):
LGTM, I only found references to old file names and the old deployment metadata names in the deploy/README.md and other docs files.

brusdev (commented):
We are moving in the right direction! To align better with the future project name, we should streamline the deployment metadata by replacing activemq-artemis-operator exclusively with arkmq-org-broker-operator, rather than using arkmq-broker-operator.

Additionally, we should treat the default USER_NAME in the entrypoint/entrypoint file as deployment metadata and update it to arkmq-org-broker-operator for consistency.

brusdev (commented):
In the file `Makefile` there are 3 references to `arkmq-broker-operator`

brusdev (approved):

PR #1357, opened 2026-03-20
[#1356] adding network policies for the operand

lavocatt:
In order to secure network communications in the cluster, the operator is now providing a NetworkPolicy to only allow the known used ports to be used to reach the broker.

The NetworkPolicy is generated by inspecting the ArtemisCR, adding any acceptor, console or connector port if necessary. The operator does not parse the broker properties however and users declaring any acceptor using broker properties have to use ResourceTemplates to override the values of the NetworkPolicy.

### Operator upgrade (breaking for broker-properties users)

When upgrading from a version without NetworkPolicy support, the new reconciler will create a NetworkPolicy for every existing ActiveMQArtemis CR on the next reconciliation. The NP allows ingress only on ports the operator knows about: default headless service ports (7800, 8161, 8778, 61616), plus any ports declared in spec.acceptors and spec.connectors. Ports configured solely through spec.brokerProperties are not detected and will be blocked. Affected users must add a ResourceTemplate to open those ports in the NP. This is the most impactful concern.

### Operator downgrade (stale NP)

If downgrading to a previous operator version, the NetworkPolicy becomes orphaned -- the older reconciler doesn't manage it. However, the NP has an OwnerReference on the ActiveMQArtemis CR, so Kubernetes garbage collection will delete it when the CR is deleted. While the CR exists, the stale NP remains in place with its last-known port list.

### OCP/Kubernetes version compatibility

NetworkPolicy (networking.k8s.io/v1) is GA since Kubernetes 1.7. No version concern for any supported OpenShift release. On clusters without a
  policy-enforcing CNI, the NP object is created but has no effect.

Restricted mode: Only Jolokia (8778) and Prometheus (8888) ports are opened. This matches the existing restricted mode behavior where no user-facing acceptors/connectors are configured.

### Testing

End to end tests can only be executed on an openshift environment or on a minikube environment with a CNI backend that understands network policies.

### Documentation

A tutorial using the calisto CNI demonstrate how to extend the network policies over a
restricted broker by performing two changes:
- opening a port for an acceptor
- limiting the accessibility of the jolokia port to the operator pod

The tutorial contains demonstration of the actual packet rejections
based on the rules.

### DIFF
--- a/controllers/broker_reconciler.go
+++ b/controllers/broker_reconciler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources"
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/containers"
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/ingresses"
+	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/networkpolicies"
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/persistentvolumeclaims"
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/pods"
 	"github.com/arkmq-org/arkmq-org-broker-operator/pkg/resources/routes"
@@ -190,6 +191,10 @@ func (reconciler *BrokerReconcilerImpl) Process(customResource *v1beta2.Broker, 
 	if err != nil {
 		reconciler.log.Error(err, "Error processing console")
 		return err
+	}
+
+	if !common.OperandNetworkPolicyDisabled() {
+		reconciler.applyNetworkPolicy(customResource)
 	}
 
 	// mods to env var values sourced from secrets are not detected by process resources
@@ -546,6 +551,22 @@ func (reconciler *BrokerReconcilerImpl) applyPodDisruptionBudget(customResource 
 		MatchLabels: matchLabels,
 	}
 
+	reconciler.trackDesired(desired)
+}
+
+func (reconciler *BrokerReconcilerImpl) applyNetworkPolicy(customResource *v1beta2.Broker) {
+
+	var existing *netv1.NetworkPolicy
+	if obj := reconciler.cloneOfDeployed(reflect.TypeFor[netv1.NetworkPolicy](), customResource.Name+"-netpol"); obj != nil {
+		existing = obj.(*netv1.NetworkPolicy)
+	}
+
+	var desired *netv1.NetworkPolicy
+	if customResource.Spec.NetworkPolicy != nil {
+		desired = networkpolicies.NewNetworkPolicyFromSpec(existing, customResource)
+	} else {
+		desired = networkpolicies.NewNetworkPolicyForCR(existing, customResource)
+	}
 	reconciler.trackDesired(desired)
 }
 
@@ -1923,7 +1944,7 @@ var orderedTypes *([]reflect.Type)
 
 func getOrderedTypeList() []reflect.Type {
 	if orderedTypes == nil {
-		types := make([]reflect.Type, 7)
+		types := make([]reflect.Type, 8)
 
 		// we want to create/update in this order
 		types[0] = reflect.TypeOf(corev1.Secret{})
@@ -1933,6 +1954,7 @@ func getOrderedTypeList() []reflect.Type {
 		types[4] = reflect.TypeOf(netv1.Ingress{})
 		types[5] = reflect.TypeOf(routev1.Route{})
 		types[6] = reflect.TypeOf(policyv1.PodDisruptionBudget{})
+		types[7] = reflect.TypeOf(netv1.NetworkPolicy{})
 		orderedTypes = &types
 	}
 	return *orderedTypes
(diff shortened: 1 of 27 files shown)


brusdev (approved):
Nice security improvement!

lavocatt:
> We could have an option to prevent [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) objects from being created and leave this responsibility to the user, see https://sdk.operatorframework.io/docs/best-practices/multi-tenancy/#networkpolicy

Good catch!

pkg/resources/networkpolicies/networkpolicy.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	if !restricted {
+		for _, acceptor := range cr.Spec.Acceptors {
+			addPort(acceptor.Port)
+		}
+
+		for _, connector := range cr.Spec.Connectors {
+			addPort(connector.Port)
```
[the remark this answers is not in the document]
lavocatt: That's a good question, a connect is an ingress, but indeed some broker A connecting to broker B would need a corresponding egress route. However, I think that falls into the resource template category, WDYT?

lavocatt:
> We could have an option to prevent [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) objects from being created and leave this responsibility to the user, see https://sdk.operatorframework.io/docs/best-practices/multi-tenancy/#networkpolicy

We could have a convention where setting spec.networkPolicy to an empty spec with no rules means "don't create one"

pkg/resources/networkpolicies/networkpolicy.go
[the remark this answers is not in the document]
lavocatt: Also by default, all egress is opened for the broker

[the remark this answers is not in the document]
lavocatt: Maybe it's better removing entirely the notion of connector from an ingress perspective. I made the mistake to think connectors were inbound. But they are not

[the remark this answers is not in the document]
brusdev: Removing entirely the notion of connectors from an ingress perspective makes sense to me. While maintaining open egress is highly permissive, implementing restrictive network policies for ingress is a solid foundational step toward securing the environment.

brusdev:
> We could have a convention where setting spec.networkPolicy to an empty spec with no rules means "don't create one"

That's a smart thought, but it feels somewhat counterintuitive. Why don't we just start with an environment variable in the operator manager that lets us toggle off network policies for all operands?

brusdev:
Could you add a doc section that describes how the operator manages network policy for the operands and how to customize or disable this feature?

lavocatt:
You're making good points @brusdev. I will add an env variable as well as the required documentation.

controllers/broker_reconciler.go
```diff
@@ -549,6 +554,22 @@ func (reconciler *BrokerReconcilerImpl) applyPodDisruptionBudget(customResource
...
+func (reconciler *BrokerReconcilerImpl) applyNetworkPolicy(customResource *v1beta2.Broker) {
+
+	var existing *netv1.NetworkPolicy
+	if obj := reconciler.cloneOfDeployed(reflect.TypeFor[netv1.NetworkPolicy](), customResource.Name+"-netpol"); obj != nil {
+		existing = obj.(*netv1.NetworkPolicy)
+	}
+
+	var desired *netv1.NetworkPolicy
+	if customResource.Spec.NetworkPolicy != nil {
+		desired = networkpolicies.NewNetworkPolicyFromSpec(existing, customResource)
```
brusdev, line 566: The reconciler should overwrite the pod selector to ensure it matches the pod labels and avoid users from setting it, reducing the effort for customizing the network policy and the risk of errors.

controllers/brokerservice_controller.go
```diff
@@ -269,17 +261,54 @@ func (reconciler *BrokerServiceInstanceReconciler) processAppSecrets() (err erro
...
+		err = reconciler.processControlPlaneOverrideSecret(apps)
 	}
 
-	return err
+	return appPorts, err
+}
+
+// buildBrokerServiceNetworkPolicy constructs a NetworkPolicySpec for a
+// BrokerService-managed broker. It opens the restricted-mode management
+// ports (Jolokia, Prometheus) plus every bound BrokerApp acceptor port.
```
gtully, line 297: This makes me think we have it wrong with apps requesting ports.

if some admin wants to restrict brokers to some port range, and provides a network policy to that effect, we can't be adding another that opens arbitrary ports.

It seems to me that the broker service needs to choose a port, based on any existing network policy and then report back.

for example, given a policy that allowed ports in the range 61615-61616, we could only support a single app acceptor and that app would get 61616. The policy is restricting what the app sees, and possibly that same service has label 61616 to make that visible.

But it is the operator/reconciler responding to the environment it finds it self in, and that means that apps don't get to choose a port. 

This thinking may or may not apply to the broker, cr. But I think it is the correct approach.

gtully (commented):
the key learning for me, network policy is additive, to open from a default restricted posture.

controllers/broker_reconciler.go
gtully: we do need to be careful to be very limited on what we enforce, or reasoned. I recall our probes got very complicated, but that was in part because we had a default before we had composition. But in general with composition, I think we have to validate and respect what is in the CR.

controllers/brokerservice_controller.go
gtully: The assumption above are wrong, in that it is not possible to restrict, but only to open with network policy. The broker service controller should add a policy for each app acceptor. The broker controller for a restricted broker should just do the control plane endpoints.

gtully: There is an open pr where I am removing the app acceptor where I can rework this to open a policy for the chosen port. That will work nicely.

gtully (approved):

PR #1374, opened 2026-04-08
[#1373] add double check match via broker service appSelectorExpressi…

gtully:
…on that can reference app service and namespaces

Assisted-by: Claude


docs/help/brokerservice-appselector.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+#### Approval-Based Access
+```yaml
+appSelectorExpression: |
+  has(appNamespace.metadata.annotations) && 
+  appNamespace.metadata.annotations["broker-access.arkmq.org/approved"] == "true"
+```
+
+## CEL Expression Guide
```
lavocatt: Probably not worth keeping the guide for the CEL language, since it's a k8s thing. A simple reference would be sufficient:

	
```
For more information on CEL syntax and capabilities:
- [CEL Language Definition](https://github.com/google/cel-spec)
- [Kubernetes CEL Documentation](https://kubernetes.io/docs/reference/using-api/cel/)
```

docs/help/brokerservice-appselector.md
```diff
@@ -0,0 +1,225 @@
...
+- Uses expression: `app.metadata.namespace == service.metadata.namespace`
+- Only BrokerApps from the service's own namespace are allowed
+- Provides namespace isolation without configuration
+
+```yaml
+spec:
+  # Omit the field entirely for same-namespace-only access
+```
+
+## Common Patterns
```
lavocatt, line 64: Would be great to have that demoed in the tutorial for the round trip.
We could get a secondary app, that doesn't get provisioned to the service due to a mismatch of selectors.

pkg/appselector/cel_bench_test.go
```diff
@@ -0,0 +1,172 @@
...
+		name string
+		expr string
+	}{
+		{"simple-equality", `app.metadata.namespace == "test"`},
+		{"prefix-match", `app.metadata.namespace.startsWith("team-")`},
+		{"list-membership", `app.metadata.namespace in ["a", "b", "c"]`},
+		{"complex-logic", `app.metadata.namespace.startsWith("team-") && app.metadata.namespace.endsWith("-prod")`},
+		{"regex-match", `app.metadata.namespace.matches("^team-[a-z]+-prod$")`},
+		{"port-check", `app.spec.acceptor.port >= 62000`},
+		{"label-check", `has(app.metadata.labels) && app.metadata.labels["tier"] == "premium"`},
```
lavocatt: It's good to test these, however negative tests are missing (simple inequality, logic inversion). Nonsensical tests are also missing, for instance, non existing logical operator.

controllers/brokerservice_controller.go
```diff
@@ -251,6 +272,9 @@ func (reconciler *BrokerServiceInstanceReconciler) processAppSecrets() (err erro
 	}
 	desired.Annotations[common.ProvisionedAppsAnnotation] = strings.Join(appIdentities, ",")
 
+	// Track rejected apps in status for user visibility
+	reconciler.status.RejectedApps = rejectedApps
```
lavocatt, line 273: Do we want to keep this list in the broker service? What if that list keeps on growing? Is there a risk to hit a k8s limit for the size of the CR?

pkg/appselector/cel.go
```diff
@@ -0,0 +1,268 @@
...
+	// celEnv is the CEL environment, created once at package initialization.
+	// It defines the available variables (app, service, appNamespace, serviceNamespace) for CEL expressions.
+	celEnv *cel.Env
+
+	// celProgramCache caches compiled CEL programs with metadata by expression string.
+	// This avoids recompiling the same expression on every evaluation.
+	// Note: Cache is unbounded, but in practice the number of unique expressions is small
+	// (typically just the default expression plus a handful of custom ones).
+	celProgramCache   = make(map[string]*celProgramMetadata)
+	celProgramCacheMu sync.RWMutex
```
lavocatt: Is the operator doing multiple reconciliation loop at the same time? I thought it wasn't to avoid races. I'm surprised to see a mutex here. It can't hurt to have one, but I'm wondering why we have to protect for concurrency.

pkg/appselector/cel.go
```diff
@@ -0,0 +1,268 @@
...
+
+var (
+	// celEnv is the CEL environment, created once at package initialization.
+	// It defines the available variables (app, service, appNamespace, serviceNamespace) for CEL expressions.
+	celEnv *cel.Env
+
+	// celProgramCache caches compiled CEL programs with metadata by expression string.
+	// This avoids recompiling the same expression on every evaluation.
+	// Note: Cache is unbounded, but in practice the number of unique expressions is small
+	// (typically just the default expression plus a handful of custom ones).
```
lavocatt: The justification for unbound is ok, however, having no ways to invalidate cache, could mean leaking memory over a long running period of time. A date based timeout could allow to avoid that, without adding too much burden on the cel codebase.

controllers/brokerservice_security_test.go
```diff
@@ -0,0 +1,473 @@
...
+	// Create BrokerApp from UNTRUSTED namespace with MANUALLY SET annotation
+	// (simulating an attacker trying to bypass access control)
+	attackerApp := &v1beta2.BrokerApp{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      appName,
+			Namespace: attackerNs,
+			Annotations: map[string]string{
+				// SECURITY ISSUE: Manually set annotation to claim access to service
+				common.AppServiceAnnotation: svcNs + ":" + svcName,
+			},
```
lavocatt, line 71: I can't find where this is enforced, can you explain how this security feature is implemented please?

lavocatt (commented):
The app selector solves an important security concerns, which application is allowed to get deployed onto a broker service.

A follow up mechanism I'd like to see is to providing app creators list of valid broker service labels they have the rights to deploy onto. With the current design it seems we have all the required information to create an automation.

pkg/appselector/cel_bench_test.go
gtully: this test could be removed, it was used to justify the cache during the impl, not ever exhaustive,

controllers/brokerservice_controller.go
gtully: I think it is important info, it is indicative of a bypass of the normal flow where the app controller makes the decision. it could possibly be replaced with an alert or event. It happens when the app cr gets the annotation possibly in error or maliciously. 
but it is an area that needs to change. using a status field would be better, as that subresource can be subject to rbac so can be restricted to the app controller role.
then it may also need a label to make it easy to index, but this needs careful management in the reconcile loop.

pkg/appselector/cel.go
gtully: both the app and service controller make use of the cache, there are two single threaded users.

pkg/appselector/cel.go
gtully: yes, but also if the operator ooms worst case it will restart and all will be good and the follower will take over in the mean time.

controllers/brokerservice_security_test.go
gtully: the selector expression will catch this if the service is locked down. The service controller applies the expression, independent of the app controller.

gtully: see the check in https://github.com/arkmq-org/activemq-artemis-operator/pull/1374/changes#diff-50c7f99fb9e870b124084ba716cb8090fb6f2f43084425e5a96f3593aa137d32R459

docs/help/brokerservice-appselector.md
gtully: deleted some of that extra content. thanks!

docs/help/brokerservice-appselector.md
gtully: I want to keep that round trip simple, and don't want to add more tutorials till we get the pki generation complete. But yes, I think is will need a tutorial

controllers/brokerservice_controller.go
gtully: it is also generated on each reconcile loop, so it reflects the current set of apps that match the selector but don't match the expression. It is not a historical field, it is current.

pkg/appselector/cel.go
gtully: if we have churn in production such that the cache is the problem, there are probably too many CRs in play for a single controller. but we can revisit. what i was avoiding was the need to configure a value.

controllers/brokerservice_controller.go
```diff
@@ -371,6 +399,78 @@ func getPeerLabelKey(cr *broker.BrokerService) string {
...
+}
+
+// validateAppForProvisioning performs all security checks to ensure an app is authorized
+// to be provisioned by this service.
+// Returns (valid, reason, error):
+//   - (true, "", nil) if app should be provisioned
+//   - (false, reason, nil) if app fails validation and should be skipped
+//   - (false, "", error) if cache is stale and reconciliation should be retried
+func (reconciler *BrokerServiceInstanceReconciler) validateAppForProvisioning(app *broker.BrokerApp, serviceKey string) (bool, string, error) {
+	if currentAnnotation, ok := app.Annotations[common.AppServiceAnnotation]; !ok || currentAnnotation != serviceKey {
```
brusdev: Does it mean client.list may return items that don't match the list option?
```
reconciler.Client.List(context.TODO(), apps, client.MatchingFields{common.AppServiceAnnotation: serviceKey})
```

pkg/appselector/cel.go
```diff
@@ -0,0 +1,268 @@
...
+
+	serviceMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(service)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert BrokerService to unstructured: %w", err)
+	}
+
+	// Fetch namespace objects for CEL evaluation
+	// If namespace fetch fails (e.g., due to RBAC), use empty maps
+	// CEL expressions that don't reference namespace fields will still work
+	appNamespaceMap, err := fetchNamespace(app.Namespace, client)
```
brusdev: Fetching namespaces should be the caller's responsibility.

pkg/appselector/cel.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		// Log warning but continue with empty map - expressions not referencing appNamespace will still work
+		appNamespaceMap = make(map[string]interface{})
```
brusdev: The map could include the namespace metadata name.

pkg/appselector/cel.go
```diff
@@ -0,0 +1,268 @@
...
+		"serviceNamespace": serviceNamespaceMap,
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to evaluate CEL expression: %w", err)
+	}
+
+	// Convert result to boolean
+	result, ok := out.(celtypes.Bool)
+	if !ok {
+		return false, fmt.Errorf("CEL expression returned non-boolean value: %v", out)
```
brusdev: Printing out only the value type would be safer.

pkg/appselector/cel.go
brusdev: Do CEL expressions that access non-existent fields fail?

brusdev: Is there a test covering this scenario?

brusdev (commented):
Using CEL expressions for selecting allowed apps is smart and powerful! It allows going beyond the Kubernetes RBAC limits.

pkg/appselector/cel.go
brusdev: Using an LRU cache seems a reasonable solution to protect the operator from possible DOS attacks. OOM errors can be difficult to troubleshoot.

gtully: how big should it be? recall that the cr watch cache from the sdk is also unlimited

controllers/brokerservice_controller.go
gtully: not that don't match the list options, but the cache may be stale and it may not reflect a modified app.

pkg/appselector/cel.go
gtully: that is a good point, we know the name of the namespace without any permissions. if an attribute is not found the expression errors out. currently the tests set the namespace permission at suite level so don't capture this scenario. It can be a follow up feature i think.

pkg/appselector/cel.go
gtully: the namespaces are optional parameters to the expression based on rbac. I think it can go either way.

pkg/appselector/cel.go
gtully: good catch, the validation check just reports the type, but it is dyn so not that informative, but I will remove the value. thanks.

pkg/appselector/cel.go
brusdev: The caller could avoid fetching the same namespace multiple times.

pkg/appselector/cel.go
gtully: there should only be one selector per broker service CR in the cache, that is a dynamic but also natural bound.

pkg/appselector/cel.go
gtully: actually, the namespace name is already available from the app and service metadata, the same namespace constraint is the default. really the use case for access to the namespaces is to get access to their metadata. I think having an empty map is reasonable if there is no namespace fetch cluster permission.

pkg/appselector/cel.go
gtully: lru cache with 1000 entries it is. thanks.

brusdev: Claude guesses the following estimated memory per cached entry:
  - Simple expressions (like the default): ~500 bytes - 2 KB
    - Example: app.metadata.namespace == service.metadata.namespace
  - Complex expressions (multiple conditions, nested): ~2 KB - 10 KB
    - More operators, function calls, deeper nesting = more AST nodes

A worst case could be: 1000 expressions = ~2 MB - 10 MB total, which is compatible with the current [operator manager resource limits](https://github.com/arkmq-org/activemq-artemis-operator/blob/main/config/manager/manager.yaml#L94).

pkg/appselector/cel.go
brusdev: > actually, the namespace name is already available from the app and service metadata, the same namespace constraint is the default. really the use case for access to the namespaces is to get access to their metadata. I think having an empty map is reasonable if there is no namespace fetch cluster permission.

Good point! What about using an Activation to resolve the namespaces only if required or better only if the namespace metadata values are required. That would avoid not required fetch and related warnings

gtully: I added a check that the CEL references the namespace variables to gate getting them, that will help limit the use to when necessary.

controllers/brokerservice_controller.go
gtully: there is something odd here. let me investigate a little, thank you. in this case the content of the cache is being compared with the content of the cache!

PR #1382, opened 2026-04-14
[#1356] Remove egress restrictions from operator NetworkPolicy

lavocatt:
The operator's NetworkPolicy egress rules listed specific ports for the Kubernetes API server (443, 6443), DNS (53), and broker management (8161, 8778). This breaks on platforms where the API server listens on non-standard ports (e.g. minikube uses 8443), causing the operator pod to crash-loop with i/o timeouts when reaching the API server.

Since the operator needs to talk to the API server, broker pods, and potentially other cluster services, and the target ports vary by platform, egress restrictions are removed. Ingress restrictions (health probes on 8081, metrics on 8383) are preserved.

Made-with: Cursor


brusdev (approved):
Removing egress in the policy at this stage makes sense to me. Also other operators have an empty egress policy, see https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheusOperator-networkPolicy.yaml

We can restrict it later if we figure out a way to create a rule for the API server.

PR #1388, opened 2026-04-17
[#1329] Deprecate ActiveMQArtemis CRD in favour of Broker CRD (brokers.broker.arkmq.org)

### LINKED ISSUE #1329: Migrate CRD from ActiveMQArtemis.amq.io to Broker.arkmq.org

Msarawan:
- Deprecated ActiveMQArtemis CRD (`activemqartemises.broker.amq.io`) in favour of Broker CRD (`brokers.broker.arkmq.org`).

Fixes: [issue#1329](https://github.com/arkmq-org/activemq-artemis-operator/issues/1329)

### DIFF
--- a/docs/tutorials/external_mqtt_clients.md
+++ b/docs/tutorials/external_mqtt_clients.md
@@ -48,12 +48,12 @@ $ kubectl create secret generic my-tls-secret \
 --from-literal=trustStorePassword=securepass
 ```
 
-### Deploy ActiveMQArtemis with an mqtt acceptor
-Use the following command to deploy ActiveMQArtemis with an mqtt acceptor:
+### Deploy Broker with an mqtt acceptor
+Use the following command to deploy Broker with an mqtt acceptor:
 ```shell script
 $ kubectl apply -f - <<EOF
-apiVersion: broker.amq.io/v1beta1
-kind: ActiveMQArtemis
+apiVersion: broker.arkmq.org/v1beta2
+kind: Broker
 metadata:
   name: artemis-mqtt-ssl
 spec:
--- a/api/v1beta1/activemqartemis_types.go
+++ b/api/v1beta1/activemqartemis_types.go
@@ -771,6 +771,7 @@ type ExternalConfigStatus struct {
 //+operator-sdk:csv:customresourcedefinitions:resources={{"ConfigMap", "v1"}}
 //+operator-sdk:csv:customresourcedefinitions:resources={{"StatefulSet", "apps/v1"}}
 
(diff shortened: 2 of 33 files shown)


config/crs/broker_activemqartemis_cr.yaml
```diff
@@ -1,3 +1,7 @@
+# DEPRECATED: The ActiveMQArtemis CRD is deprecated. Use the Broker CRD (brokers.arkmq.org) instead.
+# To use Broker CRD, update the apiVersion and kind references:
```
gtully: I think we should include this information in the deprecationWarning

gtully: the reference to the two metadata fields to change  makes it clear that it is identical I think, but we could add, "use the spec identical" ... not sure, it may be too much.

gtully: but the warning is the place where this information should be shared as it is visible, issue and solution in the one place.

docs/tutorials/external_mqtt_clients.md
```diff
@@ -10,6 +10,8 @@ weight: 110
 toc: true
 ---
 
+> **Note:** This tutorial uses the `ActiveMQArtemis` CRD which is deprecated. The same configuration works with the new `Broker` CRD by using `apiVersion: arkmq.org/v1beta2` and `kind: Broker`.
```
brusdev: The tutorials should be updated with the new Broker CRD. This will help new users use the new CRD.

brusdev (commented):
The final group of the new Broker, BrokerApp and brokerService CRDs will be `broker.arkmq.org`. We could update the group in a separate PR before completing the deprecation to reduce the total effort.

config/crs/broker_activemqartemis_cr.yaml
brusdev: "The spec is identical." is true now, however it could be false in the future. We could use something less strong like "The spec is compatible."

docs/examples/brokerservice-appselector.yaml
```diff
@@ -11,7 +11,7 @@
 
 ---
 # Default - Same namespace only (secure by default)
-apiVersion: arkmq.org/v1beta2
+apiVersion: broker.arkmq.org/v1beta2
```
brusdev, line 14: The group changes should be part of #1391

config/crs/broker_activemqartemis_cr.yaml
gtully: agree, "compatible" is a good word.

brusdev (approved):

Issue #1393, opened 2026-04-21
enforce resource template patch kind requirement in validation condition


gtully:
the patch component of a resource template needs a kind field. If it is not present we should treat the cr as invalid and reflect that is the status.

gtully:
or we could make it that kind in optional, that would be better.

brusdev:
+1 for making the kind optional. It would simplify the resource templates.

brusdev:
Resolved by https://github.com/arkmq-org/activemq-artemis-operator/commit/984e6bdec9bd6531f617f52a4a1fca85ddf8d88c

PR #1394, opened 2026-04-21
[#1393] trap missing kind in resource template patch in valid condition

gtully:
Assisted-by: Claude


brusdev (commented):
Is the following test expected to pass if the line `resourcetemplate.EnsurePatchHasKind(rawJson.Patch)` is removed?
```
		It("missing kind succeed validation", func() {
			if os.Getenv("USE_EXISTING_CLUSTER") == "true" {

				brokerCrd := generateBrokerSpec(defaultNamespace)

				brokerCrd.Spec.ResourceTemplates = []v1beta2.ResourceTemplate{
					{
						Selector: &v1beta2.ResourceSelector{Kind: ptr.To("Service")},
						Patch: &unstructured.Unstructured{
							Object: map[string]interface{}{
								"spec": map[string]interface{}{
									"publishNotReadyAddresses": false,
								},
							},
						},
					},
				}
				Expect(k8sClient.Create(ctx, &brokerCrd)).Should(Succeed())

				brokerKey := types.NamespacedName{
					Name:      brokerCrd.Name,
					Namespace: defaultNamespace,
				}

				createdCrd := &v1beta2.Broker{}
				By("Checking status")
				Eventually(func(g Gomega) {
					g.Expect(k8sClient.Get(ctx, brokerKey, createdCrd)).Should(Succeed())

					if verbose {
						fmt.Printf("STATUS: \n%v\n", createdCrd.Status)
					}
					g.Expect(meta.IsStatusConditionTrue(createdCrd.Status.Conditions, v1beta2.ValidConditionType)).Should(BeTrue())
					g.Expect(meta.IsStatusConditionTrue(createdCrd.Status.Conditions, v1beta2.ReadyConditionType)).Should(BeTrue())
				}, existingClusterTimeout, existingClusterInterval).Should(Succeed())

				By("cleaning up")
				CleanResource(createdCrd, brokerCrd.Name, defaultNamespace)
			}
		})
```

gtully:
> Is the following test expected to pass if the line `resourcetemplate.EnsurePatchHasKind(rawJson.Patch)` is removed?
> 
> ```
> 		It("missing kind succeed validation", func() {
> 			if os.Getenv("USE_EXISTING_CLUSTER") == "true" {
> 
> 				brokerCrd := generateBrokerSpec(defaultNamespace)
> 
> 				brokerCrd.Spec.ResourceTemplates = []v1beta2.ResourceTemplate{
> 					{
> 						Selector: &v1beta2.ResourceSelector{Kind: ptr.To("Service")},
> 						Patch: &unstructured.Unstructured{
> 							Object: map[string]interface{}{
> 								"spec": map[string]interface{}{
> 									"publishNotReadyAddresses": false,
> 								},
> 							},
> 						},
> 					},
> 				}
> 				Expect(k8sClient.Create(ctx, &brokerCrd)).Should(Succeed())
> 
> 				brokerKey := types.NamespacedName{
> 					Name:      brokerCrd.Name,
> 					Namespace: defaultNamespace,
> 				}
> 
> 				createdCrd := &v1beta2.Broker{}
> 				By("Checking status")
> 				Eventually(func(g Gomega) {
> 					g.Expect(k8sClient.Get(ctx, brokerKey, createdCrd)).Should(Succeed())
> 
> 					if verbose {
> 						fmt.Printf("STATUS: \n%v\n", createdCrd.Status)
> 					}
> 					g.Expect(meta.IsStatusConditionTrue(createdCrd.Status.Conditions, v1beta2.ValidConditionType)).Should(BeTrue())
> 					g.Expect(meta.IsStatusConditionTrue(createdCrd.Status.Conditions, v1beta2.ReadyConditionType)).Should(BeTrue())
> 				}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
> 
> 				By("cleaning up")
> 				CleanResource(createdCrd, brokerCrd.Name, defaultNamespace)
> 			}
> 		})
> ```

yes. the kind is not strictly necessary to apply the content of a patch to one of the managed kinds so long as the fields match, however the use of unstructured means that to unmarshall it must have a kind. With our custom unmarshaller we are dealing with an empty kind but possibly a third party watch would see the error. For this reason, unless we move away from unstructured, we should flag a missing kind as invalid but I think it ok to work without it in the code that actually does the patching.

gtully:
But it opens the question, can we just stop using unstructured in the type. In the schema it is just object. That would remove the hard kind requirement.

gtully:
that was a great comment and I think by just dealing with the patch as raw json we can avoid the use of unstructured and the need for a kind altogether, but still allow it for backward compatibility. This is a lot simpler, avoiding the need for custom marshalling.

brusdev (approved):
Great simplification! Reflecting it in one or more of the existing resource template patch examples would emphasize it.

Issue #1402, opened 2026-04-24
brokerservice should choose a port for an app, not the other way around


gtully:
To make a brokerservice a good citizen and network policy aware, it should choose a port for any app that needs to be provisioned on it. The apps binding secret contains the host and port information, and rather than dealing with port clashes via the status and conditions.

brusdev:
Kubernetes network policies are the official way to control traffic flow however other vendor-specific resources like a CiliumNetworkPolicy or Calico GlobalNetworkPolicy exist. Why should the port allocation depend on other resources?

gtully:
It respects the network policy, it could grow to respect more. But if there is no policy detected, it will be unbounded. Those other tools will also respect network policy so I think it is the right tool.

gtully:
It also means there is a single source of truth as to what ports are openable, if there needs to be

gtully:
a key observation is that network policy is to open rather than restrict, hens it does not make sense to respect a  restriction for a range,  best to be unlimited and open via policy for every chosen port.

PR #1404, opened 2026-04-28
#1402 remove acceptor from app cr, have the service choose a port

gtully:


gtully:
I am going to remove the network policy awareness pieces and layer over https://github.com/arkmq-org/arkmq-org-broker-operator/pull/1357 and just add policy per app to open the chosen port. Will revert from draft when that is done.

Issue #1385, opened 2026-04-15
App cr, consider an address only capability


gtully:
In #1253, the notion of preserving addresses that have no access has been raised. A use case where an app cr removes roles before removing addresses would provide a staged way to deal with existing data.
Currently the app cr has addresses and roles, possible allowing addresses with no roles is a way to configure just addresses. It could also allow separation of the role and address concepts. The sharing of address that exist in two app crs would of course be in play.

brusdev:
An independent BrokerAddress CR could better implement an independent address lifecycle.

gtully:
maybe, or the individual components of an app are composeable across app crs (and namespaces). 

at the moment, the roles wrap addresses and are only accessible via  port. The rbac via jaas is tied to the (acceptor) port.

If we tease these out for an individual app cr, they can be split across multiple crs. This needs some careful thought.

gtully:
SharedAddresss and AddressRef "address" this use case.

PR #1405, opened 2026-04-30
#1385 - introduce addressRef for cross app sharing

gtully:


api/v1beta2/broker_types.go
```diff
@@ -810,15 +810,15 @@ const (
...
-	AppsProvisionedConditionType           = "AppsProvisioned"
-	AppsProvisionedConditionSyncedReason   = "Synced"
-	AppsProvisionedConditionWaitingReason  = "WaitingForBroker"
-	AppsProvisionedConditionNotReadyReason = "BrokerNotReady"
+	AppsProvisionedConditionType                    = "AppsProvisioned"
+	AppsProvisionedConditionSyncedReason            = "Synced"
+	AppsProvisionedConditionWaitingReason           = "WaitingForBroker"
+	AppsProvisionedConditionNotReadyReason          = "BrokerNotReady"
+	AppsProvisionedConditionAddressClashReason      = "AddressClash"
+	AppsProvisionedConditionInvalidAddressRefReason = "InvalidAddressRef"
```
brusdev: Are AppsProvisionedConditionAddressClashReason and AppsProvisionedConditionInvalidAddressRefReason still required?

api/v1beta2/broker_types.go
```diff
@@ -828,6 +828,8 @@ const (
 	ValidConditionInvalidVersionReason   = "SpecVersionInvalid"
 	ValidConditionInvalidResourceName    = "InvalidResourceName"
 	ValidConditionAddressTypeError       = "AddressTypeError"
+	ValidConditionAddressRefError        = "AddressRefError"
+	ValidConditionAddressClashError      = "AddressClash"
```
brusdev: Are ValidConditionAddressRefError and ValidConditionAddressClashError still required?

gtully:
There is another use case in the mix with the spec.addresses field. 
The lifecycle of a shared address, to make it independent of the capabilities, such that spec.addresses can exist without any rbac, so they are retained but unavailable. The list could have X and FQQN X::Y type strings. But X as a multicast address should also be allowed.

To support that, we don't have the capability to determine intent, direct or pub sub (anycast or multicast), so we need to either have to have an attribute on addressType or two lists.

spec.anycast_addresses
spec.multicast_addresses

spec.address { {Name: x, Multicast: false} } 

The two lists seems more intuitive. 

But maybe the answer is to embrace FQQN, and only have FQQN in spec.addresses.

Anycast is X::X, multicast is X::Y and both cannot coexist.

the list would be X::X, Y::A, Y::B

Z:: is an empty multicast address. I don't know if there is value in that, but it would ensure there is on Z that is anycast in the mix.

This is mixing two concepts in the same list: lifecycle and sharing. Sharing would often be a subset.

Which leads to:
spec.addresses ( list of FQQN named addressType}
spec.shared_addresses (list of FQQN)

gtully:
I am coming to the conclusion that lifecycle is more fundamental than sharing. Maybe sharing becomes an attribute, shared=false (default) rather than repeating in a separate list.

gtully:
also, while FQQN is powerful, it is too verbose. It seems reflecting the real model, like an address composed of a name and some queueNames may be the simplest approach. I will explore that.

lavocatt:
I'm also wondering about the impact of wildcards for the MQTT subscribers.

What is a subscriber requests `A.#.#` and later on I create a queue `A.B.C`, my messages will get forwarded to the MQTT subscriber with the wildcard.

gtully:
> I'm also wondering about the impact of wildcards for the MQTT subscribers.
> 
> What is a subscriber requests `A.#.#` and later on I create a queue `A.B.C`, my messages will get forwarded to the MQTT subscriber with the wildcard.

We have to validate that and enforce via sharedAddresses. It is a good point for sure!

api/v1beta2/brokerapp_types.go
```diff
@@ -31,25 +31,52 @@ type BrokerAppSpec struct {
...
+	// Address is the address identifier (required)
+	// Can be simple name ("orders") or FQQN ("events::queue1" for SubscriberOf)
+	Address string `json:"address"`
+
+	// AppNamespace of owning app - for cross-app references  (optional)
+	AppNamespace string `json:"appNamespace,omitempty"`
+
+	// AppName of owning app - for cross-app references  (optional)
+	AppName string `json:"appName,omitempty"`
 }
```
brusdev, line 88: Adding subscriptions would align address and capability definitions, allowing to remove SubscriberOf and simplify the overall model.

api/v1beta2/broker_types.go
gtully: they are not, unless we resurrect them at a future date.

api/v1beta2/broker_types.go
gtully: good catch, the message is now capturing this info. but we may bring them back.

Issue #1417, opened 2026-05-26
Externalize the status check in a sidecar container


lavocatt:
To remove the need from:

- having to maintain a mTLS identity for the operator
- having to perform network calls during the reconciliation loop

let's experiment with a sidecar in the broker's pod to retrieve its status

PR #1418, opened 2026-05-26
[#1417]  Event-driven broker status via sidecar reconciliation

lavocatt:
Replace periodic requeue-based Jolokia polling with an event-driven
model. A native sidecar container monitors the broker log for lifecycle
events and patches the request-reconcile annotation on its own pod with
a unique UUID. The operator watches for annotation changes, reconciles,
and performs its standard mTLS Jolokia call to fetch status.

BEFORE (main):
  Operator --[periodic requeue]--> Jolokia mTLS --> Broker :8778
  (polls on timer regardless of broker state change)

AFTER (this branch):
```
  Pod
  +---------------------------------------------------+
  |  initContainers (restartPolicy: Always):          |
  |    broker-status sidecar:                         |
  |      tail reload.log for AMQ221007 | AMQ221087   |
  |        |                                          |
  |        +---> K8s API PATCH: pod annotation        |
  |              broker.arkmq.org/request-reconcile   |
  |                                                   |
  |  containers:                                      |
  |    broker (Jolokia on :8778)                      |
  +---------------------------------------------------+
                        |
                  predicate fires
                        |
                        v
                    Operator
              (Jolokia mTLS, same as before)
```

Watched log codes:
  AMQ221007 - Server is now active (startup)
  AMQ221087 - Configuration reload completed (Artemis 2.54+)

The sidecar runs as a K8s native sidecar (init container with
restartPolicy: Always, stable since K8s 1.33). Kubelet manages
restarts. The script reads the SA token fresh on each call to
survive kubelet token rotation.

The sidecar reuses the broker image (resolved per CR version via
RELATED_IMAGE_BROKER_KUBERNETES). Since the main container already
pulls it, all layers are cached — zero additional download. The
command override runs the status script, not the broker. Both
containers share the same pod boundary (network, SA token, volumes),
so reusing the image does not widen the attack surface. No separate
sidecar image to track, mirror, or configure.

Key changes:
- broker-status.sh delivered via Secret, runs in a dedicated
  sidecar init container (broker image, identical security context)
- Dedicated Pod watch with request-reconcile annotation predicate
  (Watches + EnqueueRequestsFromMapFunc + mapPodToBrokerCR)
- Remove periodic requeues from ProcessBrokerStatus
- UUID-based annotation values (no clearing needed — each event
  writes a new UUID, which fires the update predicate)
- RBAC: Role/RoleBinding scoped to specific pod names (resourceNames)
- minKubeVersion bumped to 1.33.0

Assisted-by: Cursor/Claude

### DIFF
--- a/bundle/manifests/arkmq-org-broker-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/arkmq-org-broker-operator.clusterserviceversion.yaml
@@ -7289,6 +7289,9 @@ spec:
           - create
           - delete
           - get
+          - list
+          - patch
+          - watch
         - apiGroups:
           - route.openshift.io
           resources:
@@ -7330,7 +7333,7 @@ spec:
   - email: info@arkmq.org
     name: ArkMQ
   maturity: beta
-  minKubeVersion: 1.22.0
+  minKubeVersion: 1.33.0
   provider:
     name: ArkMQ
     url: https://arkmq.org
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -174,6 +174,9 @@ rules:
   - create
   - delete
   - get
+  - list
+  - patch
+  - watch
 - apiGroups:
   - route.openshift.io
   resources:
--- a/controllers/broker_controller.go
+++ b/controllers/broker_controller.go
@@ -5,10 +5,12 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509/pkix"
+	_ "embed"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"maps"
+	"path"
 	"sort"
 
 	"github.com/RHsyseng/operator-utils/pkg/resource/compare"
@@ -39,7 +41,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	ctrl "sigs.k8s.io/controller-runtime"
 
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	rtclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/resources/environments"
 	svc "github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/resources/services"
@@ -50,6 +57,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	netv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -62,6 +70,22 @@ import (
 
 	policyv1 "k8s.io/api/policy/v1"
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
+)
+
+//go:embed broker_status_script.sh
+var brokerStatusScript string
+
+const (
+	sidecarSecretSuffix         = "-sidecar"
+	brokerStatusScriptKey       = "broker-status.sh"
+	sidecarContainerName        = "broker-status"
+	log4j2ConfigurationFileFlag = "-Dlog4j2.configurationFile="
+
+	// ReconcileMeAnnotationKey is the Pod annotation the sidecar script writes
+	// when it detects AMQ221007 (server active) or AMQ221087 (config reload
+	// completed). The operator watches for this annotation and triggers a
+	// reconcile + Jolokia status fetch.
+	ReconcileMeAnnotationKey = "broker.arkmq.org/request-reconcile"
 )
 
 // BrokerReconciler reconciles a Broker object (broker.arkmq.org/v1beta2)
@@ -84,6 +108,8 @@ func NewBrokerReconciler(cluster cluster.Cluster, logger logr.Logger, isOpenShif
 //+kubebuilder:rbac:groups=broker.arkmq.org,namespace=arkmq-org-broker-operator,resources=brokers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=broker.arkmq.org,namespace=arkmq-org-broker-operator,resources=brokers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=broker.arkmq.org,namespace=arkmq-org-broker-operator,resources=brokers/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",namespace=arkmq-org-broker-operator,resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,namespace=arkmq-org-broker-operator,resources=roles;rolebindings,verbs=create;delete;get;list;patch;watch
 
 func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	reqLogger := r.log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name, "Reconciling", "Broker")
@@ -119,9 +145,7 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 		if !reconcileBlocked {
 			err = reconciler.Process(customResource, *namer, r.Client, r.Scheme)
 		}
-		if reconciler.ProcessBrokerStatus(customResource, r.Client, r.Scheme) {
-			requeueRequest = true
-		}
+		reconciler.ProcessBrokerStatus(customResource, r.Client, r.Scheme)
 	}
 
 	brokerstatus.UpdateBlockedStatus(customResource, reconcileBlocked)
@@ -129,11 +153,6 @@ func (r *BrokerReconciler) Reconcile(ctx context.Context, request ctrl.Request) 
 
 	crStatusUpdateErr := r.UpdateBrokerCRStatus(customResource, r.Client, namespacedName)
 	if crStatusUpdateErr != nil {
-		requeueRequest = true
-	}
-
-	if !requeueRequest && !reconcileBlocked && hasExtraMountsForBroker(customResource) {
-		reqLogger.V(1).Info("resource has extraMounts, requeuing")
 		requeueRequest = true
 	}
 
@@ -152,18 +171,77 @@ func (r *BrokerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&v1beta2.Broker{}).
 		Owns(&appsv1.StatefulSet{}).
-		Owns(&corev1.Pod{}).
 		Owns(&corev1.Secret{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.Service{}).
 		Owns(&netv1.Ingress{}).
-		Owns(&policyv1.PodDisruptionBudget{})
+		Owns(&policyv1.PodDisruptionBudget{}).
+		Watches(&corev1.Pod{},
+			handler.EnqueueRequestsFromMapFunc(r.mapPodToBrokerCR),
+			builder.WithPredicates(reconcileMeAnnotationPredicate()),
+		)
 
 	if r.isOnOpenShift {
 		builder.Owns(&routev1.Route{})
 	}
 
 	return builder.Complete(r)
+}
+
+// reconcileMeAnnotationPredicate filters Pod events to only those where the
+// request-reconcile annotation was added or changed.
+func reconcileMeAnnotationPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			_, exists := e.Object.GetAnnotations()[ReconcileMeAnnotationKey]
+			return exists
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldVal := e.ObjectOld.GetAnnotations()[ReconcileMeAnnotationKey]
+			newVal := e.ObjectNew.GetAnnotations()[ReconcileMeAnnotationKey]
+			return newVal != "" && newVal != oldVal
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
+// mapPodToBrokerCR maps a Pod event to the owning Broker CR by traversing the
+// ownership chain: Pod -> StatefulSet -> Broker CR. All lookups hit the
+// informer cache (no API round-trips).
+func (r *BrokerReconciler) mapPodToBrokerCR(ctx context.Context, obj rtclient.Object) []reconcile.Request {
+	pod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return nil
+	}
+
+	for _, ownerRef := range pod.GetOwnerReferences() {
+		if ownerRef.Kind != "StatefulSet" {
+			continue
+		}
+		ss := &appsv1.StatefulSet{}
+		if err := r.Get(ctx, types.NamespacedName{
+			Namespace: pod.Namespace,
+			Name:      ownerRef.Name,
+		}, ss); err != nil {
+			return nil
+		}
+		for _, ssOwner := range ss.GetOwnerReferences() {
+			if ssOwner.Kind == "Broker" {
+				return []reconcile.Request{{
+					NamespacedName: types.NamespacedName{
+						Namespace: pod.Namespace,
+						Name:      ssOwner.Name,
+					},
+				}}
+			}
+		}
+	}
+	return nil
 }
 
 func (r *BrokerReconciler) UpdateBrokerCRStatus(desired *v1beta2.Broker, client rtclient.Client, namespacedName types.NamespacedName) error {
@@ -267,6 +345,10 @@ func (reconciler *BrokerReconcilerImpl) Process(customResource *v1beta2.Broker, 
 	reconciler.log.V(2).Info("Reconciler Processing...", "CRD ver", customResource.ResourceVersion, "CRD Gen", customResource.Generation)
 
 	reconciler.CurrentDeployedResources(customResource, client)
+
+	if err := reconciler.ensureSidecarConfig(client); err != nil {
+		return err
+	}
 
 	// currentStateful Set is a clone of what exists if already deployed
 	// what follows should transform the resources using the crd
@@ -918,7 +1000,7 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 
 		// jmx_exporter metrics perms
 		fmt.Fprintln(rbac, "securityRoles.\"mops.mbeanserver.queryMBeans\".metrics.view=true")
-		fmt.Fprintln(rbac, "securityRoles.\"mops.broker\".metrics.view=true") // for query remove filter
+		fmt.Fprintln(rbac, "securityRoles.\"mops.broker\".metrics.view=true") // we need view permission on the broker in order to locate through a query and retrieve it.
 		fmt.Fprintln(rbac, "securityRoles.\"mops.broker.getTotalMessageCount\".metrics.view=true")
 		fmt.Fprintln(rbac, "securityRoles.\"mops.broker.getTotalMessagesAcknowledged\".metrics.view=true")
 		fmt.Fprintln(rbac, "securityRoles.\"mops.broker.getTotalMessagesAdded\".metrics.view=true")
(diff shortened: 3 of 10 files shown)


controllers/broker_controller.go
```diff
@@ -133,7 +143,22 @@ func (r *BrokerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		// additional API server connections. The handler filters events to only
+		// enqueue Brokers whose ExtraMounts reference the changed Secret.
+		//
+		// This is required because the Secret informer and Pod informer process
+		// events independently. Without this watch, a Broker reconcile triggered
+		// by a Pod annotation change (sidecar refresh) may read a stale Secret
+		// ResourceVersion from the cache. This watch guarantees the reconcile
+		// runs after the Secret cache is updated, since the handler fires from
+		// the Secret informer's own event dispatch.
+		Watches(&corev1.Secret{}, &extraMountsSecretHandler{client: mgr.GetClient()})
```
lavocatt: This part, I asked the AI to add comments, but TBH I don't understand it yet, so I need to wrap my head around what's going on there.

I mean, the stale cache explanation looks plausible, though I find it strange.

gtully: we don't want to watch all secrets. there will typically be loads of those in a cluster. We are just interested in config reload on the broker. And secrets we own. We tried this before and it was not a flyer, way too chatty. hense we were polling. But with the side car we can be event driven again. We will get the extramount secret only after the secret volume mount projection is complete and we don't get it via an informer watch, we call the api server directly. Or doing that is the correct approach in this scenario.

gtully:
after some ruminating, I think the big win is being event driven. I think  having an event that says force-reconcile is sufficient. It could be an annotation applied with a patch on the pod. Most of the pod is immutable, but an annotation can be patched. Because we use a ss, we can optionally use resourceName in the patch rbac permission for the pod SA and it is still very well locked down. Using a 'please-reconcile' annotation means that the operator works as normal, and the broker status is not leaked in any way.

lavocatt:
> after some ruminating, I think the big win is being event driven. I think having an event that says force-reconcile is sufficient. It could be an annotation applied with a patch on the pod. Most of the pod is immutable, but an annotation can be patched. Because we use a ss, we can optionally use resourceName in the patch rbac permission for the pod SA and it is still very well locked down. Using a 'please-reconcile' annotation means that the operator works as normal, and the broker status is not leaked in any way.

That works for me. To prep the repo for this, I will work on #1431 #1432 and #1433 first. I'm marking the PR as draft in the meantime.

controllers/broker_controller.go
```diff
@@ -62,6 +69,20 @@ import (
...
+
+const (
+	brokerStatusScriptSuffix  = "-status-script"
+	brokerStatusScriptKey     = "broker-status.sh"
+	brokerLoggingConfigSuffix = "-logging-config"
+
+	// ReconcileMeAnnotationKey is the Pod annotation the sidecar script writes
+	// when it detects AMQ221087 (config reload completed). The operator watches
+	// for this annotation and triggers a reconcile + Jolokia status fetch.
+	ReconcileMeAnnotationKey = "broker.arkmq.org/reconcile-me"
```
gtully: maybe reconcile-requested, not sure about '-me' it seems a little personal or informal

bundle/manifests/arkmq-org-broker-operator.clusterserviceversion.yaml
```diff
@@ -7283,6 +7287,7 @@ spec:
           - create
           - delete
           - get
+          - update
```
gtully: I think this can be the "patch" permission

controllers/broker_controller.go
```diff
@@ -164,6 +183,61 @@ func (r *BrokerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
+// mapPodToBrokerCR maps a Pod event to the owning Broker CR by traversing the
+// ownership chain: Pod -> StatefulSet -> Broker CR.
+func (r *BrokerReconciler) mapPodToBrokerCR(ctx context.Context, obj rtclient.Object) []reconcile.Request {
```
gtully, line 218: the ownership is a good path, but could it be simpler based on just the naming convention?
the broker name becomes the ss name becomes the pod name?

controllers/broker_controller.go
```diff
@@ -1865,9 +1974,33 @@ func (reconciler *BrokerReconcilerImpl) ProcessBrokerStatus(cr *v1beta2.Broker,
...
 	}
 
+	if !retry {
+		reconciler.clearReconcileMeAnnotations(cr, client)
+	}
+
 	return retry
 }
 
+func (reconciler *BrokerReconcilerImpl) clearReconcileMeAnnotations(cr *v1beta2.Broker, client rtclient.Client) {
```
gtully: I am wondering if we should treat the reconcile annotation value like the resourceVersion, an increasing value, in that way we don't ever clear it and don't need to update or have update/patch permission and we just see changes.
The sidecar needs to be a little smarter, to do an increment and set or in some way find the current resource version to use that as a value. We could have a SA for the pod that has the specific patch permission

controllers/broker_controller.go
```diff
@@ -2429,16 +2562,6 @@ func validateExtraMountsForBroker(customResource *v1beta2.Broker, client rtclien
 	return nil, false
 }
 
-func hasExtraMountsForBroker(cr *v1beta2.Broker) bool {
-	if cr == nil {
```
gtully, line 2263: this is the big win, we don't loop on reconcile any more, great :-)

controllers/broker_controller.go
lavocatt: `request-reconcile` is what we are agreeing upon

controllers/broker_controller.go
lavocatt: I think ownership is better, it's not expensive thanks to the cache, and it avoids the pitfall of reconciling a pod that would accidentally use our naming convention

controllers/broker_controller.go
lavocatt: UUID are used now, and there's no need for clearing the annotation

config/rbac/role.yaml
```diff
@@ -32,6 +31,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - create
+  - delete
```
gtully: are all of these additions strictly necessary?

controllers/broker_controller.go
```diff
@@ -880,7 +961,7 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 
 		certUser := NewPropsWithHeader()
 		fmt.Fprintln(certUser, "hawtio=/CN = hawtio-online\\.hawtio\\.svc.*/")
-		fmt.Fprintf(certUser, "operator=/.*%s.*/\n", operatorCertSubject.CommonName) // regexp syntax start and with /
```
gtully, line 884: why remove those comments? Do they not make sense to you?

controllers/broker_controller.go
```diff
@@ -2329,4 +2429,146 @@ func GetDefaultLabelsForBroker(cr *v1beta2.Broker) map[string]string {
 	return defaultLabelData.Labels()
 }
 
+const brokerReloadLogAppenderFragment = `
+appender.reload_trigger.type = File
+appender.reload_trigger.name = ReloadTrigger
+appender.reload_trigger.fileName = /app/log/reload.log
```
gtully, line 2514: do we have to worry about the growth of this log file? A very long running broker could potentially have lots of reloads and this volume may be ephemeral.
I thin because we are looking for a single entry, having it roll over will be ok.

deploy/operator.yaml
```diff
@@ -166,6 +166,10 @@ spec:
           value: quay.io/arkmq-org/arkmq-org-broker-init@sha256:45c3184c51abc7b378d696953df53eaa463464eb85f36377851ab9a486f1556f
         - name: RELATED_IMAGE_BROKER_KUBERNETES_2540
           value: quay.io/arkmq-org/arkmq-org-broker-kubernetes@sha256:c4676a4647e8fac743453f1e0e6a08107a9bd7989215b39bb3ea583317366f73
+        - name: RELATED_IMAGE_BROKER_INIT_2550
+          value: quay.io/arkmq-org/arkmq-org-broker-init@sha256:1251fb2be2a640be3a71dd09339255254c3ffbf49d1059dd9ca7d13348956d6e
```
gtully, line 170: would it make sense to split out the image update into a separate commit?

controllers/broker_controller.go
```diff
@@ -917,7 +998,7 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 
 		// jmx_exporter metrics perms
 		fmt.Fprintln(rbac, "securityRoles.\"mops.mbeanserver.queryMBeans\".metrics.view=true")
-		fmt.Fprintln(rbac, "securityRoles.\"mops.broker\".metrics.view=true") // for query remove filter
```
gtully, line 921: maybe a better comment is  - we need view permission on the broker in order to locate through a query and retrieve it.

controllers/broker_controller.go
```diff
@@ -1114,7 +1209,7 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 	} else {
 		loggerOpts := corev1.EnvVar{
 			Name:  getLoginConfigEnvVarNameForBroker(),
-			Value: "-Dlog4j2.level=INFO",
```
gtully, line 1118: till now, this env var was the only way to control logging on the broker. With the use of the configurationFile property, we may need some way for expose logging config for a service.
But we will need to append this content in that case. It may need some thought.

controllers/broker_controller.go
```diff
@@ -63,6 +70,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cluster"
 )
 
+//go:embed broker_status_script.sh
```
gtully, line 74: that is a neat trick!

gtully (approved):

controllers/broker_controller.go
lavocatt: Actually I don't remember at all why :sweat_smile:
They make sense, and possibly the AI removed them at some point and I failed to catch that

controllers/broker_controller.go
lavocatt: Good catch!

config/rbac/role.yaml
lavocatt: create/delete were already there — they come from the ActiveMQArtemis marker (`pods;services;... create;delete;update`). GitHub showed them as added because pods had been split into their own rule once `patch` made the verb set diverge.

The operator never patches pods after the UUID annotation change (the script does that via the per-CR Role with `resourceNames`). Dropped unused operator `pods/patch`. Pods are back in the core resource list: `create;delete;get;list;update;watch`. Kept `roles;rolebindings` `list;patch;watch` — those are used for the status-script Role.

controllers/broker_controller.go
lavocatt: Agreed. Switched the reload appender to RollingFile (1MB, max 1 backup). `tail -F` already follows rotation. Left `append=false` off — truncating the file races the tailer.

controllers/broker_controller.go
lavocatt: The previous default replaced image logging with a stub `rootLogger`. extraMounts `-logging-config` also skipped the reload fragment entirely.

Now we always mount the operator fragment and use composite Log4j2 config (`-Dlog4j2.configurationFile=base,operator`):

- no extraMounts: `classpath:log4j2.properties` + fragment, keep `-Dlog4j2.level=INFO`
- extraMounts `-logging-config`: user file first, fragment last (no level overlay)
- existing `-Dlog4j2.configurationFile` in `JDK_JAVA_OPTIONS` or `JAVA_ARGS_APPEND`: append the fragment URI to that list instead of adding a second `-D`

The fragment is appender-only (no `rootLogger` / no logger level) so composite merge does not clobber user/image logging.

controllers/broker_controller.go
lavocatt: Applied that wording.

deploy/operator.yaml
lavocatt: No I think we shouldn't split, the work requires a 2550 broker.

lavocatt: it's been extracted now, marking this as resolved.

gtully (commented):
looks good, the composite logging url looks like a good fix.

gtully:
peeking at the Pod watch, i think we can use the WatchesMetadata and PartialObjectMetadata because the annotation and owner ref in part of that. 
but looking more, the drain controller is still in play and that watches Pods, I don't know if metadata is sufficient for it. The move to metadata only would have to be operator wide else we get duplicate cache entries which would be all wrong!. We can investigate WatchesMetadata at some point in the future, I have opened #1491

the mapping via owner ref and ss, I wonder if that is really necessary because we have a naming convention for the ss based on the cr name.

gtully:
I think a cache level filter for pods that just match our controller labels will make sense. we don't want to be tracking all pods in a cluster.

```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
    Cache: cache.Options{
        ByObject: map[client.Object]cache.ByObject{
            &corev1.Pod{}: {
                // Only cache Pods matching these labels
                Label: labels.SelectorFromSet(labels.Set{
                    "app": ....,
                }),
            },
        },
    },
})
```

we will need all values b/c the cache is shared across controllers unless there is a common shared label we can make use of.

brusdev (commented):
The event-driven approach here is a solid win — replacing a timer-based polling loop with a signal from the broker itself is exactly the right model, and the UUID trick to avoid clearing annotation state is clean.

One thing worth considering before this lands: the signalling channel is the most privileged part of the design. The script currently needs patch on its own pod, which is a mutable operation on a live resource. A slightly less privileged alternative would be to POST a core/v1 Event instead — create on events is append-only, so even if the token were stolen the worst outcome is spurious reconcile triggers rather than mutation of any live object. The operator would watch Events filtered by reason: BrokerReloaded and map back to the Broker CR via the event's involvedObject.

This pairs naturally with a dedicated projected SA token to address the automountServiceAccountToken: true change in configPodSecurity, which is a silent regression for anyone relying on the previous false default. Concretely:

- The pod runs with a dedicated SA bound only to create events instead of the default SA
- A projected volume provides a short-lived token that kubelet rotates in place before expiry
- The token is read inside signal_reconcile on every call rather than once at script startup, which also fixes the existing token-staleness bug where the shell assignment is evaluated once and then goes stale after rotation

lavocatt:
> the mapping via owner ref and ss, I wonder if that is really necessary because we have a naming convention for the ss based on the cr name.

We've already been discussing this idea, and I'm once again pushing back on the naming convention. It's more brittle (naming clash with anyone else that happen to use the same name convention) than ancestor matching and it doesn't cost more to check for grandchildren. The informer cache is already populated anyway. We're not asking more data.

lavocatt:
> The event-driven approach here is a solid win — replacing a timer-based polling loop with a signal from the broker itself is exactly the right model, and the UUID trick to avoid clearing annotation state is clean.
> 
> One thing worth considering before this lands: the signalling channel is the most privileged part of the design. The script currently needs patch on its own pod, which is a mutable operation on a live resource. A slightly less privileged alternative would be to POST a core/v1 Event instead — create on events is append-only, so even if the token were stolen the worst outcome is spurious reconcile triggers rather than mutation of any live object. The operator would watch Events filtered by reason: BrokerReloaded and map back to the Broker CR via the event's involvedObject.
> 
> This pairs naturally with a dedicated projected SA token to address the automountServiceAccountToken: true change in configPodSecurity, which is a silent regression for anyone relying on the previous false default. Concretely:
> 
>     * The pod runs with a dedicated SA bound only to create events instead of the default SA
> 
>     * A projected volume provides a short-lived token that kubelet rotates in place before expiry
> 
>     * The token is read inside signal_reconcile on every call rather than once at script startup, which also fixes the existing token-staleness bug where the shell assignment is evaluated once and then goes stale after rotation

**Token staleness** — agreed, fixing. The AUTH variable is evaluated once at script startup. After kubelet rotates the mounted token (~80% of TTL), the cached value goes stale. Moving the cat inside signal_reconcile so it reads a fresh token on every call.

**Events instead of pod PATCH** — disagree. Events are explicitly not a reliable signaling mechanism. The API server garbage-collects them (default 1h TTL) and can drop them under load. A missed Event means ConfigApplied stays false with no recovery path. The annotation PATCH persists on the pod — the operator sees it whenever it next lists/watches, regardless of timing.

The operator already watches pods owned by its StatefulSets. The annotation predicate adds zero new informers. An Event-based approach would require a separate Event informer, filtering by reason, parsing involvedObject, and mapping back to the Broker CR — more moving parts for a weaker guarantee.

On the threat model: the token lives inside the broker container. An attacker with access to it already has full access to the broker process, message data, and the ability to exec arbitrary commands. The ability to PATCH a single annotation on the pod's own metadata (the Role is scoped to resourceNames: ["{cr}-ss-0"], verb patch only) is not an escalation path.  The worst case is spurious reconcile triggers — the same outcome as the Events alternative. Rate limiting on the operator workqueue is the correct mitigation for that, independent of signaling mechanism.

**automountServiceAccountToken: true** — intentional, not a regression. On main, the default path sets automount: false because nothing needed the token. This PR sets it to true because the sidecar script requires it to signal the operator. When podSecurity.serviceAccountName is set, the user controls their own SA and automount policy — that path is unchanged. I'll add a note to the PR description to make this explicit.

A dedicated SA with projected token volume is possible but adds complexity (SA, projected volume spec, token rotation config) for a single-pod CR where the Role is already minimal.

gtully:
> One thing worth considering before this lands: the signalling channel is the most privileged part of the design. The script currently needs patch on its own pod, which is a mutable operation on a live resource. 

I had the same thought, but there is very little that is mutable in a pod, the typical pattern for change  of a pod is delete/add.

gtully:
> > the mapping via owner ref and ss, I wonder if that is really necessary because we have a naming convention for the ss based on the cr name.
> 
> We've already been discussing this idea, and I'm once again pushing back on the naming convention. It's more brittle (naming clash with anyone else that happen to use the same name convention) than ancestor matching and it doesn't cost more to check for grandchildren. The informer cache is already populated anyway. We're not asking more data.

We are currently watching all pods, If we reduce that to a set we label (which I think we need to), then any naming clash will have occurred at creation time. 
We are pulling the pod names from our naming convention and the host name of the service for the jolokia calls so the pattern works.
What you have is safe for sure, but the problem it solves goes away if we limit what we watch.
I agree it is safe and not any additional burden on the cache, but I think could  be simpler and still work, but we need to limit the cache first!

brusdev:
> automountServiceAccountToken: true — intentional, not a regression. On main, the default path sets automount: false because nothing needed the token. This PR sets it to true because the sidecar script requires it to signal the operator. When podSecurity.serviceAccountName is set, the user controls their own SA and automount policy — that path is unchanged. I'll add a note to the PR description to make this explicit.
>
> A dedicated SA with projected token volume is possible but adds complexity (SA, projected volume spec, token rotation config) for a single-pod CR where the Role is already minimal.

Does the current implementation also support non-default service accounts?

controllers/broker_controller.go
```diff
@@ -1150,9 +1257,10 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 
 	reEvalJdkOpts := generateReEvalOrdinaEnvReplacement(customResource.Spec.Env)
 
+	statusScriptLaunch := fmt.Sprintf("bash %s/%s > /dev/null 2>&1 &", statusScriptCMPath, brokerStatusScriptKey)
```
gtully: I was expecting a sidecar container here, a background script will need to be kept alive, that is the job of the pod controller. We can use a minimal image (the base of the broker image) to run the script. It will make the process visible too which is a good thing. https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

gtully (changes requested):
consider a sidcar for the script - we need it to be visible and managed for us b/c we will now depend on it.

controllers/broker_controller.go
lavocatt: Good thinking, now it does exactly that. However not using a UBI9 image since it'll be required to download that image on the machine but instead reusing the broker image. It has close to zero cost for memory & storage and a marginally more security footprint since the image contains more packages that the sidecar actually needs to function. However, it's marginal, since someone who'd break into the pod would have access to the broker container anyway. What's important is to not ship more binaries than the broker actually requires.

controllers/broker_controller.go
```diff
@@ -1155,7 +1249,53 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
...
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("64Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			RunAsNonRoot:             &runAsNonRoot,
+			RunAsUser:                func() *int64 { uid := int64(1000); return &uid }(),
```
gtully: allthough this is independent, I think it may make sense to use the same logic as the broker main container and respect any pod security context. 
Is the volume mount where the log file lives shared with the main container? if so the permissions will need to match.

gtully (commented):
check if we need to reuse the cr pod security context

controllers/broker_controller.go
```diff
@@ -64,6 +71,25 @@ import (
...
+var brokerStatusScript string
+
+const (
+	brokerStatusScriptSuffix  = "-status-script"
+	brokerStatusScriptKey     = "broker-status.sh"
+	brokerLoggingConfigSuffix = "-logging-config"
+	sidecarContainerName      = "broker-status"
+
+	log4j2ConfigurationFileFlag = "-Dlog4j2.configurationFile="
+	defaultLog4j2ConfigURI      = "classpath:log4j2.properties"
```
gtully: the logging config is looking a bit complicated. We now need to configure an appender and we reference that in -Dlog4j2.configurationFile, that may be sufficient.
If we allow a logging configmap extra mount, that should be applied first. I don;t think we need to deal with any other log4j system property.
Does a log4j2.properties exist on the classpath, ie the file referenced from: defaultLog4j2ConfigURI

gtully (changes requested):
I think we are down to the configuration of logging... and settling on the values of -Dlog4j2.configurationFile and we are  done.
Respecting any extra mount maybe the all we need to do

controllers/broker_controller.go
```diff
@@ -64,6 +72,24 @@ import (
...
+//go:embed broker_status_script.sh
+var brokerStatusScript string
+
+const (
+	brokerStatusScriptSuffix    = "-status-script"
+	brokerStatusScriptKey       = "broker-status.sh"
+	brokerLoggingConfigSuffix   = "-logging-config"
+	sidecarContainerName        = "broker-status"
+	log4j2ConfigurationFileFlag = "-Dlog4j2.configurationFile="
+	brokerLog4JDefaultURI       = "/opt/amq/conf/log4j2.properties"
```
gtully: I think we should ignore this file and let the operator do the config of logging for the broker cr, the env var JAVA_ARGS_APPEND is the get out of jail.

gtully (approved):
looks great, this is a winner!

PR #1419, opened 2026-05-28
[#1409] add gateway api support

### LINKED ISSUE #1409: NGINX Ingress Retirement - Gateway API Support?
### Discussed in https://github.com/arkmq-org/arkmq-org-broker-operator/discussions/1406

<div type='discussions-op-text'>

<sup>Originally posted by **leisma** May  5, 2026</sup>
Hi everyone,

the arkmq-org-broker-operator currently creates an NGINX Ingress when exposing the broker and the console. Since the classic Ingress API is being phased out (here the official statement from Kubernetes https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement) in favour of the Kubernetes Gateway API (`HTTPRoute`, `TLSRoute`, etc.), I was wondering if support for these newer resources is planned. I couldn't find any related issues. (sorry if I missed something!)

The Operator already supports OpenShift Routes, but in plain Kubernetes (or Gateway API–based setups), there's currently no alternative to NGINX Ingress. Are there plans to support Gateway API resources alongside existing Ingress and Route option, both for broker and console exposure?</div>

leisma:
Closes: https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1409


gtully (commented):
this looks great. lets see where the checks land.

gtully:
@leisma the ci has a hard coded go version and is failing a prelim test. with a go version update there are a few places that need update. peek at this commit to see the relevant files. https://github.com/arkmq-org/arkmq-org-broker-operator/commit/83bc22bb4d5c5d558c5990d479ff1ffdf1493db1

leisma:
Hey @gtully, I just committed the CI changes that were needed. Hopefully the checks will be green 🙏

gtully:
@leisma add the [#1409] prefix to each commit message or squash to a single commit to get past the first check.

gtully:
expected check failure, the kube api update means lots of extras in the embedded kinds, so the generation filters through and you need to included those changes. use: make build && make generate-deploy && make bundle && make helm-charts  and then git status to see what else needs to be included in the commit

leisma:
> expected check failure, the kube api update means lots of extras in the embedded kinds, so the generation filters through and you need to included those changes. use: make build && make generate-deploy && make bundle && make helm-charts and then git status to see what else needs to be included in the commit

Hopefully it works now. I just had to generate the deployment, bundle, and Helm files. It might be worth mentioning that in the contribution section of the docs, unless I overlooked it.

controllers/suite_test.go
```diff
@@ -55,9 +55,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	crconfig "sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
```
gtully: should this not be v1?

gtully:
with my existing minikube, after installing the gateway crds with: kubectl apply --server-side -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.0/standard-install.yaml
I was expecting the new test 	It("exposes acceptors and console via Gateway API routes",  to work. For some reason if fails with an invalid host name... i will need to investigate a bit.
I am asuming it is possible to validate gateway related kind creation etc, with out an actual gateway impl deployed. Is that the intent?

see if you can get a full run of: make test-mk-v with a local minikube of other kube cluster.

gtully:
or even ginkgo --json-report ./ginkgo.report -focus "artemis controller tls secret reuse exposes acceptors and console via Gateway API routes"   -r

gtully:
just for reference, the reconcile error is:

2026-06-04T20:43:33.729+0100    ERROR   Reconciler error        {"controller": "broker", "controllerGroup": "broker.arkmq.org", "controllerKind": "Broker", "Broker": {"name":"aa-controller314a","namespace":"test"}, "namespace": "test", "name": "aa-controller314a", "reconcileID": "e7127303-b451-4487-8087-951382c2d242", "error": "[\"failed to create new *v1.TLSRoute, TLSRoute.gateway.networking.k8s.io \\\"aa-controller314a-amqp-ssl-gw-0-svc-tlsrte\\\" is invalid: [spec.hostnames[0]: Invalid value: \\\"\\\": spec.hostnames[0] in body should be at least 1 chars long, spec.hostnames: Invalid value: Hostnames must be valid based on RFC-1123]\" \"failed to create new *v1.TLSRoute, TLSRoute.gateway.networking.k8s.io \\\"aa-controller314a-wconsj-0-svc-tlsrte\\\" is invalid: [spec.hostnames[0]: Invalid value: \\\"\\\": spec.hostnames[0] in body should be at least 1 chars long, spec.hostnames: Invalid value: Hostnames must be valid based on RFC-1123]\" \"failed to create new *v1.HTTPRoute, HTTPRoute.gateway.networking.k8s.io \\\"aa-controller314a-amqp-plain-gw-0-svc-httprte\\\" is invalid: spec.hostnames[0]: Invalid value: \\\"\\\": spec.hostnames[0] in body should be at least 1 chars long\"]"}

gtully:
the default value "" for ingressHost in the acceptor is being used as is in error. The validation seems to suggest that a domain is sufficient, then we need to provide a matched value for ingresSHost when we make the TLSRoute

gtully (changes requested):
we need to check for empty ingressHost in the TLSRoute generation.

leisma:
> we need to check for empty ingressHost in the TLSRoute generation.

Hey @gtully, sorry, I'm a bit busy right now. I'll take a look later or over the weekend. But thank you for the review!

gtully:
> > we need to check for empty ingressHost in the TLSRoute generation.
> 
> Hey @gtully, sorry, I'm a bit busy right now. I'll take a look later or over the weekend. But thank you for the review!

no worries @leisma , thanks for your contribution. there is no rush!

controllers/suite_test.go
leisma: Yes you are ritght, I just fixed it.

leisma:
Hey @gtully I fixed the issue with the `ingressHost` and with that did some more tests and noticed some reconcilation issues, that I also fixed. Locally i managed to run the e2e tests with make test-mk-v on my local minikube - with and without installed Gateway API CRDs.

gtully:
great, thanks. It looks like a rebase against main in the last step, there have been some clashing commits that need to manual intervention

Makefile
```diff
@@ -134,8 +140,8 @@ manifests: controller-gen
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
-	sed -i 's~^LABEL version=.*~LABEL version="$(VERSION)"~g' Dockerfile
-	sed -i 's~\tVersion = ".*"~\tVersion = "$(MANAGER_VERSION)"~' version/version.go
+	sed -i.bak 's~^LABEL version=.*~LABEL version="$(VERSION)"~g' Dockerfile && rm Dockerfile.bak
```
gtully, line 143: any reason to leave the  .bak generation?

AGENTS.md
```diff
@@ -251,13 +253,46 @@ kubectl get deployment ingress-nginx-controller -n ingress-nginx \
...
+**macOS only — make the ingress reachable on localhost**
+
+The E2E suite dials `clusterIngressHost:443`, where `clusterIngressHost` is the
+hostname of the API server URL (`controllers/suite_test.go`). With the docker
+driver on macOS that resolves to `127.0.0.1`, and nothing listens there — on
+Linux it resolves to the minikube node IP, where the ingress controller is
+reachable directly, which is why CI needs no extra step.
+
+Run the tunnel in a **separate terminal** and leave it open (it blocks and
+needs root for port 443):
```
gtully, line 265: redflag on giving root to an ai, leave it as requiring a tunnel.

controllers/suite_test.go
```diff
@@ -489,8 +495,12 @@ func createControllerManager(disableMetrics bool, watchNamespace string) {
 
 	managerCtx, managerCancel = context.WithCancel(ctx)
 
+	skipNameValidation := true
 	mgrOptions := ctrl.Options{
 		Scheme: scheme.Scheme,
+		Controller: crconfig.Controller{
+			SkipNameValidation: &skipNameValidation,
```
gtully, line 502: why is this necessary now?

gtully (commented):
this is coming together nicely, great work. Some small comments inline

Makefile
leisma: Yes, BSD sed (macOS) requires -i to take a suffix argument. With a bare -i it consumes the expression as the suffix and then treats the filename as the script . So make generate fails on macOS and leaves the file untouched. sed -i '' fixes macOS but breaks GNU sed, so -i.bak … && rm *.bak is the only spelling that works on both. It's also the existing convention in this Makefile; the bundle target has used it at lines 342-349 since before this PR.
 
 But I agree, it has nothing to do with the feature that is implemented in this PR. Should I revert it?

controllers/suite_test.go
leisma: It's required by the controller-runtime bump in this PR (v0.16.3 → v0.24.1). Since v0.19, controller-runtime enforces process-wide unique controller names via pkg/controller/name.go:                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                     
```
  if usedNames.Has(name) {                                                                                                                                                                                                                                                                                                                                                           
      return fmt.Errorf("controller with name %s already exists. Controller names must be unique "+                                                                                                                                                                                                                                                                                  
          "to avoid multiple controllers reporting the same metric. ...", name)                                                                                                                                                                                                                                                                                                      
  }    
```                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                     
The suite restarts the manager repeatedly in the same process, so from the second createControllerManager call on, the same names are registered again: activemqartemis, brokercluster and broker fail to register (the error is only logged, so they end up silently unregistered) and the spec then hangs until the test timeout.         
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
SkipNameValidation is set in test code only, main.go leaves the validation on. The alternative, generating unique controller names per manager restart, would make test names diverge from production and touch every controller's SetupWithManager, so this option seemed the smaller change and ok for me for test cases.

AGENTS.md
leisma: I only left the hint about the need of adding a tunnel to minicube on mac.

PR #1426, opened 2026-06-11
[#1334] Dev.latest build references dev.latest operand snapshots

lavocatt:
CI resolves dev.latest operand tags to sha256 digests via skopeo and
    injects DEFAULT_BROKER_VERSION=0.0.0, DEFAULT_BROKER_COMPACT_VERSION,
    DEFAULT_BROKER_KUBE_IMAGE, DEFAULT_BROKER_INIT_IMAGE, and
    RELATED_IMAGE_BROKER_*_000 env vars into the dev helm chart, bundle
    CSV, and kustomize overlay.
    
    At startup, if DEFAULT_BROKER_VERSION is 0.0.0, the operator registers
    version 0.0.0 in the supported version maps. ResolveBrokerVersion uses
    GetDefaultVersion() for the no-version-specified case, so a default CR
    resolves to 0.0.0 and picks up the dev.latest operand digests. CRs
    that pin an explicit version are unaffected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,8 +241,22 @@ jobs:
           podman login --username=${{ secrets.QUAY_USERNAME }} --password=${{ secrets.QUAY_PASSWORD }} quay.io
           podman manifest push $IMAGE_NAME:dev.latest docker://quay.io/${{ secrets.QUAY_NAMESPACE }}/$IMAGE_NAME:dev.latest
 
+      - name: Resolve dev.latest operand digests
+        if: ${{ github.event_name == 'push' }}
+        id: operand-digests
+        run: |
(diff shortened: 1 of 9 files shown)


brusdev (commented):
This is a great improvement! It allows users to test the dev.latest operands easily without having to hack the source code. I really like the idea of using an environment variable for this. Is it possible to reuse one of our existing env vars for this purpose?

- DEFAULT_BROKER_VERSION
- DEFAULT_BROKER_COMPACT_VERSION
- DEFAULT_BROKER_KUBE_IMAGE
- DEFAULT_BROKER_INIT_IMAGE

brusdev:
We need to swap out version 0.0.0 on the snapshot operand for the real version. Since 0.0.0 doesn't match any valid Apache Artemis release, it should trigger a BrokerVersionAligned condition failure with the VersionMismatch reason.

brusdev:
We should apply the snapshot tag to container images pointing to snapshot operands.

Issue #1433, opened 2026-06-18
Split out the restricted mode artemis cr into its own controller


gtully:
With https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1431 we will free up the Broker cr name. We should make use of that for the restricted mode CR. Strip it down to the minimum and pull out the conditional logic of the restricted mode boolean of the current controller into a new broker controller. We can follow the templates from the app and service cr. There are too many changes in behaviour when restricted=true that make the switch untenable and the shared code base will need to evolve independently as restricted is rolled out by the broker service.

It may make sense to pull the new Broker, App and Service CRs into the v1 group, I don't see any good reason to retain the v1beta2 tag. That can be a followup.

PR #1435, opened 2026-06-23
 [#1433] Extract restricted mode as as Broker CR

lavocatt:
BrokerCluster is now unconditionally non-restricted. All restricted-mode
    tests and tutorials moved to use the Broker CR.
    
    Remove the Restricted flag from Broker spec. Broker always runs embedded
    Java with mTLS Jolokia, no init containers, no clustering.


gtully:
this is a great simplification and the CR names better match intent which is a big win. v1beta2 is perfect for users migrating from ActivemqArtemisCr to BrokerCluster, but we may want to introduce v1 for Broker,BrokerService and BrokerApp.

.cursor/plans/extract_broker_cr_phases_3a0e3b8f.plan.md
```diff
@@ -0,0 +1,256 @@
+---
```
lavocatt: Ah, I left the plan, will remove it soon

lavocatt:
This is a first step, this involves extracting the restricted mode into its own CR. I have intentionally duplicated code between the BrokerCluster and the Broker to minimize the diff against main.

A next step will require to extract common code into a suite of packages. It will also be the opportunity to break down common.go into sub packages with specific intentes:
- certs
- brokerProperties
- persistenceVolume
- etc

A next step will enable the go linter in the ci as well.

Sidecar is orthogonal to these, and should probably be the actual next step

gtully:
I notice that the deploymentPlan remains shared. It will take a good bit of thought full work to replace it in the Broker with a BrokerSpec that has just the minimal we need. We will need to revisit the intent of each attribute to determine if it adds value and makes sense. This can be a next step.

gtully (approved):
This is an important first step. There is a great opportunity to produce a simple Broker CR that is very focused on secure operation vs configuration. Each attribute that we remove will need careful consideration and this split allows us to start on that.

gtully:
is there any impediment to going to v1 for broker/app/service ?

lavocatt:
> is there any impediment to going to v1 for broker/app/service ?

I don't think there is, but wondering if this shouldn't be part as a follow up as well.

IMO this PR shouldn't change service and app and simply extract Broker.

gtully:
The work can continue in https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1439

Issue #1440, opened 2026-07-08
Add golangci-lint with ratchet mechanism to enforce lint quality over time


lavocatt:
# Add golangci-lint with ratchet mechanism to enforce lint quality over time

## Problem

The project has no unified static analysis standard. CI only enforces `go vet -composites=false` and `go fmt`. Lint behavior depends entirely on each contributor's editor:

- Neovim with gopls+staticcheck shows hundreds of warnings
- VS Code/Cursor with default settings shows none
- Nothing gates lint quality at merge time

This means lint debt accumulates silently and contributors have no shared contract on code quality.

## Proposal

Introduce `golangci-lint` as the single lint authority, enforced in CI with a two-layer ratchet that ensures lint issues can only decrease over time.

### 1. Add `.golangci.yml`

Define which linters are enabled. The specific set should be determined by running golangci-lint on the current codebase and evaluating signal-to-noise ratio. Preserve the existing `composites=false` govet setting. This file becomes the shared source of truth for CI, Neovim, and Cursor/VS Code.

### 2. Add a Makefile target

```makefile
lint:
	golangci-lint run ./...
```

### 3. Add `.vscode/settings.json`

```json
{
  "go.lintTool": "golangci-lint",
  "go.lintFlags": ["--fast"]
}
```

This ensures Cursor and VS Code users see the same findings as CI without manual configuration.

#### Layer 1: Hard gate on new code (zero tolerance)

Run `golangci-lint run --new-from-rev=$(git merge-base HEAD origin/main)`. This must produce **zero findings**. Contributors cannot introduce new lint issues in lines they touch.

#### Layer 2: Ratchet on total codebase debt

1. Checkout the merge-base commit, run `golangci-lint run ./...`, count total findings → `count_before`
2. Checkout PR HEAD, run `golangci-lint run ./...`, count total findings → `count_after`
3. Decision:
   - `count_after > count_before` → **FAIL**. The PR increased total lint debt. (already covered by layer 1)
   - `count_after == count_before` → **WARNING**. Print: "You touched the codebase and left the same amount of lint debt. Consider fixing a nearby issue."
   - `count_after < count_before` → **PASS**. Print: "Lint debt decreased from N to M. Thank you."

`echo "::warning::Lint debt unchanged ($count_before → $count_after). Consider fixing a nearby issue."`

#### Skip condition

[...]

PR #1442, opened 2026-07-10
Fix issue #1440: Add golangci-lint with ratchet mechanism to enforce lint quality over time

Robmurt4:
Fix issue #1440: Add golangci-lint with ratchet mechanism to enforce lint quality over time

- Added lint command to makefile. 
- Excluded '.vscode/settings.json' from .gitignore and added lines to ensureCursor and VS Code users see the same findings as CI without manual configuration.
- Made changes to '.github/workflows/ci.yaml' to allow for two layer CI enforcement. 
- Added golang ci yaml file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,58 @@ jobs:
             fi
           done
 
+  lint:
+    name: Lint
+    if: ${{ github.event_name == 'pull_request' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: '1.25.9'
+          cache: true
+
+      - name: Install golangci-lint
+        run: make lint-install
+
+      - name: Layer 1 — hard gate on new code
+        run: |
+          MERGE_BASE=$(git merge-base HEAD origin/main)
+          ./bin/golangci-lint run --new-from-rev=${MERGE_BASE} ./...
+
+      - name: Layer 2 — ratchet on total codebase debt
+        run: |
+          # Skip entirely if no .go files changed
+          MERGE_BASE=$(git merge-base HEAD origin/main)
+          if ! git diff --name-only ${MERGE_BASE} HEAD | grep -q '\.go$'; then
+            echo "No .go files changed — skipping debt ratchet."
+            exit 0
+          fi
+
+          # Count findings at merge-base
+          git checkout --quiet ${MERGE_BASE}
+          COUNT_BEFORE=$(golangci-lint run ./... 2>/dev/null | grep -c ' (errcheck)\| (ineffassign)\| (staticcheck)\| (unused)' || true)
+          git checkout --quiet -
+
+          # Count findings at PR HEAD
+          COUNT_AFTER=$(golangci-lint run ./... 2>/dev/null | grep -c ' (errcheck)\| (ineffassign)\| (staticcheck)\| (unused)' || true)
+
+          echo "Lint debt: before=${COUNT_BEFORE} after=${COUNT_AFTER}"
+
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -145,6 +145,9 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet -composites=false ./...
 
+.PHONY: lint
+lint: ## Run golangci-lint against code.
+	golangci-lint run ./...
```
gtully: This will want to use an executable from localbin that is downloaded if not present. See the pattern for KUSTOMIZE or ENVTEST

.vscode/settings.json
gtully, line 1: probably best to just document what is needed to add to settings for vscode integration rather than provide a . directory. Even just add a link to the -lint docs so that it does not get stale as a reference.

.github/workflows/ci.yml
```diff
@@ -60,6 +60,60 @@ jobs:
...
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: '1.25.9'
+          cache: true
+
+      - name: Install golangci-lint
```
gtully, line 79: possibly the download to localbin bit of the new make lint target can be reused here to do the install?

gtully (changes requested):

Makefile
Robmurt4: I have edited the Makefile to do this as requested.

.vscode/settings.json
Robmurt4: Is there a specific place I should add this documentation?

.github/workflows/ci.yml
Robmurt4: As requested, now command "make lint-install" is run to reuse the localbin code from the make file.

.vscode/settings.json
Robmurt4: Removed .vscode/settings.json, added lines specifiying how to add this in ./docs/help/building.md, and removed file and line `!.vscode/settings.json` n the .gitignore.

Makefile
```diff
@@ -290,8 +291,26 @@ $(KUSTOMIZE): $(LOCALBIN)
...
+	echo "curl -Ss ${LINT_INSTALL_SCRIPT} | bash -s -- $(subst v,,$(LINT_VERSION)) $(LOCALBIN);"
+	test -s $(LOCALBIN)/golangci-lint || { curl -Ss ${LINT_INSTALL_SCRIPT} | bash -s -- v$(subst v,,$(LINT_VERSION)) $(LOCALBIN); }	
+	
+
+lint: lint-install
+	$(LOCALBIN)/golangci-lint run ./...
+
 .PHONY: controller-gen
-controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+controller-gen: $(CONTROLLER_G-EN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
```
gtully: that looks like a typo

Robmurt4: Yes this was a typo. Has now been fixed.

gtully:
it would be  a good idea to squash the commits into one.

gtully (approved):
this looks good. 
I am wondering about the .golangci.yaml, can we run with more default config or is there a good reason for each piece of config.
Is the yaml picked up by the make lint target?  Giving it a run, there looks like there is lots of work to do.

Robmurt4:
> this looks good. I am wondering about the .golangci.yaml, can we run with more default config or is there a good reason for each piece of config. Is the yaml picked up by the make lint target? Giving it a run, there looks like there is lots of work to do.

The yaml is picked up, 'golangci-lint run" automatically searches for a .golangci.yml file in the hierarchy. 

My main reasons for the configuration were to focus on bugs and correctness. Exclusions are manly style related. I can spend a bit more time looking at it though.

PR #1443, opened 2026-07-17
[#1427] Use Broker label for managed resources created by Broker CRs

### LINKED ISSUE #1427: Use Broker label for managed resources created by Broker CRs
**Describe the feature**
Currently, the ArkMQ Broker Operator applies the ActiveMQArtemis label to all managed resources. While this is appropriate for resources generated by the deprecated ActiveMQArtemis Custom Resources (CRs), it should not be applied to resources generated by the newer Broker CRs.

Managed resources should dynamically reflect the type of CR that created them to prevent confusion and ensure proper resource tracking.

Current Behavior
All managed resources, regardless of whether they are created via an ActiveMQArtemis CR or a Broker CR, are tagged with the ActiveMQArtemis label.

Expected Behavior
Deprecated ActiveMQArtemis CRs: Managed resources should continue to use the ActiveMQArtemis label to maintain backwards compatibility.

New Broker CRs: Managed resources must use the Broker label.

> [!TIP]

Msarawan:
Fixes: [issue#1427](https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1427)

### DIFF
--- a/pkg/utils/selectors/label.go
+++ b/pkg/utils/selectors/label.go
@@ -3,19 +3,22 @@ package selectors
 const (
 	LabelAppKey      = "application"
 	LabelResourceKey = "ActiveMQArtemis"
+	LabelBrokerKey   = "Broker"
 )
 
 type LabelerInterface interface {
 	Labels() map[string]string
 	Base(baseName string) *LabelerData
 	Suffix(labelSuffix string) *LabelerData
+	ResourceKey(key string) *LabelerData
 	Generate()
 }
 
 type LabelerData struct {
-	baseName string
-	suffix   string
-	labels   map[string]string
+	baseName    string
+	suffix      string
+	resourceKey string
+	labels      map[string]string
 }
 
 func (l *LabelerData) Labels() map[string]string {
@@ -32,10 +35,21 @@ func (l *LabelerData) Suffix(labelSuffix string) *LabelerData {
 	return l
 }
 
+// ResourceKey sets the tracking label key. Defaults to LabelResourceKey ("ActiveMQArtemis")
+// when not called. Use LabelBrokerKey ("Broker") for Broker CRs.
+func (l *LabelerData) ResourceKey(key string) *LabelerData {
+	l.resourceKey = key
+	return l
+}
+
 func (l *LabelerData) Generate() {
+	key := l.resourceKey
+	if key == "" {
+		key = LabelResourceKey
+	}
 	l.labels = make(map[string]string)
 	l.labels[LabelAppKey] = l.baseName + "-" + l.suffix //"-app"
-	l.labels[LabelResourceKey] = l.baseName
+	l.labels[key] = l.baseName
 }
 
 func GetLabels(crName string) map[string]string {
--- a/pkg/utils/selectors/label.go
+++ b/pkg/utils/selectors/label.go
@@ -1,8 +1,9 @@
 package selectors
 
 const (
-	LabelAppKey      = "application"
-	LabelResourceKey = "ActiveMQArtemis"
+	LabelAppKey             = "application"
+	LabelActiveMQArtemisKey = "ActiveMQArtemis"
+	LabelBrokerKey          = "broker"
 )
 
 type LabelerInterface interface {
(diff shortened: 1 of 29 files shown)


lavocatt (changes requested):
Thanks @Msarawan , the behavioral change is correct however I don't necessarily agree with how the API is shaped:

My concern is with the `label.go` changes: `ResourceKey(string)` accepts any string,
but there are exactly two valid tracking keys in this codebase. This is a free-form
API in a two-value domain, and the most dangerous invalid value (`LabelAppKey`)
silently corrupts the output — `Generate()` overwrites the `application` label and
produces no tracking label. No compile error, no panic, just wrong selectors at runtime.

The implicit default (forget to call `ResourceKey()` → silently get `ActiveMQArtemis`)
adds another failure mode: a future CRD type author has to *know* to call it.

I'd prefer we close both holes mechanically rather than relying on callers getting it
right. Two constructor functions would do it:

```go
  func NewBrokerLabeler() *LabelerData {
      return &LabelerData{resourceKey: LabelBrokerKey}
  }

  func NewActiveMQArtemisLabeler() *LabelerData {
      return &LabelerData{resourceKey: LabelResourceKey}
  }
```

The choice gets made at construction and the type system ensures you can't skip it.
Callers become: `NewBrokerLabeler().Base(name).Suffix("app").Generate().`

Minor: the E2E test redeclares "Broker" and "ActiveMQArtemis" as local string
constants instead of using selectors.LabelBrokerKey / selectors.LabelResourceKey.
If the package constants change, the test silently checks stale values.

pkg/utils/selectors/label.go
```diff
@@ -3,6 +3,7 @@ package selectors
 const (
 	LabelAppKey      = "application"
 	LabelResourceKey = "ActiveMQArtemis"
+	LabelBrokerKey   = "Broker"
```
brusdev: As per the Kubernetes docs on object names (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), we can use label names with just lowercase alphanumerics and dashes. So, something like `broker` works, or we can use a prefix like `broker.arkmq.org/name`.

pkg/utils/selectors/label.go
```diff
@@ -3,6 +3,7 @@ package selectors
 const (
 	LabelAppKey      = "application"
 	LabelResourceKey = "ActiveMQArtemis"
```
brusdev: How about renaming LabelResourceKey to LabelActiveMQArtemisKey to make its intent clearer?

brusdev (commented):
LGTM, the implementation is clean, and the unit tests provide solid coverage for the changes.

pkg/utils/selectors/label.go
Msarawan: makes sense, updating it now. thank you!

gtully:
We seem to use the cr.Name as the value of two labels, I would like a common label for all pods that the controller can make, that would reduce the watch cache load.

We set application and Broker or Artemis. Could we use application as  the common k/v 
 application: broker.arkmq.org

and then add broker.arkmq.org/broker or broker.arkmq.org/cluster using the cr name as appropriate.
 
Would that work?

With this pattern we should be able to limit what is in the controller watch cache using a label match of application: broker.arkmq.org (I think)

Msarawan:
> We seem to use the cr.Name as the value of two labels, I would like a common label for all pods that the controller can make, that would reduce the watch cache load.
> 
> We set application and Broker or Artemis. Could we use application as the common k/v application: broker.arkmq.org
> 
> and then add broker.arkmq.org/broker or broker.arkmq.org/cluster using the cr name as appropriate.
> 
> Would that work?
> 
> With this pattern we should be able to limit what is in the controller watch cache using a label match of application: broker.arkmq.org (I think)

@gtully , Thanks for your input — I agree the pattern of a common application: `broker.arkmq.org` label plus domain-specific keys (broker.arkmq.org/broker / broker.arkmq.org/cluster) with the CR name on a single label would be cleaner and would help limit what we put in the controller watch cache.

But, for [issue#1427](https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1427) the scope is specifically to use the correct tracking label per CR type (broker for Broker CRs, ActiveMQArtemis for deprecated ActiveMQArtemis/BrokerCluster). I’d like to keep this PR focused on that so we can land the fix without a larger refactor.

I’ll open a follow-up issue to track the unified labeling scheme and watch-cache filtering. Does that work for you?

brusdev:
I created a follow-up issue from the @gtully's comment: #1495

brusdev (approved):

Issue #1446, opened 2026-07-22
ProducerOf queues missing from Prometheus metrics collection


viswanathannachiappan:
**Describe the bug**
The operator's processControlPlaneOverrideSecret() method only collects ConsumerOf queues for Prometheus metrics, ignoring ProducerOf queues. This results in incomplete metrics visibility in Grafana for producer-only queues.

Expected Behavior
When a BrokerApp declares queues in ProducerOf, those queues should appear in the Prometheus metrics endpoint (/metrics) and subsequent Grafana dashboards.

Actual Behavior
Only queues declared in ConsumerOf appear in the Prometheus metrics configuration. Queues declared only in ProducerOf are omitted from the metrics whitelist.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #1447, opened 2026-07-22
[#1446] Included ProducerOf queues in Prometheus metrics collection

viswanathannachiappan:


gtully (dismissed):
it would be great to include a test modification that will protect this change into the future

gtully (approved):
looks good, the test nicely protects the change

Issue #1423, opened 2026-06-09
Include namespace in broker controller and reconciler logs


brusdev:
**Describe the feature**
Most of the log messages from the broker controller/reconciler only include the CR name. When the operator is installed cluster-wide, the CR name alone is not sufficient to uniquely identify the resource.

> [!TIP]
> Vote this issue reacting with :+1: or :-1:

PR #1451, opened 2026-07-28
[#1423] Included namespace in broker contl and reconciler logs

makella19:


controllers/brokercluster_reconciler.go
```diff
@@ -669,7 +669,7 @@ func (reconciler *BrokerClusterReconcilerImpl) syncMessageMigration(customResour
 			reconciler.log.V(2).Info("Won't set up scaledown for non persistent deployment")
 			return
 		}
-		reconciler.log.V(2).Info("we need scaledown for this cr", "crName", customResource.Name, "scheme", scheme)
+		reconciler.log.V(2).Info("we need scaledown for this cr", "crName", customResource.Name, "crNamespace", customResource.Namespace, "scheme", scheme)
```
lavocatt: This instance uses `crNamespace` and not `CRD.Namespace`

controllers/brokercluster_reconciler.go
```diff
@@ -2130,7 +2130,7 @@ func MakeContainerPorts(cr *v1beta2.BrokerCluster) []corev1.ContainerPort {
 
 func (reconciler *BrokerClusterReconcilerImpl) PodTemplateSpecForCR(customResource *v1beta2.BrokerCluster, namer common.Namers, currentStatefulSet *appsv1.StatefulSet, client rtclient.Client) (*corev1.PodTemplateSpec, error) {
 
-	reqLogger := reconciler.log.WithName(customResource.Name)
+	reqLogger := reconciler.log.WithValues("CRD.Name", customResource.Name, "CRD.Namespace", customResource.Namespace)
```
lavocatt: Beware, here you changed withName to withValues

controllers/broker_controller.go
```diff
@@ -313,7 +313,7 @@ func (reconciler *BrokerReconcilerImpl) getFromDeployed(kind reflect.Type, name
 
 func (reconciler *BrokerReconcilerImpl) ProcessStatefulSet(customResource *v1beta2.Broker, namer common.Namers, client rtclient.Client) (*appsv1.StatefulSet, error) {
 
-	reqLogger := reconciler.log.WithName(customResource.Name)
+	reqLogger := reconciler.log.WithValues("CRD.Name", customResource.Name, "CRD.Namespace", customResource.Namespace)
```
lavocatt: Beware, here you changed withName to withValues

controllers/broker_controller.go
```diff
@@ -873,7 +873,7 @@ func MakeContainerPortsForBroker(cr *v1beta2.Broker) []corev1.ContainerPort {
 
 func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1beta2.Broker, namer common.Namers, currentStatefulSet *appsv1.StatefulSet, client rtclient.Client) (*corev1.PodTemplateSpec, error) {
 
-	reqLogger := reconciler.log.WithName(customResource.Name)
+	reqLogger := reconciler.log.WithValues("CRD.Name", customResource.Name, "CRD.Namespace", customResource.Namespace)
```
lavocatt: Beware, here you changed withName to withValues

lavocatt (commented):
The commit addresses the right problem but takes a suboptimal approach — it manually adds "CRD.Namespace" to individual `reqLogger` / `WithValues` call sites (21 lines), while leaving ~115 other `reconciler.log` calls in the same two files still without namespace context.

A simpler and complete fix: add the namespace to the base logger at construction time in both `NewBrokerReconcilerImpl` and `NewBrokerClusterReconcilerImpl`:

log: `parent.log.WithValues("CRD.Name", customResource.Name, "CRD.Namespace", customResource.Namespace)`,

This is a one-line change per constructor, and every `reconciler.log` call in both files inherits namespace automatically — no need to touch individual call sites at all.

Given that both constructors are structurally identical in how they initialize the logger; this should be extracted into a shared helper to avoid duplicating the pattern and ensure consistency if the log context evolves later.

The per-site `reqLogger := reconciler.log.WithValues("CRD.Name", ..., "CRD.Namespace", ...)` declarations introduced by this commit would then become redundant (re-adding keys already present on the base logger). They can be dropped.

Note: the standalone `ctrl.Log.WithValues(...)` calls in package-level functions (`AssertBrokersAvailable`, `AssertBrokersAvailableForBroker`, etc.) still need explicit `WithValues` since they don't go through `reconciler.log`.

controllers/broker_controller.go
lavocatt: There's duplicate CRD.Name now that we have a properly configured context.

If you want to clean it up that would be great.

controllers/brokercluster_reconciler.go
lavocatt: There's duplicate CRD.Name now that we have a properly configured context.

If you want to clean it up that would be great.

lavocatt (approved):
LGTM thanks

Issue #1439, opened 2026-07-03
Broker CR refactor


lavocatt:
- [x] Filter unused fields from Broker CR
  - [x] deployment plan
  - [x] Image
  - [x] etc
- [ ] Extract common code into a suite of packages.
  - [ ]     certs
  - [ ]     brokerProperties
  - [x]     persistenceVolume
  - [ ]     etc
- [ ] Going to v1 for broker/app/service

lavocatt:
@makella19:

Following up on #1472 (brokerProperties extraction), two more packages would complete the extraction of inline config generation from broker_controller.go and brokerservice_controller.go:

JVM/JAAS config files currently built inline with fmt.Fprintf:

- [ ] _security.config (JVM security properties)
- [ ] login.config (JAAS realm definition)
- [ ] cert user/role mappings (consumed by TextFileCertificateLoginModule)
- [ ] _cert.pemcfg (PEM keystore provider config)

 observability config files:

- [ ] _jolokia.config (Jolokia agent properties)
- [ ] _prometheus_exporter_config (JMX exporter YAML)


Scope: broker_controller.go and brokerservice_controller.go only

PR #1453, opened 2026-08-06
[#1439] Broker CR refactor - Extract common code into suite of pkgs -…

makella19:
… Volumes

### DIFF
--- a/controllers/broker_controller.go
+++ b/controllers/broker_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/RHsyseng/operator-utils/pkg/resource/compare"
 	brokerstatus "github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/broker/status"
 	brokerversion "github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/broker/version"
+	"github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/brokervolumes"
 	"github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/resources"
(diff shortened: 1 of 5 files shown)


lavocatt (commented):
The extraction is clean and the simplifications (dropping unused `namer`, removing the always-nil errors) are good calls.

One concern: `BrokerCluster` still carries its own copies of `MakeVolumes`, `MakeExtraVolumeMounts`, `MakeVolumeMounts`, and `checkExistingPersistentVolumes`. The base logic is identical — the only differences are the field path (`cr.Spec.X` vs `cr.Spec.DeploymentPlan.X`) and BrokerCluster layering SSL acceptor/connector/console volumes on top.

Since this is setting the pattern for the rest of #1439, consider having the shared functions take plain data instead of a CR type:

```go
func BaseVolumes(name string, persistent bool, extras []corev1.Volume, epvcs []VolumeClaimTemplate) []corev1.Volume

func ExtraVolumeMounts(extras []corev1.Volume, mounts []corev1.VolumeMount, epvcs []VolumeClaimTemplate) []corev1.VolumeMount
```

Both reconcilers call these with their own field paths. BrokerCluster appends its SSL volumes on top. Same idea for `checkExistingPersistentVolumes` — extract the per-PVC owner ref removal, let BrokerCluster wrap it in its ordinal loop.

This way the package stays CR-agnostic and you avoid inheriting two diverging copies as the refactor continues.

Also: no tests for the new package yet — the existing tests in `activemqartemis_controller_unit_test.go` cover the BrokerCluster version that's still inline.

lavocatt (approved):
Great work, thanks!

Issue #1454, opened 2026-08-06
Make pub/sub scale tutorial


Robmurt4:
- Create a pub/sub scale tutorial based on the [ginkgo test](https://github.com/arkmq-org/arkmq-org-broker-operator/blob/main/controllers/activemqartemis_pub_sub_scale_test.go). 
- The tutorial should be runnable using the [Markdown Runner](https://github.com/arkmq-org/markdown-runner)
- Further enhancements -> getting the brokers to mesh.

PR #1455, opened 2026-08-06
[#1454] Make pub/sub scale tutorial

Robmurt4:
- [#1454] Make Pub/Sub scale tutorial.
- Created 'docs/tutuorials/pub_sub_scale.md' that is a tutorial file that can be run with Markdown-Runner based on the [ginkgo test](https://github.com/arkmq-org/arkmq-org-broker-operator/blob/main/controllers/activemqartemis_pub_sub_scale_test.go).


docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,510 @@
...
+    matchLabels:
+      app: consumer3
+  template:
+    metadata:
+      labels:
+        app: consumer3
+    spec:
+      containers:
+        - name: consumer3
+          image: quay.io/arkmq-org/arkmq-org-broker-kubernetes:artemis.2.54.0
```
lavocatt: cat <<EOT >> deploy.yml
        image: quay.io/arkmq-org/arkmq-org-broker-kubernetes:artemis.${BROKER_VERSION}
EOT

You could make the consumer and producer use the same broker version as the one deployed, automatically.

See the prometheus locked down tutorial.

docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,510 @@
...
+spec:
+  selector:
+    ActiveMQArtemis: pub-sub-broker
+  ports:
+    - port: 62616
+      targetPort: 61616
+EOF
+````
+
+> A different port (`62616`) is used on the Service to avoid conflicts with the per-pod Services that the operator creates on the same cluster IP range. The broker's acceptor remains on `61616` inside the pod.
```
lavocatt: I find it surprising that this would be required. Unless we're exposing the broker's services via an ingress, in which case, that would certainly be a problem. If these are only cluster-internal services I don't see how a clash could occur. But maybe I'm wrong!

docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,510 @@
...
+
+---
+
+### Step 5 — Deploy the consumers
+
+Each consumer runs the `artemis perf consumer` command in a retry loop. The loop is needed because the connection router **refuses** connections that arrive on the wrong broker pod with an AMQP error — the consumer exits, sleeps 1 second, and retries until it lands on the correct one.
```
lavocatt: If I understand correctly, a customer connects to the load balancer and every connection attempt has a 50/50 chance to land on the correct broker?

lavocatt (commented):
This is a very interesting tutorial to read. Thanks!

lavocatt:
IMO, there's no need to keep and maintain 3 tutorials to present prometheus, mtls certs and app & services

We should think about merging:
- service_app_crd_round_trip.md
- prometheus_locked_down.md

The one you are working on.

My thinking is that the one you are working on is a super set of the two others, and, as we don't want to advocate deploying Broker CRD manually instead of using service & app, I think this should be the logical next step

Robmurt4:
> IMO, there's no need to keep and maintain 3 tutorials to present prometheus, mtls certs and app & services
> 
> We should think about merging:
> 
>     * service_app_crd_round_trip.md
> 
>     * prometheus_locked_down.md
> 
> 
> The one you are working on.
> 
> My thinking is that the one you are working on is a super set of the two others, and, as we don't want to advocate deploying Broker CRD manually instead of using service & app, I think this should be the logical next step

Currently this is WIP. @viswanathannachiappan created a camel app that supports mqtt that we will try integrate into this tutorial instead of doing manual deployments. 

As far as merging with the other two tutorials, I would like to wait to consult with @gtully when he returns from PTO, as I understand there is more work to be done here.

lavocatt:
> As far as merging with the other two tutorials, I would like to wait to consult with @gtully when he returns from PTO, as I understand there is more work to be done here.

Also I didn't realize that you're using BrokerCluster for this tutorial. So that would possibly goes against the idea of merging them with the app/service ones.

docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,810 @@
...
+---
+
+This tutorial walks through deploying a publish/subscribe topology where:
+
+- A **Camel producer** (`camel-jms-app:mqtt`, `APP_ROLE=producer`) publishes messages to the `COMMANDS` MQTT5 topic on any broker over **mutual TLS**.
+- A **Camel consumer** (`camel-jms-app:mqtt`, `APP_ROLE=consumer`) subscribes to `COMMANDS`, appends `[PROCESSED BY APP]` to each message, and re-publishes it to the same topic.
+- **AMQP federation** links the two broker pods so that a message published to either broker is forwarded to consumers on both.
+- All MQTT5 traffic is secured with **mTLS**: the broker presents a server certificate and requires a client certificate signed by the same CA.
+
+> **_NOTE:_** This tutorial requires a running Kubernetes cluster with **cert-manager** installed. All client workloads run as Deployments inside the cluster using the `quay.io/rh-ee-vnachiap/camel-jms-app:mqtt` image.
```
gtully, line 20: this note in not required, the tutorial does the start and install of certmanager

docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,810 @@
...
+    - 'securityRoles."\$ACTIVEMQ_ARTEMIS_FEDERATION.#".control-plane.createNonDurableQueue=true'
+    - 'securityRoles."\$ACTIVEMQ_ARTEMIS_FEDERATION.#".control-plane.createAddress=true'
+    - 'securityRoles."\$ACTIVEMQ_ARTEMIS_FEDERATION.#".control-plane.consume=true'
+    - 'securityRoles."\$ACTIVEMQ_ARTEMIS_FEDERATION.#".control-plane.send=true'
+
+    # AMQP federation: each broker's complete connection config is self-contained
+    # under its own pod prefix.  All AMQPConnections.target.* properties MUST be
+    # under the same prefix as the URI — if credentials live in the shared file and
+    # the URI lives in the per-pod file, Artemis creates two separate connection
+    # objects and the per-pod one (with the URI but no credentials) wins, causing
```
gtully, line 459: this looks like a bug, it related to the a call to parseUri in the getTransportConnections accessor of the amqp connection configuration in the broker config logic, the current workaround of duplicate config seems sensible but is not ideal. There may be a better way or there may be need for a bug fix but for this tutorial it is no harm to demonstrate how it currently works.

docs/tutorials/pub_sub_scale.md
```diff
@@ -0,0 +1,810 @@
...
+    matchLabels:
+      app: consumer
+  template:
+    metadata:
+      labels:
+        app: consumer
+    spec:
+      initContainers:
+        - name: write-config
+          image: quay.io/rh-ee-vnachiap/camel-jms-app:mqtt
```
gtully, line 564: I think this camel app source should be contained in the tutorial, possibly in a sub dir _<turorial-name>_src and it gets built as part of the prerequisit or when we talk about the app. 
The image can be loaded into minikube with image load etc... that will avoid a stale container and a quay roundtrip.

gtully (commented):
a few comments, but this is a great exposition of what we have

gtully:
part of the value of this tutorial is that it is the raw version. a broker app/service supporting mqtt can be simpler by design and this can be used for contrast and eventually replaced.

docs/tutorials/pub_sub_scale.md
gtully: i makes me wonder how the test can work: https://github.com/arkmq-org/arkmq-org-broker-operator/blob/main/controllers/activemqartemis_pub_sub_scale_test.go#L174
is it just that the uri is always first in the test and not in the tutorial?

docs/tutorials/pub_sub_scale.md
gtully: actually, it may make sense to have a directory for this tutorial, and in there have the src and .md file.

docs/tutorials/pub_sub_scale.md
Robmurt4: Have removed this note now.

docs/tutorials/pub_sub_scale.md
Robmurt4: Shall I reuse the camel app that is being introduced in @viswanathannachiappan's PR https://github.com/arkmq-org/arkmq-org-broker-operator/pull/1452 for this?

docs/tutorials/pub_sub_scale.md
Robmurt4: I have updated the tutorial to be more in line with the test's approach, no longer duplicates config.

docs/tutorials/pub_sub_scale.md
Robmurt4: Right, so I have implemented it in this in a similar way to PR https://github.com/arkmq-org/arkmq-org-broker-operator/pull/1452

Issue #1457, opened 2026-08-12
The operator should manage the certs life cycle for the service & apps


lavocatt:
As a user, when deploying a service, I shouldn't have to generate certs beforehand, and I'm expecting that the operator handles the cert lifecyle while still being able to customize cert behavior such as renewal interval

PR #1458, opened 2026-08-12
[#1457] three-plane PKI architecture for namespace-scoped chain of trust

lavocatt:
Previously, deploying a BrokerService with TLS required users to
know internal naming conventions and to place cert-manager
ClusterIssuers, Certificates, and CA bundles at the correct spots
before any service or app could be deployed. This was a cluster-admin
task — clusterwide resources were involved, and the admin had to
coordinate certificate distribution manually between service owners
and app teams before they could work together.

Now, all PKI is fully operator-managed and namespace-scoped. There
are no clusterwide resources involved. RBAC can grant one team the
right to create BrokerServices and another team the right to create
BrokerApps. They do not need to exchange certificates manually — the
operator handles the full chain of trust automatically.

See docs/design/pki_architecture.md for more details


api/v1beta2/brokerservice_types.go
```diff
@@ -53,6 +53,12 @@ type BrokerServiceSpec struct {
 	//+optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="App Selector Expression"
 	AppSelectorExpression string `json:"appSelectorExpression,omitempty"`
+
+	// PKI configures the chain-of-trust certificate properties.
```
gtully, line 57: this term: chain-of-trust - it seems novel, but it does not seem informative to me. Any alternatives?
is this the control plane pki?

controllers/brokerapp_controller.go
```diff
@@ -195,16 +210,94 @@ func (reconciler BrokerAppInstanceReconciler) processBindingSecret() error {
...
+	for _, obj := range objects {
+		reconciler.TrackDesired(obj)
+	}
+
+	// For cross-namespace, add finalizer so we can clean up provisioning
+	// secrets from both the service and app namespace on deletion
+	if !sameNamespace && !controllerutil.ContainsFinalizer(reconciler.instance, cot.FinalizerName) {
+		controllerutil.AddFinalizer(reconciler.instance, cot.FinalizerName)
+		if err := reconciler.Update(ctx, reconciler.instance); err != nil {
+			return fmt.Errorf("failed to add chain-of-trust finalizer: %w", err)
```
gtully, line 333: i see chain-of-trust again and cot as the package... I just don't know if it should surface in errors etc, but I guess naming is hard. It is control plane pki cleanup?

controllers/brokerapp_controller.go
```diff
@@ -218,23 +311,48 @@ func (reconciler *BrokerAppReconciler) Reconcile(ctx context.Context, request ct
...
-					err = processor.SyncDesiredWithDeployed(processor.instance)
+		if err = processor.InitDeployed(instance, processor.getOwned()...); err == nil {
+			if err = processor.ensureAppPKI(ctx); err == nil {
+				if err = processor.resolveBrokerService(); err == nil {
+					if err = processor.processBindingSecret(); err == nil {
+						certErr := processor.ensureAppCert(ctx)
+						// Always sync so tracked PKI resources are materialized
+						// even when cross-namespace cert copies aren't ready yet.
+						if syncErr := processor.SyncDesiredWithDeployed(processor.instance); syncErr != nil {
+							err = syncErr
```
gtully: sync error should only override err if err is nil  to not hide it.

controllers/brokerapp_controller.go
```diff
@@ -218,23 +311,48 @@ func (reconciler *BrokerAppReconciler) Reconcile(ctx context.Context, request ct
 		return ctrl.Result{}, err
 	}
 
+	// Handle finalizer for cross-namespace cert cleanup
```
gtully, line 367: that may be the correct name for chain-of-trust errors... cross namespace cert cleanup

controllers/brokerapp_controller.go
```diff
@@ -1508,11 +1649,30 @@ func (r *BrokerAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&broker.BrokerApp{}).
 		Owns(&corev1.Secret{}).
+		Owns(&cmv1.Issuer{}).
+		Owns(&cmv1.Certificate{}).
 		Watches(&broker.BrokerService{}, r.enqueueAppsForService()).
 		Watches(&broker.BrokerApp{}, r.enqueueAppsForReferencedApp()).
+		Watches(&cmv1.Certificate{}, handler.EnqueueRequestsFromMapFunc(r.crossNsCertToApp)).
 		WithOptions(controller.Options{
-			// capacity allocation requires serial processing
```
gtully, line 1514: this is still true I think... as in the comment helps, no?

pkg/broker/jolokia/client.go
```diff
@@ -0,0 +1,154 @@
...
+		return fmt.Errorf("failed to parse root CA from %s", cot.RootCertSecretName(c.serviceOwner))
+	}
+
+	cfg.InsecureSkipVerify = false
+	cfg.Certificates = []tls.Certificate{cert}
+	cfg.RootCAs = pool
+	return nil
+}
+
+func (c *StatusClient) configureLegacy(cfg *tls.Config) {
```
gtully, line 144: what is the  legacy use case, or is it really legacy? I am wondering if we need it at all?

gtully (commented):
Great work, am still working through the implications and need to verify there is no alternative to the finalizer

gtully:
We also need to consider having an amqp broker connection in the control plane, which could be cross namespace or even cross cluster, this needs some though

lavocatt:
> We also need to consider having an amqp broker connection in the control plane, which could be cross namespace or even cross cluster, this needs some though

I think this is doable, if you look at the MQTT test that is upgraded to actually trusting certs, all it needed is a valid SAN for the ingress outside the cluster. Control plane to control plane would do the same.

controllers/brokerapp_controller.go
lavocatt: yea, the AI removed them for some reason :shrug: Will add them back! TBH I haven't checked the code yet, I'm still on the cycle spec & tests

api/v1beta2/brokerservice_types.go
lavocatt: chain of trust is what I have been using so far to describe a self signed trust anchor to the leaf certs chain. If you get inspired with a better name, I'll take it

controllers/brokerapp_controller.go
lavocatt: Possibly it could/should :slightly_smiling_face:

pkg/broker/jolokia/client.go
lavocatt: I've designed the implementation so that it's additive only to the existing capabilities of the Broker CRD, without damaging it's capacity of being deployed on its own. This could have some value, or not. I was thinking, what if a customer wants to customize broker properties manually, they'd have to fire a Broker, and configure it manually, the legacy way being the only way to give certs.

gtully: or the broker cr would need to be accompanied by some cert manager crds, the use case of a standalone broker cr is real, maybe we need to trust existing secrets and drop the PKINameClash logic. Not sure of the implications of that, but it would be possibly the most consistent approach.

docs/design/pki_architecture.md
```diff
@@ -0,0 +1,666 @@
...
+
+of trust automatically.
+
+**Tradeoff.** cert-manager is now a hard startup dependency — the operator
+
+crashes if it is absent, and a runtime watcher degrades BrokerService status
+
+if cert-manager disappears post-boot. All conditional `hasCertManager` logic
+
+has been removed; the operator always assumes cert-manager is present.
```
brusdev, line 32: Why must cert-manager be a requirement? Can users provide their certificates without cert-manager?

docs/design/pki_architecture.md
```diff
@@ -0,0 +1,666 @@
...
+
+## Overview
+The operator provisions three independent PKI domains, each with its own
+
+self-signed root CA, namespace-scoped Issuers, and leaf certificates. Every TLS
```
brusdev, line 37: The self-signed root CAs s are a valid and secure solution for internal mTLS between controlled components. However, they don't work well with: regulated industries with compliance requirements, external client authentication and multi-cluster or multi-organization trust federation.

docs/design/pki_architecture.md
```diff
@@ -0,0 +1,666 @@
...
+
+all. Three planes give each domain independent lifecycle.
+
+**API surface leakage.** Exposing customization for app certs also exposes it
+
+for operator certs, creating misconfiguration risk on internal plumbing. The
+
+three-plane model gives each domain exactly the API surface it needs: none for
+
+operator, optional for metrics, full for app.
```
brusdev, line 70: Only the **blast radius** problem seems directly related to a single CA. The other 2 seems more related to a single plane.

docs/design/pki_architecture.md
```diff
@@ -0,0 +1,666 @@
...
+signed by app-alpha's CA, you are app-alpha — no one else can forge that
+identity.
+
+A public CA (Let's Encrypt, DigiCert) makes the trust domain the entire
+internet. Anyone who can obtain a certificate from that CA — which is trivially
+anyone — can present a valid client certificate. The CA doesn't just sign a
+name; it vouches for identity within a trust domain. With a private CA, "this
+cert is signed by my CA" means "I issued this identity." With a public CA,
+"this cert is signed by Let's Encrypt" means "someone proved domain ownership"
+— which says nothing about authorization to connect to a broker.
```
brusdev, line 197: This seems correct: domain validation is a critical step in the CA issuance process, where the CA's sole responsibility is to verify the certificate holder's identity. It is then up to the applications using the certificate to enforce proper access controls and authorize specific operations for that identity.

I don't see why this is an argument to opt for private PKI.

docs/design/pki_architecture.md
```diff
@@ -0,0 +1,666 @@
...
+  capabilities:
+    - producerOf:
+        - address: my-topic
+```
+
+With this configuration, the app-plane server cert will contain both the
+cluster-internal DNS names (generated by the operator) and the external
+hostnames. An Ingress or OpenShift Route configured for SSL passthrough will
+route external TLS connections directly to the broker, and the client's TLS
+verification will succeed.
```
brusdev, line 238: The external clients would need the self-signed root CA or the broker cert to validate the TLS connection.

brusdev (commented):
Great work! I reviewed just the design doc for now since the PR is a draft. Two key items caught my attention: the cert-manager requirement and the self-signed root CAs.

Issue #1473, opened 2026-08-19
Test isolation: use UUID-prefixed namespaces per spec


lavocatt:
## Problem

Tests currently run in shared, well-known namespaces (`test`, `other`, `restricted`). This creates fragile inter-test dependencies:

- A killed test run leaves resources behind that poison subsequent runs (e.g. `secrets "broker-user-cred-credentials-secret" already exists`)
- Tests implicitly depend on namespace state they don't control
- No path to parallel test execution (`--procs=N`)

## Proposal

Introduce UUID-prefixed namespaces so each spec runs in isolation.

### Behavior

When a test requests namespace `test`, it gets `test-<short-uuid>` (e.g. `test-a3f1b9`). Each spec creates its own namespace and tears it down via `DeferCleanup`.

### Disabling random prefixes

An environment variable (e.g. `TEST_DISABLE_NS_PREFIX=true`) disables the UUID prefix, falling back to the original fixed namespace names. This is useful for:
- Debugging a specific test interactively
- Reproducing issues that depend on a known namespace name
- Backward compatibility during migration

### Acceptance criteria

- [ ] Every test spec uses a UUID-prefixed namespace by default
- [ ] Setting `TEST_DISABLE_NS_PREFIX=true` disables prefixing — namespaces use their original names
- [ ] Namespace creation and deletion is handled automatically (DeferCleanup)
- [ ] Existing tests pass with both modes (prefixed and non-prefixed)
- [ ] No test relies on hardcoded namespace names from other specs

### Future work

Once namespace isolation is in place, evaluate parallel execution feasibility:
- Manager singleton / metrics port binding per process
- Cluster-scoped resource conflicts (ClusterRoles, ClusterIssuers)
- Node resource pressure under concurrent broker deployments

PR #1476, opened 2026-08-20
[#1473] isolate test namespace to avoid cross test pollution

lavocatt:
Making the test suite more robust with independent namespaces.

### DIFF
--- a/controllers/spec_namespace_test.go
+++ b/controllers/spec_namespace_test.go
@@ -1,0 +1,251 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	crand "crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/arkmq-org/arkmq-org-broker-operator/v2/pkg/utils/common"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	specNamespaceDefaultBase    = "test"
+	specNamespaceOtherBase      = "other"
+	specNamespaceRestrictedBase = "restricted"
+	specNamespace1Base          = "ns-one"
+	specNamespace2Base          = "ns-two"
+	specNamespace3Base          = "ns-three"
+
+	specNamespaceHexBytes       = 3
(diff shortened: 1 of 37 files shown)


gtully (commented):
I guess the intention of using a few namespace was to try and enforce test isolation, but it looks like we have failed. But I think you have identified some of the problems that are test specific which is good. I guess a benefit of isolation could be that we could run some specs in parallel?

Gingko has a rand seed that governs test run order, and can be configured to make it static. I wonder if something similar can work for the prefix generation, possibly a similar pattern

controllers/activemqartemis_work_queue_test.go
```diff
@@ -319,69 +324,64 @@ var _ = Describe("work queue", func() {
...
 				checkJOBSRoutedNonZero := func(metrics string) bool {
 					return strings.Contains(metrics, "artemis_routed_message_count{address=\"JOBS\",broker=\"amq-broker\",}") && !strings.Contains(metrics, "} 0.0")
 				}
 				Eventually(func(g Gomega) {
+					g.Expect(metricsCheck(g, "0", checkJOBSRoutedNonZero) || metricsCheck(g, "1", checkJOBSRoutedNonZero)).Should(BeTrue())
+				}, existingClusterTimeout, existingClusterInterval*5).Should(Succeed())
 
-					g.Expect(metricsCheck(g, "0", checkJOBSRoutedNonZero)).Should(Equal(true))
-					g.Expect(metricsCheck(g, "1", checkJOBSRoutedNonZero)).Should(Equal(true))
+				By("deploying " + fmt.Sprintf("%d", numConsumers) + " consumers to consume in batches of " + numMessagesToConsume)
```
brusdev: Placing consumers after producers alters the test's semantics.

controllers/spec_namespace_test.go
```diff
@@ -0,0 +1,242 @@
...
+func specNamespacePrefixDisabled() bool {
+	if os.Getenv("DEPLOY_OPERATOR") == "true" {
+		return true
+	}
+	if v, ok := os.LookupEnv("TEST_DISABLE_NS_PREFIX"); ok {
+		disabled, err := strconv.ParseBool(v)
+		return err == nil && disabled
+	}
+	return false
+}
```
brusdev, line 58: This function checks whether a random suffix is required, so using prefix in the name is a bit confusing. Since prefix/suffix logic is just an implementation detail for namespace isolation, could we rename the function to reflect its main purpose instead?

controllers/spec_namespace_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+func shortHexID() string {
+	b := make([]byte, specNamespaceHexBytes)
+	if _, err := crand.Read(b); err != nil {
```
brusdev, line 62: We can use GinkgoRandomSeed() to seed standard library RNG (math/rand/v2) for reproducible failures.

controllers/spec_namespace_test.go
```diff
@@ -0,0 +1,242 @@
...
+		panic(fmt.Sprintf("failed to read random bytes for spec namespace: %v", err))
+	}
+	return hex.EncodeToString(b)
+}
+
+func uniqueSpecNamespace(base string) string {
+	if specNamespacePrefixDisabled() {
+		return base
+	}
+	return base + "-" + shortHexID()
```
brusdev, line 72: The base string should not include the separator used by specNamespaceBase to split the base and the suffix.

brusdev (commented):
Great work on the test suite! Proper test isolation will make troubleshooting failures much easier and prevent test pollution.

controllers/activemqartemis_work_queue_test.go
lavocatt: I had problems with this test, it was so flaky, I tried to make it pass reliably..

lavocatt:
Closing this work, it adds too much running time to the tests.

Issue #1474, opened 2026-08-19
Test suite: fix stdin leak and cleanup bugs in test utilities


lavocatt:
## Context

Found during review of #1438. These are independent bug fixes that should land before the Ginkgo migration.

### 1. Stdin leak in pod exec helpers (`common_util_test.go`)

`RunCommandInPodWithNamespace` and `ExecOnPod` pass `Stdin: true` / `Stdin: os.Stdin` to the remote exec stream. This leaks the host process stdin into the remote command, which can cause hangs or unexpected behavior in CI.

**Fix:** `Stdin: false` / `Stdin: nil`

### 2. `testWatchNamespace` cleanup skipped on failure (`controllermanager_test.go`)

Namespace cleanup runs *after* the test function call, not in a `defer`. If `testFunc(g)` panics or fails hard, namespaces and the replacement controller manager are never cleaned up, poisoning subsequent specs.

**Fix:** move cleanup to `defer`

### 3. Stale cert-manager secrets not cleaned up (`common_util_test.go`)

`InstallCert` and `InstallCaBundle` delete the Certificate/Bundle CR before re-creating, but cert-manager does NOT garbage-collect the associated Secret. A stale secret from a previous run can cause tests to pass with old certificate material.

**Fix:** explicitly delete the target secret and wait for its removal before re-creating the CR.

### 4. `readCloser.Close()` return value ignored (`common_util_test.go`)

Minor: `defer readCloser.Close()` should be `defer func() { _ = readCloser.Close() }()` to satisfy linters.

## Relation

Prerequisite cleanup for #1375. See also #1473 for namespace isolation.

PR #1477, opened 2026-08-21
[#1474] Fix stdin leak and cleanup bugs in test utilities

rii13024:
Extracted from PR #1438 (Ginkgo migration). No production code
is touched. No test logic changes. No Ginkgo/Gomega conversion is included.

---

#### Bug 1: Stdin leak in `RunCommandInPodWithNamespace` and `ExecOnPod`

Both helpers set `PodExecOptions.Stdin: true` and passed `os.Stdin` into
`remotecommand.StreamOptions`. This attached the host process's own stdin FD
to every pod-exec call. In CI there is no terminal attached, so the remote
command blocks waiting for input that never arrives and the FD is never
released. Fix: `Stdin: false` / `Stdin: nil`.

#### Bug 2: `testWatchNamespace` cleanup skipped on failure

Cleanup (`shutdownControllerManager`, four `deleteNamespace` calls,
`createControllerManagerForSuite`) ran inline after `testFunc(g)`. Ginkgo's
`Expect` failure handler panics, so any failing spec skips the entire cleanup
block, leaving orphaned namespaces and a live replacement controller manager
for the rest of the suite. Fix: wrap the cleanup block in
`defer func() { ... }()` placed **before** the `switch`/`testFunc` call so
it runs unconditionally.

#### Bug 3: Stale secrets in `InstallCert` / `InstallCaBundle`

cert-manager and trust-manager do **not** delete the target Secret when the
`Certificate` or `Bundle` CR is deleted. Both helpers deleted the CR before
re-creating it but left the Secret untouched, so a subsequent call returned
stale certificate material. Fix: explicitly delete the target Secret and
`Eventually` poll for `errors.IsNotFound` before re-creating the CR.

> **Note on `InstallCert`:** `secretKey` was originally declared after the
> `Eventually` wait for the cert. The stale-secret block needs it earlier, so
> the declaration is hoisted and the later `secretKey :=` re-declaration is
> removed. This is a required structural change, not incidental cleanup.

Post-review update: The original stale-secret guard used if k8sClient.Get(...) == nil, which silently skips the delete on any non-nil error (transient API error, RBAC denial, etc.), masking real failures. Both InstallCert and InstallCaBundle now capture the error and assert it with client.IgnoreNotFound - the same idiom already used one line below on the Delete call in each block. Copilot flagged InstallCaBundle; InstallCert has the structurally identical pattern and was fixed in the same change.

Copilot also raised a race concern: deleting the Secret while the CR still exists could let the controller recreate it mid-wait. This is not a real race - both functions delete the CR before the stale-secret block runs, so there is no live controller to recreate the secret during the IsNotFound wait. No ordering change is needed.

#### Bug 4: Ignored `Close()` error in `LogsOfPod` / `GetOperatorLog`

`defer readCloser.Close()` and `defer podLogs.Close()` discarded the `error`
return without an explicit `_ =`, triggering the `errcheck` linter. Fix:
`defer func() { _ = x.Close() }()`.

---

### What was deliberately NOT included

Several related changes present in `feature/ginkgo-conversion` were reviewed
and explicitly dropped to keep this PR minimal:

| Change | Decision |
|---|---|
| `duration` → `existingClusterTimeout` in `RunCommandInPodWithNamespace` | **Dropped** - incidental timeout tweak; `duration = 10s` is correct on `main` |
| `timeout` → `existingClusterTimeout` in `ExecOnPod` | **Dropped** - same reason; `timeout = 30s` is correct on `main` |
| `k8sClient.Delete(ctx, &cmCert)` → `_ = k8sClient.Delete(...)` | **Dropped** - linter cleanup, not part of the stale-secret fix |
| `k8sClient.Delete(ctx, &bundle)` → `_ = k8sClient.Delete(...)` | **Dropped** - same reason |
| `createControllerManager(true, ...)` → `createControllerManager(...)` | **Dropped** - separate signature refactor unrelated to #1474 |
| `managerCancel == nil` guard in `testWatchNamespace` | **Dropped** - unrelated guard from the Ginkgo migration |
| `createNamespace` `USE_EXISTING_CLUSTER` conditional refactor | **Dropped** - unrelated cleanup |
| `InstallCaBundle` Bundle-sync wait improvement | **Dropped** - beyond the stale-secret fix scope |
| `g.Expect(k8sClient.Update(...)).Should(Succeed())` in `testWatchNamespace` | **Dropped** - unrelated assert fix |

---

### Files changed

| File | Bugs fixed |
|---|---|
| `controllers/common_util_test.go` | 1, 3, 4 |
| `controllers/controllermanager_test.go` | 2 |

---

### Test results

✅ Ran the full E2E test suite locally — all passing.
**263/277 specs passed (0 failed, 14 skipped)**, ~3695s runtime.

### DIFF
--- a/controllers/common_util_test.go
+++ b/controllers/common_util_test.go
@@ -490,7 +490,7 @@ func RunCommandInPodWithNamespace(podName string, podNamespace string, container
 		VersionedParams(&corev1.PodExecOptions{
 			Container: containerName,
 			Command:   command,
-			Stdin:     true,
+			Stdin:     false,
 			Stdout:    true,
 			Stderr:    true,
 		}, runtime.NewParameterCodec(scheme.Scheme))
@@ -508,7 +508,7 @@ func RunCommandInPodWithNamespace(podName string, podNamespace string, container
 	defer cancel()
 
 	err = exec.StreamWithContext(execCtx, remotecommand.StreamOptions{
-		Stdin:  os.Stdin,
+		Stdin:  nil,
 		Stdout: &consumerCapturedOut,
 		Stderr: &consumerCapturedErr,
 		Tty:    false,
@@ -576,7 +576,7 @@ func LogsOfPod(podWithOrdinal string, brokerName string, namespace string, g Gom
 		}, runtime.NewParameterCodec(scheme.Scheme)).Stream(context.TODO())
 	g.Expect(err).To(BeNil())
 
-	defer readCloser.Close()
+	defer func() { _ = readCloser.Close() }()
 
 	result, err := io.ReadAll(readCloser)
 	g.Expect(err).To(BeNil())
@@ -605,7 +605,7 @@ func ExecOnPod(podWithOrdinal string, brokerName string, namespace string, comma
 		VersionedParams(&corev1.PodExecOptions{
 			Container: brokerName + "-container",
 			Command:   command,
-			Stdin:     true,
+			Stdin:     false,
 			Stdout:    true,
 			Stderr:    true,
 		}, runtime.NewParameterCodec(scheme.Scheme))
@@ -621,7 +621,7 @@ func ExecOnPod(podWithOrdinal string, brokerName string, namespace string, comma
 	defer cancel()
 
 	err = exec.StreamWithContext(execCtx, remotecommand.StreamOptions{
-		Stdin:  os.Stdin,
+		Stdin:  nil,
 		Stdout: &outPutbuffer,
 		Stderr: &errBuffer,
 		Tty:    false,
@@ -729,7 +729,7 @@ func GetOperatorLog(ns string) (*string, error) {
 
 						var podLogs io.ReadCloser
 						if podLogs, err = req.Stream(context.Background()); err == nil {
-							defer podLogs.Close()
+							defer func() { _ = podLogs.Close() }()
 							buf := new(bytes.Buffer)
 							if _, err = io.Copy(buf, podLogs); err == nil {
 								str := buf.String()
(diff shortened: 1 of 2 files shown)


lavocatt (commented):
Good scoping discipline — four independent fixes, cleanly separated from the Ginkgo migration, with an explicit table of what was dropped and why. Each fix is mechanically correct. Two things worth discussing before merge.

## 1. `testWatchNamespace` defer can itself panic

The cleanup block now runs in a `defer`, which is strictly better than the inline version that was skipped entirely on failure. But `deleteNamespace` calls `g.Expect` / `g.Eventually` internally — if `testFunc(g)` panicked via Gomega's failure handler, further assertions on the same `g` instance will also panic. This means `createControllerManagerForSuite()` at the end of the defer may never execute, leaving the suite's controller manager in a broken state for all subsequent specs.

A `recover()` wrapper would make the cleanup truly unconditional:

```go
defer func() {
    defer func() {
        if r := recover(); r != nil {
            fmt.Fprintf(GinkgoWriter, "cleanup panic (recovered): %v\n", r)
        }
        createControllerManagerForSuite()
    }()
    shutdownControllerManager()
    deleteNamespace(namespace1, true, g)
    deleteNamespace(namespace2, true, g)
    deleteNamespace(namespace3, true, g)
    deleteNamespace(restrictedNamespace, true, g)
}()
```

This guarantees `createControllerManagerForSuite()` runs even if namespace deletion panics. Not a hard blocker — the PR is already better than `main` — but without this, the fix only covers the "testFunc fails" case, not the "cleanup itself fails" case.

## 2. `InstallCert` line 1046 — pre-existing dead code, but now load-bearing context

The stale-secret block's safety depends on the Certificate CR being already gone when it runs. That's true — `CleanResource` at line 1020-1023 handles it. But line 1046 (`k8sClient.Delete(ctx, &cmCert)`) then deletes a struct that was just constructed and never created — it's a silent no-op. Same pattern in `InstallCaBundle` at line 1098.

This isn't something the PR introduced, and I wouldn't ask you to fix it here. But it's worth a follow-up cleanup — the dead `Delete` calls make the control flow harder to reason about, and someone reading the stale-secret block might incorrectly think the CR is still alive at that point.

## The rest

- **Stdin fix**: Correct. `Stdin: false` / `nil` is the right call — these commands produce output, they don't consume input. The `os` import stays valid via `os.LookupEnv`.
- **Stale secret cleanup**: The `Expect(client.IgnoreNotFound(getErr)).To(Succeed())` pattern is correct — fails on real errors, passes through NotFound. The race concern Copilot raised doesn't apply since both functions delete the CR before the stale-secret block.
- **Close() linter fix**: Trivially correct.

lavocatt (approved):
LGTM, thanks @rii13024!

lavocatt:
Will wait for the results of your E2E test suite on your local machine and the CI result before merging, but good job on this one.

Issue #1468, opened 2026-08-18
Improve test coverage for pkg/brokervolumes


lavocatt:
## Context

PR #1453 introduced `pkg/brokervolumes` with shared volume logic extracted from both reconcilers. The package ships with unit tests, but coverage is shallow — tests verify structure (non-nil, length) without asserting the behavioral contracts the reconcilers depend on.

### MakeVolumes
- No combined test: `persistent=true` + extras + EPVCs together. The append ordering (PVC → extras → EPVCs) is untested.
- Persistent volume test asserts `PersistentVolumeClaim != nil` but never checks that `ClaimName` matches the `name` argument.

### MakeExtraVolumeMounts
- EPVC-to-mount matching (lines 59–75) is completely untested — only the extras path is exercised.
- No test combines extras and EPVCs simultaneously, which is the actual call pattern from both reconcilers.
- `ReadOnly`, `SubPath` preservation from user-provided mounts is unverified.

### MakeVolumeMounts
- Single test, only checks names exist. Does not verify `DataMountPath` (`/app`) on the persistent mount or ordering (persistent first, extras after).
- EPVC path untested (inherited gap from MakeExtraVolumeMounts).

### RemovePVCOwnerRef
- Missing: PVC-not-found case (should be silent).
- Missing: PVC with zero owner refs (early return on line 95).
- Missing: multiple matching UIDs in the same owner ref list.

## Suggested approach

Add a composition test per function that mirrors the real call site — all parameters populated — and asserts claim names, mount paths, and ordering. Add targeted cases for the EPVC mount-matching branch and the RemovePVCOwnerRef edge cases listed above.

Relates to #1439.

PR #1478, opened 2026-08-21
[#1468] Improve test coverage for pkg/brokervolumes

Robmurt4:
Solves issue [#1468]

**MakeVolumes Test:** 
- now asserts ClaimName equals name argument passed to MakeVolumes.
- asserts ClaimName for both the broker PVC and the EPVC, preserves the extra volume's VolumeSource, and asserts append order (PVC → extras → EPVCs) by index.

**MakeExtraVolumeMounts:**
- asserts use of default EPVC mount path when no user mount is provided
- asserts use of user-provided mount for EPVC when names match
- verifies ReadOnly, SubPath preservation from user-provided mounts

**MakeVolumeMounts:**
- Replaces existing test with one that mirrors full call-site with persistent mount, extra volume with user override, and EPVC. Asserts that index 0 is the persistent mount at DataMountPath (/app) with ReadOnly: false, index 1 is the extra with the user's MountPath/ReadOnly/SubPath intact, and index 2 is the EPVC at its default path.

**RemovePVCOwnerRef:**
- PVC-not-found case
- PVC with zero owner refs case
- multiple matching UIDs in the same owner ref list case

### DIFF
--- a/pkg/brokervolumes/volumes_test.go
+++ b/pkg/brokervolumes/volumes_test.go
@@ -19,10 +19,37 @@ var _ = Describe("MakeVolumes", func() {
 		Expect(MakeVolumes("my-broker", false, nil, nil)).To(BeEmpty())
 	})
 
-	It("returns a PVC volume when persistent", func() {
+	It("returns a PVC volume when persistent with ClaimName matching the name argument", func() {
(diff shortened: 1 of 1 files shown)


makella19 (approved):
LGTM! thanks!

PR #1488, opened 2026-08-28
Add EXTRA_BROKER_PROPERTIES env var to extend broker.properties paths

brusdev:
Introduce EXTRA_BROKER_PROPERTIES, a new operator-internal env var that users can set in spec.env with a comma-separated list of extra .properties file or directory paths. The operator appends ,$(EXTRA_BROKER_PROPERTIES) to its own -Dbroker.properties= value; Kubernetes resolves the $(VAR_NAME) reference at pod admission time so the operator's paths always come first.

### DIFF
--- a/controllers/broker_controller.go
+++ b/controllers/broker_controller.go
@@ -1184,6 +1184,11 @@ func (reconciler *BrokerReconcilerImpl) brokerPropertiesConfigSystemPropValue(mo
 		}
 	}
 
+	// Append user-defined extra paths via Kubernetes env var substitution.
+	// The trailing comma is safe: Java String.split(",") strips trailing empty tokens,
+	// so Artemis parseProperties receives no empty path when the var is unset.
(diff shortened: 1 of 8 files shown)


controllers/broker_controller.go
```diff
@@ -1184,6 +1184,11 @@ func (reconciler *BrokerReconcilerImpl) brokerPropertiesConfigSystemPropValue(mo
 		}
 	}
 
+	// Append user-defined extra paths via Kubernetes env var substitution.
+	// The trailing comma is safe: Java String.split(",") strips trailing empty tokens,
+	// so Artemis parseProperties receives no empty path when the var is unset.
+	result = result + ",$(EXTRA_BROKER_PROPERTIES)"
+
```
lavocatt, line 1191: I notice that the addition is made to the two controllers, should it be extracted in a common package for the env variable handling?

Issue #1485, opened 2026-08-26
Switch operator-generated broker properties to JSON format


lavocatt:
## TL;DR

Replace the inline `fmt.Fprintf` property generation in the broker and service controllers with Go structs serialized to JSON via `json.Marshal`. Artemis already supports JSON as a broker properties format. This separates intent from serialization, adds type safety, closes a property injection vulnerability, improves test surface, and gives contributors a clear reference path from the broker's test suite to the operator's Go structs.

## Context

The operator generates several `.properties` files that Artemis parses via `ConfigurationImpl.parseFileProperties()`. These are built inline with `fmt.Fprintf` as flat key=value strings with manually dot-separated paths and quoted keys.

Artemis supports JSON and YAML as broker properties file formats alongside `.properties`. The file extension determines the parser. All three flatten into the same internal representation, so the operator can emit JSON with no change on the broker side.

## Why change

The operator builds broker configuration by manually serializing Go values into flat `.properties` key-paths using `fmt.Fprintf`. This was the only option when the code was written. JSON broker properties were added to Artemis later, opening an opportunity to simplify. There are four concrete problems with continuing the manual approach.

**1. Contributors carry unnecessary cognitive load.** Every `fmt.Fprintf` line encodes both *what* the operator wants to configure and *how* to serialize it into the properties format. Adding a broker property today means: figure out the dot-path from Artemis source, manually quote keys that contain dots, handle comma-separated arrays by hand, and hope the escaping is right. With JSON structs, the contributor looks at the broker's config test (`artemis-server/src/test/java/.../JsonConfigurationFullTest.java`) and its test resource (`artemis-server/src/test/resources/broker-full-config.json`), then builds the corresponding Go struct in the operator. This creates a feedback loop: there's an incentive to update that test when new broker properties are added on the Artemis side, and an incentive to use it as reference when building the Go struct on the operator side.

[...]

PR #1492, opened 2026-09-02
[#1439] extraction of inline config generation

makella19:
Extracts inline config file generation from broker_controller.go and brokerservice_controller.go, these are the non-Artemis config files
The 4 Artemis-parsed broker properties files are left inline — these will be handled by #1485


pkg/brokerproperties/configs.go
```diff
@@ -0,0 +1,169 @@
...
+	fmt.Fprintf(buf, "    type: GAUGE\n")
+	fmt.Fprintf(buf, "  - pattern: 'org.apache.activemq.artemis<broker=\"%s\"><>TotalMessagesAcknowledged'\n", brokerName)
+	fmt.Fprintf(buf, "    help: Number of messages consumed since start\n")
+	fmt.Fprintf(buf, "    name: artemis_total_consumed_message_count\n")
+	fmt.Fprintf(buf, "    type: COUNTER\n")
+	fmt.Fprintf(buf, "  - pattern: 'org.apache.activemq.artemis<broker=\"%s\"><>TotalMessagesAdded'\n", brokerName)
+	fmt.Fprintf(buf, "    help: Number of messages produced since start\n")
+	fmt.Fprintf(buf, "    name: artemis_total_produced_message_count\n")
+	fmt.Fprintf(buf, "    type: COUNTER\n")
+	return buf.Bytes()
```
lavocatt: These looks like templates to me, and I'm wondering if, we shouldn't treat them as such.

I.E have a `templates/prometheus.template` file. Where we'd search and replace the `$BROKER_NAME` with the `brokerName`.

There's a neat golang property that we can use for loading that template as a string:

```
//go:embed templates/prometheus.template
var prometheusTemplate string
```

lavocatt (commented):
I like how this is simplifying the code of the controller. I think we can get that a notch better with a template approach. Most of the customization is done on a few fields, such as `mountPathRoot`, `realm`, `CN` etc.


The idea would be to get a raw file under a `templates/xyz.template` load it with a `go:embed` annotation, and let customize the few fields that needs to be customized.

WDYT?

pkg/brokerproperties/configs.go
lavocatt: Maybe https://pkg.go.dev/text/template is interesting for us

controllers/broker_controller.go
```diff
@@ -812,16 +812,18 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
 	additionalSystemProps := []string{}
 	{
 		mountPathRoot := common.SecretPathBase + getPropertiesResourceNsNameForBroker(customResource).Name
-		brokerPropertiesMapData["_security.config"] = brokerproperties.SecurityConfigData(mountPathRoot)
+		cfg := brokerproperties.TemplateConfig{
```
lavocatt: I'm thinking that the package should be `templates` instead of `brokerproperties`

Therefore you'd have `templates.Config{ ... }`

it's a bit counter intuitive because we paste all of that in `brokerPropertiesMapData`, but these templates results aren't broker properties per say.

controllers/broker_controller.go
```diff
@@ -812,16 +812,18 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
...
 
 		additionalSystemProps = append(additionalSystemProps, fmt.Sprintf("-Djava.security.properties=%s/_security.config", mountPathRoot))
 
-		brokerPropertiesMapData[brokerproperties.JaasConfigKey] = brokerproperties.LoginConfigData(
-			common.HttpAuthenticatorRealm,
-			mountPathRoot,
-			common.GetCertUsersKey(common.HttpAuthenticatorRealm),
-			common.GetCertRolesKey(common.HttpAuthenticatorRealm),
-		)
+		brokerPropertiesMapData[brokerproperties.JaasConfigKey] = brokerproperties.RenderConfig("login_config", cfg)
```
lavocatt: `brokerproperties.RenderConfig("login_config", cfg)`

Could then be
`
templates.render(loginConfig, cfg)`

I'd also get the names of each config in a const that gets imported from the templates package 🙂

pkg/brokerproperties/configs.go
```diff
@@ -2,111 +2,93 @@ package brokerproperties
...
+	CACertPath       string
+	ServerCertPath   string
+	ServerKeyPath    string
+	Alias            string
+	CertKeyPath      string
+	CertCrtPath      string
+	PemCfgPath       string
+	CATrustStorePath string
+	BrokerName       string
+	Queues           []QueueEntry
```
lavocatt: Possibly that would also be good within the `templates` package.

lavocatt (commented):
Very good progress, I like the way the templates are extracted in a sub directory and the fact that the variables they require is clearly visible. It's a win. A few comments, let me know what you think about them.

lavocatt:
I'm also thinking that every template could have its own dedicated struct. So you'd set the config for template XYZ, and not send the same struct for template ABC. WDYT?

controllers/brokerservice_controller.go
```diff
@@ -946,15 +947,11 @@ func (reconciler *BrokerServiceInstanceReconciler) getTrustStorePath(_ *broker.B
...
-	buf := brokerproperties.NewPropsWithHeader()
-
-	certSecretName := certSecretName(service)
-
-	fmt.Fprintf(buf, "source.key=/amq/extra/secrets/%s/tls.key\n", certSecretName)
-	fmt.Fprintf(buf, "source.cert=/amq/extra/secrets/%s/tls.crt\n", certSecretName)
-
-	return buf.Bytes()
+	certSecret := certSecretName(service)
+	return templates.Render(templates.PemCfg, templates.PemCfgConfig{
```
lavocatt, line 953: One thing, what happens if the template can't be rendered (like an error of some sort?)? templates.Render should possibly return err too.

gtully (approved):
this does tidy up the production of properties. nice work,

controllers/brokerservice_controller.go
```diff
@@ -945,16 +948,12 @@ func (reconciler *BrokerServiceInstanceReconciler) getTrustStorePath(_ *broker.B
...
-
-	fmt.Fprintf(buf, "source.key=/amq/extra/secrets/%s/tls.key\n", certSecretName)
-	fmt.Fprintf(buf, "source.cert=/amq/extra/secrets/%s/tls.crt\n", certSecretName)
-
-	return buf.Bytes()
+func (reconciler *BrokerServiceInstanceReconciler) makePemCfgProps(service *broker.BrokerService) ([]byte, error) {
+	certSecret := certSecretName(service)
+	return templates.Render(templates.PemCfg, templates.PemCfgConfig{
+		CertKeyPath: fmt.Sprintf("/amq/extra/secrets/%s/tls.key", certSecret),
+		CertCrtPath: fmt.Sprintf("/amq/extra/secrets/%s/tls.crt", certSecret),
```
lavocatt: How about we use path.Join here? 

```
CertKeyPath: path.Join(secretsBasePath, certSecret, tls.key)
CertCrtPath: path.Join(secretsBasePath, certSecret, tls.crt)
```

controllers/broker_controller.go
```diff
@@ -1011,68 +1006,30 @@ func (reconciler *BrokerReconcilerImpl) PodTemplateSpecForCR(customResource *v1b
...
-		fmt.Fprintf(prometheusConfig, "    name: artemis_total_consumed_message_count\n")
-		fmt.Fprintf(prometheusConfig, "    type: COUNTER\n")
-		fmt.Fprintf(prometheusConfig, "  - pattern: 'org.apache.activemq.artemis<broker=\"%s\"><>TotalMessagesAdded'\n", environments.ResolveBrokerNameFromEnvs(customResource.Spec.Env, customResource.Name))
-		fmt.Fprintf(prometheusConfig, "    help: Number of messages produced since start\n")
-		fmt.Fprintf(prometheusConfig, "    name: artemis_total_produced_message_count\n")
-		fmt.Fprintf(prometheusConfig, "    type: COUNTER\n")
-
-		brokerPropertiesMapData[PrometheusConfigFileName] = prometheusConfig.Bytes()
+		certSecretPath := common.SecretPathBase + operandCertSecretName
+		caCertPath := fmt.Sprintf("%s%s/%s", common.SecretPathBase, caSecret, caSecretKey)
```
lavocatt: I would use path.Join here as well

lavocatt (approved):
Great changes. It really makes the code easier to follow.
While you're at it, I'd change the string concatenation for path to using path.Join() which makes it clearer that we're working with subdirectories string builds.

PR #1493, opened 2026-09-02
[#1475] Enable unit-only test mode without envtest/cluster

### LINKED ISSUE #1473: Test isolation: use UUID-prefixed namespaces per spec

### LINKED ISSUE #1474: Test suite: fix stdin leak and cleanup bugs in test utilities

### LINKED ISSUE #1475: Test suite: enable unit-only mode without envtest/cluster
## Context

Found during review of #1438. The test suite currently requires either `KUBEBUILDER_ASSETS` or `USE_EXISTING_CLUSTER=true` to run any tests, including pure unit tests that don't touch the API server.

## Proposal

Add infrastructure so that unit tests can run without spinning up envtest:

1. **Label constant** — define a `unit` label constant and tag all pure-unit `It` blocks with `Label(unitLabel)`
2. **BeforeSuite guard** — detect when only unit-labeled specs are selected (via `GinkgoLabelFilter()`) and skip control plane setup
3. **`k8sClient == nil` Skip guards** — integration/E2E specs that require a cluster should `Skip()` gracefully when no cluster is available, rather than panicking on a nil client
4. **Always disable metrics in test** — `createControllerManager` should bind metrics to `"0"` to avoid port conflicts when multiple test processes run

### Acceptance criteria

- [ ] `go test ./controllers/... --ginkgo.label-filter=unit` passes without `KUBEBUILDER_ASSETS`
- [ ] Integration tests still pass with envtest
- [ ] E2E tests still pass against a real cluster
- [ ] No test silently skips when it should run

## Relation

Prerequisite for #1375 (Ginkgo migration). See also #1473 (namespace isolation) and #1474 (test utility bug fixes).

rii13024:
## Summary

Closes #1475

Adds the minimal infrastructure so that:

```
go test ./controllers/... --ginkgo.label-filter=unit
```

passes with **no** `KUBEBUILDER_ASSETS` and **no** running cluster, while all integration and E2E tests continue to work exactly as before.

## Sub-Tasks included (all four)

**Sub-Task 1 — `unitLabel` constant + tag pure-unit `It` block**
- Added `unitLabel = "unit"` to the `const` block in `suite_test.go`
- Applied `Label(unitLabel)` to `"test resolving watching namespace"` in `controllermanager_test.go` — the only Ginkgo spec with zero cluster interaction (calls only `common.ResolveWatchNamespaceForManager`)

**Sub-Task 2 — `isUnitOnlyRun()` + `BeforeSuite`/`AfterSuite` guards**
- Added `isUnitOnlyRun()` helper: exact match on `GinkgoLabelFilter() == "unit"`; compound filters like `"unit && foo"` still go through normal setup
- `BeforeSuite`: early-return when `isUnitOnlyRun()`, skipping all envtest/manager bootstrap
- `AfterSuite`: teardown path guarded with `!isUnitOnlyRun()`

**Sub-Task 3 — `k8sClient == nil` Skip guards**
- Added `if k8sClient == nil { Skip(...) }` as the true first statement in 13 integration `It` blocks that call `k8sClient` directly without a `USE_EXISTING_CLUSTER` guard:
  - `activemqartemis_controller_test.go` × 11
  - `activemqartemisaddress_controller_test.go` × 1
  - `activemqartemissecurity_controller_test.go` × 1

**Sub-Task 4 — Always bind metrics to `"0"`**
- Removed `disableMetrics bool` parameter from `createControllerManager`
- `Metrics.BindAddress = "0"` is now unconditional, preventing port conflicts when multiple test processes run concurrently
- Updated all 7 call sites: 1 in `suite_test.go`, 4 in `controllermanager_test.go`, 2 in `activemqartemis_controller_test.go`

## Relation to other PRs

- **#1477** (fixes #1474): merged as `919c2210` — branch rebased onto `cd304139` (5 commits past the merge), no conflicts
- **#1476** (fixes #1473): closed, never merged — none of its changes landed; no conflicts

## Verification

```
go test ./controllers/... --ginkgo.label-filter=unit
# Running Suite: Controller Suite
# Will run 1 of 270 specs
# Ran 1 of 270 Specs in 0.004 seconds
# SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 269 Skipped
```

`go build ./...` and `go vet ./...` both clean.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -103,6 +103,11 @@ const (
 
 	// Default ingress domain for tests
 	defaultTestIngressDomain = "tests.arkmq-org.io"
+
+	// unitLabel identifies Ginkgo specs that exercise pure Go logic with no
+	// API server or cluster dependency. Specs tagged with this label can run
+	// without KUBEBUILDER_ASSETS or USE_EXISTING_CLUSTER=true.
+	unitLabel = "unit"
 )
 
 var (
@@ -482,10 +487,10 @@ func cleanUpTestProxy() {
 }
 
 func createControllerManagerForSuite() {
-	createControllerManager(false, "")
+	createControllerManager("")
 }
 
-func createControllerManager(disableMetrics bool, watchNamespace string) {
+func createControllerManager(watchNamespace string) {
 
 	managerCtx, managerCancel = context.WithCancel(ctx)
 
@@ -515,10 +520,9 @@ func createControllerManager(disableMetrics bool, watchNamespace string) {
 		}
 	}
 
-	if disableMetrics {
-		// if we can shutdown metrics port, we don't need disable it.
-		mgrOptions.Metrics.BindAddress = "0"
-	}
+	// Always disable the metrics port to avoid bind conflicts when multiple
+	// test processes run concurrently.
+	mgrOptions.Metrics.BindAddress = "0"
 
 	waitforever := time.Duration(-1)
 	mgrOptions.GracefulShutdownTimeout = &waitforever
@@ -878,7 +882,19 @@ func setUpK8sClient() {
 	Expect(k8sClient).NotTo(BeNil())
 }
 
+// isUnitOnlyRun reports whether the active Ginkgo label filter selects only
+// unit-labeled specs. When true, no control-plane setup is needed.
+// The check is an exact string match: compound filters like "unit && foo"
+// still require a running cluster and must not bypass setup.
+func isUnitOnlyRun() bool {
+	return GinkgoLabelFilter() == unitLabel
+}
+
 var _ = BeforeSuite(func() {
+	if isUnitOnlyRun() {
+		return
+	}
+
 	opts := zap.Options{
 		Development: true,
 		DestWriter:  GinkgoWriter,
@@ -942,6 +958,10 @@ func cleanUpPVC() {
 }
 
 var _ = AfterSuite(func() {
+	if isUnitOnlyRun() {
+		return
+	}
+
 	By("tearing down the test environment")
 	if os.Getenv("USE_EXISTING_CLUSTER") == "true" {
 		cleanUpPVC()
--- a/controllers/activemqartemis_controller_test.go
+++ b/controllers/activemqartemis_controller_test.go
@@ -3459,6 +3459,9 @@ var _ = Describe("artemis controller", func() {
 	Context("SS delete recreate Test", func() {
 		It("deploy, delete ss, verify", func() {
 
+			if k8sClient == nil {
+				Skip("requires a cluster (run with KUBEBUILDER_ASSETS or USE_EXISTING_CLUSTER=true)")
+			}
 			crd := generateArtemisSpec(defaultNamespace)
 			By("Deploying the CRD " + crd.ObjectMeta.Name)
 			Expect(k8sClient.Create(ctx, &crd)).Should(Succeed())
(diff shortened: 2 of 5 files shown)


rii13024 (commented):
Addressed in 158d0414. Moved the `isUnitOnlyRun()` check to be an early return at the very top of `AfterSuite`, before any env-var checks (`USE_EXISTING_CLUSTER`, `DEPLOY_OPERATOR`). This mirrors the pattern already used in `BeforeSuite` and means no teardown path — including `uninstallOperator` — can execute during a unit-only run regardless of which env vars are set. The `else if !isUnitOnlyRun()` on the envtest teardown branch is now a plain `else`.

controllers/suite_test.go
```diff
@@ -103,6 +103,11 @@ const (
 
 	// Default ingress domain for tests
 	defaultTestIngressDomain = "tests.arkmq-org.io"
+
+	// unitLabel identifies Ginkgo specs that exercise pure Go logic with no
+	// API server or cluster dependency. Specs tagged with this label can run
+	// without KUBEBUILDER_ASSETS or USE_EXISTING_CLUSTER=true.
+	unitLabel = "unit"
```
lavocatt, line 110: I think that, we already identify unit tests by setting USE_EXISTING_CLUSTER=false in the CI, have you looked into that?

PR #1497, opened 2026-09-04
[#1496] Order prometheus config keys for consistent reconcile

### LINKED ISSUE #1496: Use an ordered map for prometheus config generation to force consistent generation
At https://github.com/arkmq-org/arkmq-org-broker-operator/blob/main/controllers/brokerservice_controller.go#L1117 we iterate over a map, which is unordered.
We need to make that consistent, iterate over ordered keys, such that we generate the same content every time in our reconcile.

See SortedKeys in broker properties for some inspiration.

> [!TIP]

MrBeldum:
Fixes #1496

## Summary
- `generatePrometheusConfig` iterated `appQueues` with an unordered `range` over a map, so prometheus YAML could differ across reconciles.
- Added `brokerproperties.SortedKeysBool` (same pattern as `SortedKeys` / `SortedKeysByteValue`) and use it when emitting `includeObjectNameAttributes`.

## Test plan
- [x] Unit coverage for `SortedKeysBool` ordering and empty map
- [ ] Existing BrokerService prometheus/reconcile tests still pass


gtully (approved):
thanks for the quick fix

PR #1500, opened 2026-09-09
[#1499] Fix nightly E2E failure in broker-label-test

### LINKED ISSUE #1499: Nightly E2E fails: broker-label-test missing cert-manager setup
- Nightly E2E CI fails on main after [#1443](https://github.com/arkmq-org/arkmq-org-broker-operator/pull/1443) merged. 
  https://github.com/arkmq-org/arkmq-org-broker-operator/actions/runs/34199934185/job/101976171374

**Root cause**

- The `broker-label-test` E2E test deploys a Broker CR without installing the required cert-manager secrets (operator cert, operand mTLS cert, prometheus cert). The Broker controller blocks reconciliation until those exist, so the StatefulSet is never created and the test times out.

- The Broker label implementation from [#1443](https://github.com/arkmq-org/arkmq-org-broker-operator/pull/1443)  is correct; the test setup was incomplete.

**Why PR CI didn’t catch it**
- PR-scoped E2E (test-mk-pr-v) only runs tests matching `broker-service|restricted` rbac. The broker-label-test label is excluded, so this test only runs in the full nightly suite (test-mk-v).

**Fix**
- Add cert-manager setup to broker_managed_resource_labels_test.go (same pattern as controll_plane_test.go).

Msarawan:
fixes: [Issue#1499](https://github.com/arkmq-org/arkmq-org-broker-operator/issues/1499)

### DIFF
--- a/controllers/broker_managed_resource_labels_test.go
+++ b/controllers/broker_managed_resource_labels_test.go
@@ -19,6 +19,8 @@ package controllers
 import (
 	"os"
 
+	cmv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	cmmetav1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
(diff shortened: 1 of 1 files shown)


controllers/broker_managed_resource_labels_test.go
```diff
@@ -97,17 +168,25 @@ var _ = Describe("broker managed resource labels", Label("broker-label-test"), f
...
 				g.Expect(createdBrokerCr.Status.ScaleLabelSelector).NotTo(ContainSubstring(selectors.LabelActiveMQArtemisKey + "="))
 			}, existingClusterTimeout, existingClusterInterval).Should(Succeed())
 
 			By("cleaning up")
-			CleanResource(createdBrokerCr, createdBrokerCr.Name, defaultNamespace)
+			CleanResource(&createdBrokerCr, createdBrokerCr.Name, defaultNamespace)
 		})
 	})
 
 	Context("ActiveMQArtemis CR", func() {
```
gtully, line 181: is that context text ok? seems like it should be Broker CR

Msarawan: Yes, the two contexts are intentional. The first (Broker CR) verifies the new Broker tracking label. The second (ActiveMQArtemis CR) ensures ActiveMQArtemis CRs still use the `ActiveMQArtemis` label and weren’t affected by the Broker labeler changes.

Issue #1506, opened 2026-09-11
BrokerService not becoming ready due to event driven reconciliation change


Robmurt4:
On the Build step of the CI, the tutorial docs/tutorials/secure_messaging_with_monitoring.md is failing on the step due to the BrokerService not becoming ready:
```bash {"stage":"deploy_service", "label":"wait for service", "runtime":"bash"}
kubectl wait BrokerService messaging-service -n broker-tutorial --for=condition=Ready --timeout=300s
```

The following error is output: 
<img width="2956" height="1178" alt="Image" src="https://github.com/user-attachments/assets/ec4d7f9c-eccd-4902-b0c9-402ff431735b" />

Testing locally, the following status is displayed after running `kubectl describe` on the BrokerService: 

> not ready broker status {[{Valid True 1 2026-09-11 17:41:58 +0000 UTC ValidationSucceded } {BrokerPropertiesApplied Unknown 0 2026-09-11 17:41:58 +0000 UTC UnableToRetrieveStatus no available brokers from deployed condition} {Deployed False 0 2026-09-11 17:41:58 +0000 UTC ResourceError failed to create status-script Role: roles.rbac.authorization.k8s.io "messaging-service-sidecar" is forbidden: user "system:serviceaccount:broker-tutorial:arkmq-org-broker-controller-manager" (groups=["system:serviceaccounts" "system:serviceaccounts:broker-tutorial" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
> {APIGroups:[""], Resources:["pods"], ResourceNames:["messaging-service-ss-0"], Verbs:["patch"]}} {Ready False 0 2026-09-11 17:41:58 +0000 UTC WaitingForAllConditions Some conditions are not met}] {[] [] []} app.kubernetes.io/component=broker-service,app.kubernetes.io/instance=messaging-service,app.kubernetes.io/managed-by=arkmq-org-broker-operator,application=messaging-service-app,broker.arkmq.org/peer-index=0,broker.arkmq.org/service=messaging-service,broker=messaging-service [] {2.56.0 quay.io/arkmq-org/arkmq-org-broker-kubernetes@sha256:373383631042dabae18a1390cbb8558789cdd7c74d67f1bfa0938f07d50b5ae6 } {true true true true}}

[...]

PR #1507, opened 2026-09-14
[#1506] sidecar rbac requires patch on pods

lavocatt:
The sidecar's RBAC requires patch on pods. The operator know has patch on pods otherwise it can't create a role with that permission. Caught by the nightly running the tutorials as the E2E tests failed to surface this issue.


Robmurt4 (approved):
LGTM

gtully (approved):

Issue #1505, opened 2026-09-11
Certmanager related tests don't work on a cluster with just certmanager installed


gtully:
executing the test suite: make test-mk-v on a clean minikube will install cert manager and trust manager and all is ok.
However if certmanager on its own in already installed, the logic to install trust manager does not kick in.
the problem is the method: CertManagerInstalled in https://github.com/arkmq-org/arkmq-org-broker-operator/blob/main/controllers/common_util_test.go#L1016 that is gating installing for each of the relevant tests does not check for trust manager being present, but the logic to install trust manager is tied to the cert manager install function.
If the check can be expanded to check both and the logic to install be smart enough to install one or both, it will negate the need to manually install trust manager.

tomatotomata:
Hi Gary, I was thinking of taking this one. My plan would be to keep the change focused in `common_util_test.go`: have `CertManagerInstalled` check both cert-manager and trust-manager, then make setup install only whichever dependency is missing, with an E2E-first test covering cert-manager-only, trust-manager-only, and both-present cases. Does that boundary match what you had in mind? Let me know what you think.

gtully:
there is some logic that tracks whether a test installed cert manager, and then uninstalls in the after method. It would be good to have that same logic around installing trust manager. 
It may be simpler to introduce separate methods for cert and trust manager installation and uninstall and call both for tests that require both. That means the relevant tests need updating but I think that may be best.
In any event, any PR contribution is greatly appreciated :-) Thank you @tomatotomata

tomatotomata:
I was thinking of separating the cert-manager and trust-manager setup and teardown helpers, then keeping tests that need both explicit about calling both. For coverage, I would add cert-manager-only, trust-manager-only, and both-present cases, and verify cleanup removes only the dependency that a test installed. That seems to follow your suggestion without changing unrelated test setup. Does that scope sound right?

gtully:
yes, but the existing tests will validate for you, once you run on a cluster that does not have trust manager installed.

tomatotomata:
I ran `make test-mk-pr-v` on a fresh cluster with no trust-manager installed. The existing broker-service setup installed trust-manager itself, so the missing-dependency path was exercised.

The run passed 11 of the 12 selected controller specs, then failed in SharedAddresses Scenario 11 (`brokerservice_address_sharing_test.go:1473`; expected two addresses, got none). That scenario is outside this change, so I could not call the full target green.

[...]

PR #1512, opened 2026-09-17
[#1505] Split cert-manager and trust-manager test setup

tomatotomata:
## Summary

- Split cert-manager and trust-manager install and uninstall helpers.
- Update integration fixtures that create trust-manager Bundles so they install trust-manager even when cert-manager is already present.
- Track test-owned trust-manager setup and preserve the shared namespace while either manager is still installed.

## Testing

- `make test-unit` passed.
- Focused real-cluster regression passed: 1 spec passed, 278 skipped.


controllers/common_util_test.go
```diff
@@ -956,17 +956,20 @@ func WaitForPods(crName string, iPods ...int32) {
 	}
 }
 
+func addJetstackHelmRepo() error {
+	cmd := exec.Command(helmCmd, "repo", "add", "jetstack", "https://charts.jetstack.io", "--force-update")
+	return cmd.Run()
+}
```
lavocatt, line 962: I'm wondering if we could make use of a bash script instead of individual commands.
Reference that bash script in the code, and execute it wit exec.Command.

It would mean an easier time to upgrade in the future instead of hunting individual commands in the comon_util_test.go file.

controllers/common_util_test.go
```diff
@@ -981,36 +984,50 @@ func InstallCertManager() error {
...
-		fmt.Printf("error waiting cert-manager %v\n", err)
+func UninstallTrustManager() error {
+	cmd := exec.Command(helmCmd, "uninstall", "-n", "cert-manager", "trust-manager", "--wait", "--timeout", "5m")
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+
+	// Remove the shared namespace only when cert-manager is not still installed.
+	if CertManagerInstalled() {
+		return nil
```
lavocatt, line 1010: Is there any reasonable situation where cert-manager isn't installed before trust-manager?

controllers/common_util_test.go
lavocatt: And I'm also wondering if these steps shouldn't be CI prerequisite instead.

I'd have an easier time with a global script that provision minikube, sets the ingress passthrough, deploys cert-manager, trust manager and then we can run the E2E tests. The tests would assume we have an infra that is ready.

It would also mean that contributors would have a script to execute to setup their own minikube test infrastructure on their local machine.

@gtully WDYT?

lavocatt (commented):
This looks good to me. Thanks for the contribution!

@gtully I'm wondering if we're not abstracting the installation bits in the wrong place though.

My thinking is that we could have a global script we run in CI scripts/minkube_setup.sh --profile $testProfile --memory $memoryAmount --cpu $cpuCount --projectionTime $projectionTime

The E2E test suite would assume a setup cluster.

Contributors would invoke the same script on their local machine prior to any make mk-*-v tests

WDYT?

controllers/common_util_test.go
gtully: @lavocatt I am with you on the prerequisites... it is how I run a local dev cluster. It makes sense to have such a script in the repo and referenced from ci etc. These changes only make it more obvious. It may be a separate PR though.

tomatotomata:
Thanks, that matches what I saw when I reran Scenario 11 on both the PR head and current `main`: it passed in both cases, so I’m treating the earlier 11-of-12 suite result as an unresolved environment or test-order issue rather than blaming this change. For this PR I was thinking of keeping the setup and teardown helpers separate, with each test class removing only the manager it installed. The shared profile-based setup script sounds useful if several suites need the same resource settings, but I’d keep it as a follow-up unless you’d prefer that broader setup here. I also checked the latest CI run: it is `action_required` and has no jobs, so there is no new failing log to diagnose yet. If repo-side approval is required for this fork run, someone with access will need to approve it.

PR #1514, opened 2026-09-17
Enable multi-tenant scraping with app-cert identity

lavocatt:
Apps collocated on one BrokerService can now scrape `:8888/metrics` with the certificate they already have, and get back only their own queues.


gtully (approved):
great. per tenant metrics is a bit part of the value prop. the need for the app cert up front makes sense as we can respect their content for the rbac rules based on CN.
Extracting the roles from the certs is also an option we can explore. see: https://issues.apache.org/jira/browse/ARTEMIS-5959

